PR #204, opened 2024-02-07
chore(analyze): Regenerate WebAssembly and bindings

blaine-arcjet:
This updates our Analyze WebAssembly and bindings, which includes the preinitialization with Wizer. You'll notice that this adds 500kb to the size of the wasm file.

We're already fighting file size problems, so we might want to ship the next release without this change.


blaine-arcjet:
Closing in favor of #334

PR #205, opened 2024-02-07
feat: Leverage local rate limiting if available in environment

### LINKED ISSUE #50: Provide local rate limit rule within TypeScript SDK if supported by environment
In non-edge/non-serverless environments, we could support a local rate limit rule using an in-memory data structure. We should implement this rule and feature detect the environment to know if it should be added to a ruleset.

blaine-arcjet:
This implements tracking local rate limits

Closes #50


blaine-arcjet:
We've decided that local rate limit only has limited benefits to the users, so we're going to close this out as not planned.

PR #221, opened 2024-02-10
docs(examples): Expanded AI example with rate limit by user ID

davidmytton:
Improvements to the Next.js example apps:

- Added an example for rate limiting by custom user ID.
- Improved the comments throughout the example.
- Added an environment var type definition to avoid TS complaining.
- Includes a new example to [submit to the Next.js repo](https://github.com/vercel/next.js/blob/canary/contributing/examples/adding-examples.md). Maintaining a copy of the code here so we can keep track of all the example apps we create.


examples/nextjs-13-pages-wrap/package-lock.json
blaine-arcjet, line 1: I'm concerned about the size of the change on this package-lock when no package.json was changed. It also seems to have generated by a different version of npm since it removed `name` in the linked path

blaine-arcjet (commented):
Just one concern on the lockfile churn in the next.js 13 example. I can look into it.

blaine-arcjet (approved):

blaine-arcjet:
Disabling the swc minify doesn't seem to stop it from failing to download the swc binaries. We just need to put up with the flakiness and add it back to the merge queue if it fails.

PR #227, opened 2024-02-13
docs: Update HTTP version

davidmytton:
We're now only supporting HTTP2. This is handled automatically by the Next.js SDK (which uses gRPC by default), but update the readme for the plain JS SDK.


blaine-arcjet (approved):

davidmytton:
The build failure is fixed in https://github.com/arcjet/arcjet-js/pull/221

PR #228, opened 2024-02-13
feat!: Separate ArcjetRequest and ArcjetRequestDetails types to accept record of headers

### LINKED ISSUE #33: Support plain objects for headers
The `ArcjetRequest` and `ArcjetRequestDetails` types currently limit the headers field to `Headers` because we usually work with the `ArcjetHeaders` abstraction; however, we might want to support records of headers being passed (possibly to support node's default `http.IncomingMessage`).

blaine-arcjet:
This separates our `ArcjetRequest` type, which users provide to the SDK's `protect()` function, and the `ArcjetRequestDetails`, which is an implementation detail of the protocol. This allows us to support various definitions for headers (and other fields in the future) while still normalizing for consumption by local and remote rules internally.

Closes #33


davidmytton (approved):
Looks good. https://github.com/arcjet/arcjet/issues/1263 to track the docs changes.

PR #245, opened 2024-02-21
feat: Add `withRule` API for adding adhoc rules

### LINKED ISSUE #193: Allow adding additional rules for one execution of `protect` without changing the client
We want users to be able to vary rules given context they may only have in the handler, but we don't want them to construct a whole new SDK client during each request.

We should add another API, such as `withRule()` that allows the SDK to be augmented for one call to `protect()`. For example:

```javascript
const aj = arcjet(...);

async function handler(req) {
  const decision = await aj.withRule(...someConfig).protect(req);
}
```

blaine-arcjet:
This adds the `withRule` API to our SDK. This allows adding adhoc rules (e.g. within a handler) in case the rules rely on some other logic within the application.

These adhoc rules are only available via augmented SDK returned by `withRule()` and don't affect the rest of the application. The augmented SDK also has `withRule` so these can be chained or augmented futher.

Closes #193


blaine-arcjet:
@davidmytton I updated the clerk-rl example to show how this works in the context of Clerk. Please let me know what you think of the usage.

davidmytton (approved):
Looks good. I slimmed the comments to match the slimmed down versions in the docs, and also added a bot rule to show how the rules get chained.

PR #247, opened 2024-02-23
feat: Add decorate package to set rate limit headers

### LINKED ISSUE #38: Better access to rate limit headers
Rate limiting headers are currently not easy to access, requiring `if (decision.reason.isRateLimit())` and so only available if the top level reason is a rate limit rule. 

```ts
if (decision.reason.isRateLimit()) {
    res.setHeader("RateLimit-Limit", decision.reason.max.toString());
    res.setHeader(
      "RateLimit-Remaining",
      decision.reason.remaining.toString()
    );
    const reset = decision.reason.resetTime?.getTime();
    if (reset !== undefined) {
      res.setHeader("RateLimit-Reset", (reset / 1000).toString());
    }
  }
```

This is fine where the rate limit is part of another protection e.g. protect signup, where we don't want to expose them. However, for APIs we might want to return the headers to the client on every request.

This issue could be to add the `decorate` method.

### LINKED ISSUE #40: TS SDK: Add helper function to decorate headers based on rule results
Currently, a user has easy access to the final `reason` but can also iterate the `results` on a decision to get more detailed information about every rule. We want to expose a helper function that can decorate headers on a request given a decision (we'll iterate the results and build appropriate headers).

blaine-arcjet:
This is the first pass at adding a decorate package that allows users to add RateLimit and RateLimit-Policy headers given an ArcjetDecision.

This follows the draft spec at https://ietf-wg-httpapi.github.io/ratelimit-headers/draft-ietf-httpapi-ratelimit-headers.html which has gone through a few iterations and will probably receive more updates soon, as it expired in December 2023.

This implementation requires some service changes to add fields to the protobuf.

Closes #38 
Closes #40 (we'll be able to add other decoration functions to this new package)


decorate/index.ts
```diff
@@ -0,0 +1,212 @@
...
+  decision: ArcjetDecision,
+) {
+  const rateLimitReasons = decision.results
+    .map(extractReason)
+    .filter(isRateLimitReason);
+
+  const policies = new Map<number, number>();
+  for (const reason of rateLimitReasons) {
+    if (policies.has(reason.max)) {
+      // TODO: Warn? Or does it make sense to bail since there is conflicting policies?
```
davidmytton, line 149: I suggest log a warning, but then set them anyway. The developer should know what they want when calling this. It's easier for us to become more strict e.g. throwing, later on that it will be to go the other way.

blaine-arcjet: Updated. There are a couple cases where we should return, but those are now `logger.error`

blaine-arcjet:
Due to our deny caching, the `reset` value may be "stuck" until the window has passed. This seems to be okay as per [the spec](https://ietf-wg-httpapi.github.io/ratelimit-headers/draft-ietf-httpapi-ratelimit-headers.html#name-reliability-of-the-reset-ke):
> Consider that service limit might not be restored after the moment referenced by the [reset keyword](https://ietf-wg-httpapi.github.io/ratelimit-headers/draft-ietf-httpapi-ratelimit-headers.html#ratelimit-reset-keyword) ([Section 3.4](https://ietf-wg-httpapi.github.io/ratelimit-headers/draft-ietf-httpapi-ratelimit-headers.html#ratelimit-reset-keyword)), and the reset keyword value may not be fixed nor constant.Subsequent requests might return a higher reset keyword value to limit concurrency or implement dynamic or adaptive throttling policies.

Which links to the [reset section](https://ietf-wg-httpapi.github.io/ratelimit-headers/draft-ietf-httpapi-ratelimit-headers.html#ratelimit-reset-keyword) saying:

> The client MUST NOT assume that all its service limit will be reset at the moment indicated by the reset keyword. The server MAY arbitrarily alter the reset keyword value between subsequent requests; for example, in case of resource saturation or to implement sliding window policies.

That being said, we are planning to revisit the caching mechanisms as part of https://github.com/arcjet/arcjet/issues/532 and should make sure we account for this.

davidmytton (approved):

blaine-arcjet:
@SocketSecurity ignore-all

PR #266, opened 2024-02-28
feat(decorate): Allow decorating Headers object directly

blaine-arcjet:
This allows decorating a `Header` object directly. This really cleans up the App Router example 🎉


davidmytton (approved):
Nice 😎

Issue #267, opened 2024-02-28
General thoughts on `withArcjet` for Nodejs SDK


blaine-arcjet:
The `withArcjet` wrapper we have in the Next SDK doesn't feel quite right for the Node.js SDK. Instead, it feels like `withArcjet` should return a connect/express-style middleware. However, in that situation, the decision should probably be attached to request (is that beneficial to users?).

Here's the initial sketch I did before realizing that it felt out of place:

```ts

interface ServerResponseLike {
  writeHead(statusCode: number, headers?: Record<string, string>): this;
  end(body: string): this;
}

function isServerResponseLike(val: unknown): val is ServerResponseLike {
  if (val === null) {
    return false;
  }

  if (typeof val !== "object") {
    return false;
  }

  if (!("writeHead" in val)) {
    return false;
  }

  if (!("end" in val)) {
    return false;
  }

  if (typeof val.writeHead !== "function" || typeof val.end !== "function") {
    return false;
  }

  return true;
}

/**
 * Wraps a Node.js handler.
 *
 * @param arcjet An instantiated Arcjet SDK
 * @param handler The request handler to wrap
 * @returns If the request is allowed, the wrapped `handler` will be called. If
 * the request is denied, a response will be returned based immediately and
 * the wrapped `handler` will never be called.
 */
export function withArcjet<Args extends [ArcjetNodeRequest, ...unknown[]], Res>(
  arcjet: ArcjetNode<WithoutCustomProps>,
  handler: (...args: Args) => Promise<Res>,
) {
  return async (...args: Args) => {
    const request = args[0];
    const response = args[1];
    const decision = await arcjet.protect(request);
    if (decision.isDenied() && isServerResponseLike(response)) {
      // TODO(#222): Content type negotiation using `Accept` header
      if (decision.reason.isRateLimit()) {
        response.writeHead(429, { "Content-Type": "application/json" });
        response.end(JSON.stringify({ code: 429, error: "Too Many Requests" }));
      } else {
        response.writeHead(403, { "Content-Type": "application/json" });
        response.end(JSON.stringify({ code: 403, error: "Forbidden" }));
      }
      return;
    } else {
      return handler(...args);
    }
  };
}
```

davidmytton:
[...]

PR #268, opened 2024-02-28
feat: Implement initial nodejs SDK

### LINKED ISSUE #49: Add default transport in `arcjet` npm package
Currently, we don't provide any default transport inside the `arcjet` npm package and instead create one in `@arcjet/next` when we create the SDK. We should provide a default in the main package, but it also needs to not break the edge compilation target.

blaine-arcjet:
This is an initial version of the Arcjet SDK for Node.js, which drastically smooths the process of using Arcjet with a simple Node server.

It implements much of the same logic as the Next SDK, but pulls data exclusively from a Node.js `IncomingMessage` type.

I've not implemented `createMiddleware` or `withArcjet` because they need addition considerations, for which I opened #267.

Closes #49

Draft because we need a bunch of docs that I've linked to in the README that don't exist.


davidmytton (approved):

davidmytton:
Should we have a TS example as well?

blaine-arcjet:
> Should we have a TS example as well?

I don't think so, as that requires a transpilation step since I really don't want to recommend ts-node.

davidmytton (approved):

PR #290, opened 2024-03-04
docs: Add node SDK and move core to utility section

blaine-arcjet:
Since we generally don't want people using the root `arcjet` package over one of the SDKs, I've moved it to the utility section. I've also added the node SDK to the SDK section.


davidmytton (approved):
Good idea. I've also added a few cross-links in the READMEs to highlight our other SDKs.

PR #296, opened 2024-03-04
chore(examples): Disable telemetry to tighten harden-runner

blaine-arcjet:
This disables telemetry in the nextjs examples following https://nextjs.org/telemetry

I'm doing this to allow us to restrict our harden-runner allowlist further.


davidmytton (approved):
Good idea.

PR #334, opened 2024-03-08
feat(analyze): Replace wasm-bindgen with jco generated bindings

blaine-arcjet:
This replaces our wasm-bindgen generated code with [jco](https://github.com/bytecodealliance/jco) generated code. Switching to `jco` aligns us with our desire to use the WebAssembly Component Model throughout our codebase. Through the Component Model, we can define concrete types in a `wit` file, like our `bot-type` enum:

```
enum bot-type {
    unspecified,
    not-analyzed,
    automated,
    likely-automated,
    likely-not-a-bot,
    verified-bot,
}
```

This allows us to have one definition and do minimal-to-no translation. This will also allow us to eventually do away with `JSON.stringify` and `JSON.parse` on inputs/outputs on the Wasm boundary.

Additionally, the `--instantiation` mode of jco allows us to do away with our custom wasm instantiation.

Since `jco` is used to produce various wasm "core" modules, I've moved the base64 module generation into a Rollup plugin in the `analyze` package. Any time a `.wasm` import (but not a `.wasm?module` import) is encountered during the build, it is translated into a virtual module and the base64 version is produced during the `load` phase.


analyze/wasm/arcjet_analyze_js_req.component.wasm
blaine-arcjet, line 1: Planning to shave off some more KB from this file with wasm-opt 117

davidmytton (approved):

analyze/wasm/arcjet_analyze_js_req.component.wasm
blaine-arcjet: We decided to leverage some more rust optimizations exclusively until I solve some problems with the binaryen setup

Issue #337, opened 2024-03-08
Consider removing Next.js linting rules from our generic rules


blaine-arcjet:
These rules do some weird and specific stuff for Next.js apps. We don't need them in our code eslint config and they'll still be applied to examples.

PR #344, opened 2024-03-11
chore: Make next a peerDep in our eslint package

blaine-arcjet:
This makes next a peerDep in our eslint package (similar to #339). I've made the version `>=14` because we only use this with next 14.

This is might be removed completely when we decide on #337


blaine-arcjet:
@SocketSecurity ignore-all

davidmytton (approved):

PR #559, opened 2024-04-11
chore(examples): Added new Next.js + React Hook Form example

### LINKED ISSUE #576: Dependency resolution for examples
With https://github.com/arcjet/arcjet/issues/1397 adding the Vercel deploy button requires referencing the repo path to clone. This installs the dependencies in the build on Vercel, but as we reference our own packages as relative files the install fails.

https://github.com/arcjet/arcjet-js/blob/a8ac93884b9b647a5111ae497fe887dae880878e/examples/nextjs-14-app-dir-rl/package.json#L12

It seems like there are a couple of options to solve this:

1. We change the dependency to point to `latest` which is what next.js does for its examples. The downside of this seems to be that it would pull the latest version from NPM, which would mean we couldn't test against the development `HEAD`

https://github.com/vercel/next.js/blob/c2b62d49c75f25c5f3423c370c9dbf8cc51b67ba/examples/api-routes/package.json#L9

2. We could add an override to the root `package.json` which would apply when we install the deps for the example within repo, but then the example `package.json` would reference `latest` when cloned in isolation. Will this work with how we have set up the workspace?

For discussion @blaine-arcjet 

**Build failed**

Screenshot 2024-04-12 at 9 37 53 AM

davidmytton:
Added new Next.js + React Hook Form example. Also adds the Vercel Deploy button ref https://github.com/arcjet/arcjet/issues/1397 and closes #576.


examples/nextjs-14-react-hook-form/README.md
```diff
@@ -0,0 +1,62 @@
+<a href="https://arcjet.com" target="_arcjet-home">
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="https://arcjet.com/arcjet-logo-minimal-dark-mark-all.svg">
+    <img src="https://arcjet.com/arcjet-logo-minimal-light-mark-all.svg" alt="Arcjet Logo" height="128" width="auto">
+  </picture>
+</a>
+
+# Protecting a Next.js React Hook Form with Arcjet
+
+[![Deploy with Vercel](https://vercel.com/button)](https://vercel.com/new/clone?repository-url=https%3A%2F%2Fgithub.com%2Farcjet%2Farcjet-js%2Ftree%2Fdavid%2Fexamples-react-hook-form%2Fexamples%2Fnextjs-14-react-hook-form&project-name=aj-react-hook-form&repository-name=aj-react-hook-form&redirect-url=https%3A%2F%2Fapp.arcjet.com%2Fintegrations%2Fvercel&developer-id=oac_1GEcKBuKBilVnjToj1QUwdb8&integration-ids=oac_1GEcKBuKBilVnjToj1QUwdb8)
```
blaine-arcjet, line 10: This will need to be changed to `main` before we merge

examples/nextjs-14-react-hook-form/README.md
```diff
@@ -0,0 +1,62 @@
...
+([docs](https://docs.arcjet.com/signup-protection/concepts)), but they can also
+be used separately on different routes.
+
+## How to use
+
+1. From the root of the project, install the SDK dependencies.
+
+   ```bash
+   npm ci
+   ```
```
blaine-arcjet, line 43: This step shouldn't be needed if we are using `latest` to download the arcjet packages. It is only needed to setup the monorepo dependencies if we are using a relative path for the SDK

examples/nextjs-14-react-hook-form/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+2. Enter this directory and install the example's dependencies.
+
+   ```bash
+   cd examples/nextjs-14-react-hook-form
```
blaine-arcjet, line 48: Is this accurate if it is meant to deploy with the vercel button?

examples/nextjs-14-react-hook-form/public/thirteen.svg
blaine-arcjet, line 1: Should this be 14? 🤔

blaine-arcjet (commented):
Since this is a Vercel-deploy specific example, can we change the example name to `vercel-next-react-hook-form`?

For the UI stuff, I just skimmed it, as I am unfamiliar with all the libraries used and I don't know how they interact.

examples/nextjs-14-react-hook-form/public/thirteen.svg
davidmytton: It's not used so I'll remove it.

examples/nextjs-14-react-hook-form/README.md
davidmytton: It's a generic example, but can be deployed with the Vercel button. So I've kept this as part of the general instructions.

examples/nextjs-14-react-hook-form/README.md
davidmytton: Ah yes, thanks.

davidmytton:
> Since this is a Vercel-deploy specific example, can we change the example name to vercel-next-react-hook-form?

I'd like to keep it as-is because although we're using this for the Vercel deploy button example, it's generic and was originally created as a generic example for tomorrow's blog post :)

Otherwise, have address your comments @blaine-arcjet

blaine-arcjet:
@SocketSecurity ignore npm/@swc/counter@0.1.3

PR #608, opened 2024-04-16
chore!: Deprecate calling `protect()` with no rules

### LINKED ISSUE #605: Deprecate calling `protect()` with no rules configured in the SDK
Currently, calling `protect()` with no rules configured in the SDK will still run the Shield rule implicitly in Arcjet's Decide service. 

We are deprecating the implicitness of Shield, so we'll want to warn if `protect()` is called with no rules. After the deprecation phase, this may turn into an explicit error.

blaine-arcjet:
Closes #605 

This adds a warning log if `protect()` is called with no rules configured that lets users know they should be configuring at least one rule, and suggests Shield because that is the common case where no rule would be specified before this behavior change.

### DIFF
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -1140,6 +1140,13 @@ export default function arcjet<
 
     const context: ArcjetContext = { key, fingerprint };
 
+    if (rules.length < 1) {
+      // TODO(#607): Error if no rules configured after deprecation period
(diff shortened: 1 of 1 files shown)


davidmytton (approved):
I've opened https://github.com/arcjet/arcjet/pull/1840 to document this.

PR #609, opened 2024-04-16
feat: Add configurable shield rule

### LINKED ISSUE #606: Add `shield` rule

blaine-arcjet:
This adds the configurable shield rule for users to add to their rules.

I also updated the protobuf generated code and cleaned up the fingerprint stuff that was removed.

@davidmytton I modified most examples to add `shield()` since we generally want people to enable it, but let me know if you had something else in mind.

Closes #606


davidmytton (approved):
Yeh looks good - including it in all the examples is a good idea so we can encourage usage.

PR #757, opened 2024-05-14
feat: Create Bun.sh adapter

### LINKED ISSUE #475: Official support for Bun
Either confirm the Node package works on Bun or provide a separate package with Bun support.

blaine-arcjet:
This creates an Arcjet SDK adapter for [bun.sh](https://bun.sh/). This was much the same as the Node.js adapter, but I needed to introduce a new API to the adapter: `aj.handler()`. This API is used to wrap the request handler (called `fetch()` in Bun) to do some preprocessing on the request—in this case looking up the IP with `server.requestIP(request)` before storing it in a WeakMap.

Users *could* use this without using the `aj.handler()` function, but they'd need to have a proper IP existing in the headers, such as the `Fly-Client-IP` on fly.io; however, those won't exist when running locally and the `protect()` call will error without using `aj.handler()`.

TODO:
- [x] Add to CI and dependabot
- [x] SDK Stack in protobuf
- [x] Examples

Closes #475


arcjet-bun/index.ts
```diff
@@ -0,0 +1,229 @@
...
+  // The base URL for the Arcjet API. Will default to the standard production
+  // API unless environment variable `ARCJET_BASE_URL` is set.
+  const baseUrl = options?.baseUrl ?? defaultBaseUrl();
+
+  // Transport is the HTTP client that the client uses to make requests.
+  const transport =
+    options?.transport ??
+    createConnectTransport({
+      baseUrl,
+      httpVersion: "1.1",
```
davidmytton, line 72: Is switching this to `2` blocked by https://github.com/arcjet/arcjet/issues/1180 ?

blaine-arcjet: Yep

blaine-arcjet:
@SocketSecurity ignore npm/@types/ws@8.5.10

examples/bun-rl/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "bun-rl",
+  "module": "index.ts",
+  "type": "module",
+  "workspaces": [
+    "../../*"
+  ],
```
blaine-arcjet, line 7: Since bun doesn't support `file:` specifiers, I used workspaces to resolve our monorepo packages.

blaine-arcjet:
The bun examples fall into similar camp to our nodejs examples in that we'd need to run the servers to test them so I'm not going to add them to CI.

examples/bun-hono-rl/index.ts
```diff
@@ -0,0 +1,42 @@
...
+    }),
+  ],
+});
+
+const app = new Hono();
+
+app.get("/", async (c) => {
+  const userId = "user123"; // Replace with your authenticated user ID
+
+  const decision = await aj.protect(c.req.raw, { userId, requested: 9 }); // Deduct 9 tokens from the bucket
```
blaine-arcjet, line 27: This works but I think we're going to want an actual Hono adapter that tries to determine which runtime it's inside. That was we can just do `aj.protect(ctx)` with the hono context.

blaine-arcjet:
Examples are done so I think this is good for a full review.

blaine-arcjet:
Oh, and I chose `handler` instead of `wrap` for the API to wrap `fetch` handler.

davidmytton:
I push a change to the README to run the server in hot reload mode because that makes it easier for the developer to make changes. I also added some log lines to the server to show the result so it's clearer what is happening.

However, when I run the `bun-rl` example, I get 2 calls to Arcjet. Are you also seeing this?

```
bun run --hot index.ts
Started server http://localhost:3000
Arcjet Request ID req_01hy04aq4ge9yaamr9pg4e29t1
Arcjet decision DENY
Arcjet Request ID lreq_01hy04aq53fn18k0dvvfr6dh1r
Arcjet decision DENY
```

blaine-arcjet:
Just realized that this needs readme and stuff

davidmytton (approved):
When you load this in a browser you get 2 requests which is what confused me in my earlier comment. I've added a note about this to the README.

PR #775, opened 2024-05-16
feat: Create SvelteKit adapter

### LINKED ISSUE #754: SvelteKit Adapter
We're adding a SvelteKit example in #738 but we should make it really slick with an adapter.

blaine-arcjet:
This introduces a SvelteKit adapter so users of SvelteKit can pass a `RequestEvent` directly to their `aj.protect()` method.

I also updated the example to use this new adapter.

Closes #754


blaine-arcjet:
Depends on #783 and #784

PR #784, opened 2024-05-17
fix(analyze): Leverage string interpolation to import Wasm files on edge runtime

blaine-arcjet:
While working on #775, I found that Vite doesn't support the `?module` syntax that we use for loading Wasm files in the edge runtime. Instead, we can use the `/* @vite-ignore */` comment but it only applies to imports using string interpolation.

I checked the output of `next build` and it seems to prepare all the `.wasm` files in our `./wasm/` directory as "edge chunks" when I use this syntax, so I think we should be good in both scenarios.

### DIFF
--- a/analyze/index.ts
+++ b/analyze/index.ts
@@ -18,23 +18,27 @@ async function moduleFromPath(path: string): Promise<WebAssembly.Module> {
   }
 
   if (process.env["NEXT_RUNTIME"] === "edge") {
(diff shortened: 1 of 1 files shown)


blaine-arcjet:
One of the examples failed because it tried to parse the WebAssembly Component that we commit into the project. I've modified my original changes to account for this and just use a `wasmPrefix` variable to trick the bundlers into thinking they are loading dynamic files.

davidmytton (approved):

PR #796, opened 2024-05-21
docs: Remove wording that implies is Shield is added by default

blaine-arcjet:
This updates the wording when describing shield to avoid implying it is added by default. While it is true through the deprecation period, we want people to update their configuration to explicitly opt-in to shield.

I also cleaned up other various typos and wording.


davidmytton (approved):
I've also added links to the new adapters.

PR #799, opened 2024-05-21
fix(ip)!: Rework priority of IP detection

### LINKED ISSUE #798: Rework IP header lookups
We want to lookup IP headers for various platforms only if we know we are running on that platform. These are usually the most trusted headers since the service added them, but only if the SDK is running in that platform (otherwise they are untrustworthy and probably spoofed).

See: https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip

blaine-arcjet:
Closes #798 

This is a breaking change to how we detect IP from headers. I've changed it to prefer IP values on a "request-like" object, such as `ip` or `socket.remoteAddress`.

I also prioritized `CF-Connecting-IP` and `Fly-Client-IP` over `X-Forwarded-For` headers but they are only used when actually deployed to Cloudflare or Fly respectively. We'll want to guard other platforms where an environment variable is detectable but I couldn't others in some cursory research.

Additionally, we now iterate the `X-Forwarded-For` IPs in reverse since the first items are the easiest to spoof. For example, a user can just submit whatever IP address as the first entry in `X-Forwarded-For` along with their request.

All of the above should be documented via code comments too.

Oh, and I removed the `CF-Pseudo-IPv4` header because it will always be wrong since Cloudflare generates a non-public IP address as the translation of the IPv6 so we're preferring `CF-Connecting-IPv6` correctly now.

### DIFF
--- a/ip/index.ts
+++ b/ip/index.ts
@@ -583,8 +583,57 @@ export interface RequestLike {
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
 function findIP(request: RequestLike, headers: Headers): string {
+  // Prefer anything available via the platform over headers since headers can
+  // be set by users. Only if we don't have an IP available in `request` do we
+  // search the `headers`.
   if (isGlobalIP(request.ip)) {
     return request.ip;
+  }
+
+  const socketRemoteAddress = request.socket?.remoteAddress;
+  if (isGlobalIP(socketRemoteAddress)) {
+    return socketRemoteAddress;
+  }
+
+  const infoRemoteAddress = request.info?.remoteAddress;
+  if (isGlobalIP(infoRemoteAddress)) {
+    return infoRemoteAddress;
+  }
+
+  // AWS Api Gateway + Lambda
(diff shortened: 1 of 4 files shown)


ip/index.ts
```diff
@@ -630,20 +670,13 @@ function findIP(request: RequestLike, headers: Headers): string {
     return fastlyClientIP;
   }
 
-  // Akamai and Cloudflare
+  // Akamai
   // True-Client-IP
   const trueClientIP = headers.get("true-client-ip");
```
davidmytton, line 675: Should we put this within the conditional that checks whether we're running on CF?

blaine-arcjet: No, because at https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#true-client-ip-enterprise-plan-only Cloudflare states that it is for enterprise customers only and 
>There is no difference between the True-Client-IP and CF-Connecting-IP headers besides the name of the header. Some Enterprise customers with legacy devices need True-Client-IP to avoid updating firewalls or load-balancers to read a custom header name.

Which means the we'll find CF-Connecting-IP first so this is only to support Akamai, which doesn't have an env var we can look at.

davidmytton (approved):

PR #824, opened 2024-05-24
feat!: Separate `@arcjet/headers` package from core

### LINKED ISSUE #224: Create headers package
We currently expose `ArcjetHeaders` from the core SDK, but we should make it a separate package.

blaine-arcjet:
Closes #224 

This separates the `ArcjetHeaders` package into its own package. This means it is no longer exported from `arcjet` and anyone consuming it will need to depend on `@arcjet/headers` directly.

Separating it out gave me a chance to think through unnatural usage in non-TypeScript environments so I added tests and guards around those situations. Having this package separate will allow us to improve/fix it independent of arcjet core.


davidmytton (approved):
Logged https://github.com/arcjet/arcjet/issues/2206 to update the docs before this is released.

PR #841, opened 2024-05-29
feat!: Allow ArcjetContext extension via new argument to core `protect()`

blaine-arcjet:
This is a breaking change to the core `arcjet` package because it adds a new `ctx` argument as the first parameter to `protect()`. This argument is meant to be used by adapters to provide additional information when calling `protect()`—for example, the `ARCJET_KEY` secret passed into Cloudflare's handler.  Most users won't notice this because they are using an adapter which don't expose the `ctx` argument to the user.

### DIFF
--- a/arcjet-bun/index.ts
+++ b/arcjet-bun/index.ts
@@ -197,7 +197,7 @@ function withClient<const Rules extends (Primitive | Product)[]>(
         props ?? {},
       ) as ArcjetRequest<ExtraProps<Rules>>;
 
-      return aj.protect(req);
+      return aj.protect({}, req);
     },
     handler(
       fetch: (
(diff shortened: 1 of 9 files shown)


davidmytton (approved):
I really like this approach. So with the existing adapters like `arcjet-next` the `ctx` prop is passed in through the adapter, so it's transparent to the developer (they don't need to provide it).

But with adapters where we do need it, like a future Cloudflare or Hono adapter, it can be provided? The client instance is still created outside the handler, but we pass through the key on each invocation where we have access to the environment variables.

blaine-arcjet:
> But with adapters where we do need it, like a future Cloudflare or Hono adapter, it can be provided? The client instance is still created outside the handler, but we pass through the key on each invocation where we have access to the environment variables.

Yep! We'll expose it to the users where necessary, but we should augment it ourselves wherever possible.

Issue #855, opened 2024-06-03
Allow logger to be customized with a structured logger


blaine-arcjet:
We want compatibility with `pino` so users can swap our logger for it when they want structured logging with custom transports.

Issue #51, opened 2023-10-23
Avoid nodejs-specific imports in generic packages


blaine-arcjet:
Our JS version of fingerprinting imports `crypto` which is a nodejs-specific import. We should instead rely upon the `Crypto` or `SubtleCrypto` global (which is also available in node).

- [x] https://github.com/arcjet/arcjet-js/issues/885

blaine-arcjet:
I'm currently working to remove our use of `util.format` but I also need to dive into our use of `process.env` since that global doesn't exist in Cloudflare Workers.

Issue #769, opened 2024-05-15
Tests in Bun adapter should be written with the Bun test harness


blaine-arcjet:
We currently don't test adapters. When we add them to the Bun adapter, we should write them with the Bun test harness instead of Jest.

blaine-arcjet:
Bun 1.2.6 added support for `node:test` that runs on top of the Bun test runner.

See https://bun.sh/blog/bun-v1.2.6#initial-support-for-node-test

wooorm-arcjet:
I guess this can be closed, as it’s about using `node:test` everywhere. Or, the issue could be renamed `Add tests to adapters`

blaine-arcjet:
You are correct that adapters don't currently have tests. We should change the title.

PR #856, opened 2024-06-03
feat: Create sprintf package to replace util.format

blaine-arcjet:
This adds a new `@arcjet/sprintf` package to this monorepo, which provides a platform-independent formatter function. This will allow us to move closer to #51 by replacing the `util.format` import!

I've kept the implementation pretty close to the [reference library](https://github.com/pinojs/quick-format-unescaped) because [pino](https://github.com/pinojs/pino) uses it for formatting strings and we want pino to be a drop-in replacement for our logger when users want to [swap in structured logging](https://github.com/arcjet/arcjet-js/issues/855).


sprintf/README.md
```diff
@@ -0,0 +1,70 @@
...
+<p>
+  <a href="https://www.npmjs.com/package/@arcjet/sprintf">
+    <picture>
+      <source media="(prefers-color-scheme: dark)" srcset="https://img.shields.io/npm/v/%40arcjet%2Fsprintf?style=flat-square&label=%E2%9C%A6Aj&labelColor=000000&color=5C5866">
+      <img alt="npm badge" src="https://img.shields.io/npm/v/%40arcjet%2Fsprintf?style=flat-square&label=%E2%9C%A6Aj&labelColor=ECE6F0&color=ECE6F0">
+    </picture>
+  </a>
+</p>
+
+[Arcjet][arcjet] platform-independent replacement for [util.format][node-util].
```
davidmytton, line 19: It would be worth noting the rationale for building this - why have we created our own `sprintf` package?

sprintf/jest.config.js
```diff
@@ -0,0 +1,16 @@
...
+  moduleFileExtensions: ["js"],
+  coverageDirectory: "coverage",
+  collectCoverage: true,
+  // If this is set to default (babel) rather than v8, tests fail with the edge
+  // runtime and the error "EvalError: Code generation from strings disallowed
+  // for this context". Tracking in
+  // https://github.com/vercel/edge-runtime/issues/250
+  coverageProvider: "v8",
+  verbose: true,
+  testEnvironment: "node",
```
davidmytton, line 13: Is it worth testing in the different runtimes we support? E.g. https://bun.sh/docs/cli/test

blaine-arcjet: Not really. I don't want to spend further time right now to figure out how to get Bun into the CI and run different sets of tests in the monorepo (ref #769).

There are no node imports but we don't want jest running in jsdom so this is just the easiest with our current harness.

sprintf/README.md
davidmytton: Kind of, but it doesn't explain the general rationale e.g. `platform-independent so we can support multiple runtimes in varying environments, such as Edge Runtime, Node.js, Bun, etc.`

blaine-arcjet: Okay, but that's the definition of platform-independent 🙃 
```suggestion
[Arcjet][arcjet] platform-independent replacement for [util.format][node-util].

This package is platform-independent in order to support multiple runtimes in varying environments, such as Edge Runtime, Node.js, Bun, Deno, and Cloudflare Workers.
```

PR #858, opened 2024-06-03
feat(logger)!: Align logger with Pino API

### LINKED ISSUE #822: Logger isn't configured if environment variables aren't set upon module import
Currently, the singleton logger is constructed when `@arcjet/logger` is imported. This means it uses the `ARCJET_LOG_LEVEL` environment variable at that moment. If the environment variable is set async, such as via Deno's dotenv stdlib, the logger doesn't get constructed with the correct log level.

### LINKED ISSUE #855: Allow logger to be customized with a structured logger

blaine-arcjet:
This tries to align our logger interface with Pino so users will be able to swap it in when they want structured logging and custom transports.

I also tried to mimic `pino-pretty` when it was easy (no colors, no pretty-printing on nested objects).

Currently a draft because I need to change all the usage throughout our SDK.

Closes #822 (once I make it configurable throughout each SDK)
Closes #855


logger/index.ts
```diff
@@ -1,45 +1,97 @@
...
   }
 }
 
+export type LogLevel = "debug" | "info" | "warn" | "error";
+
+export interface LoggerOptions {
+  level: LogLevel;
+}
+
+const PREFIX = "✦Aj";
```
blaine-arcjet, line 26: The logs would look nicer if this were `✦AJ` because we know print `✦Aj INFO my message` to align closer to pino-pretty. But I know the brand guidelines say we are supposed to use the lowercase `j`.

cc @davidmytton @morekid

protocol/index.ts
```diff
@@ -744,8 +744,24 @@ export interface ArcjetShieldRule<Props extends {}> extends ArcjetRule<Props> {
...
+  debug(obj: Record<string, unknown>, msg?: string, ...args: unknown[]): void;
+  info(msg: string, ...args: unknown[]): void;
+  info(obj: Record<string, unknown>, msg?: string, ...args: unknown[]): void;
+  warn(msg: string, ...args: unknown[]): void;
+  warn(obj: Record<string, unknown>, msg?: string, ...args: unknown[]): void;
+  error(msg: string, ...args: unknown[]): void;
+  error(obj: Record<string, unknown>, msg?: string, ...args: unknown[]): void;
+}
+
+export interface ArcjetContext {
```
blaine-arcjet, line 763: Switched this to an interface because otherwise using `Omit<>` seems to drop the inference on other values.

logger/index.ts
morekid: `✦Aj INFO` vs `✦AJ INFO `

I see your point alignment-wise, but the current version is more readable in general (full caps is usually less so, and "AJ" in particular looks a bit jarring), also easier to understand as "Arcjet" abbreviation, and preserves its recognizability when encountered elsewhere.

morekid: we could consider just having the unicode `✦` eventually for all logs if that gets in the way? not sure it'll be enough to differentiate and make them look like Arcjet logs specifically.

blaine-arcjet: > not sure it'll be enough to differentiate and make them look like Arcjet logs specifically.

I think we need some sort of differentiator, like `Aj`. I just think `Aj INFO` looks weird because `INFO`/`WARN`/`ERROR` are fully capitalized.

davidmytton: I wonder if this is a scenario where an exception to the case rule makes sense. We're Arcjet, not ArcJet, so normally I'd say `Aj` is the correct casing. But in this context where the whole prefix to the log is upper case, `Aj` looks a bit weird. Does the context overrule being consistent?

logger/index.ts
```diff
@@ -1,45 +1,93 @@
...
+
+function getOutput(obj: unknown, msg: unknown, args: unknown[]) {
+  let output = getMessage(obj, msg, args);
+  if (typeof output !== "string") {
+    return;
+  }
+
+  if (typeof obj === "object" && obj !== null) {
+    for (const [key, value] of Object.entries(obj)) {
+      output += `\n    ${key}: ${tryStringify(value)}`;
```
blaine-arcjet, line 61:
```suggestion
      output += `\n✧    ${key}: ${tryStringify(value)}`;
```
@morekid like this?

blaine-arcjet: We keep the examples in the `examples/` directory in this repository so they can be tested with the workspace changes. Here's the output from one of them:
```
✦Aj DEBUG Decide response
✧    id: "req_01hzpvr7nkftn9ct33a62q0vrz"
✧    fingerprint: "fp_1_3e48ef9d22e096da6838540fb846999890462c8a32730a4f7a5eaee6945315f7"
✧    path: "/api/arcjet"
✧    runtime: "node"
✧    ttl: 0
✧    conclusion: "ALLOW"
✧    reason: {"type":"SHIELD","shieldTriggered":false}
✧    ruleResults: [{"ruleId":"","ttl":0,"state":"RUN","conclusion":"ALLOW","reason":{"type":"SHIELD","shieldTriggered":false}}]
✦Aj LATENCY decideApi: 88.984ms
✦Aj LATENCY remote: 89.027ms
```

blaine-arcjet: We're going to avoid this right now because it is not a multi-line text description but instead properties from the structured logger.

logger/index.ts
blaine-arcjet: I think it's decided that we're just going to keep the lowercase j so the logs look like:

Notice that I did add 2 extra spaces so the object properties wouldn't be exactly aligned with `DEBUG`
```
✦Aj DEBUG Decide request to https://decide.arcjet.orb.local:4082
✦Aj DEBUG Decide response
      id: "req_01hzpy6jz7ftn8ajsn2kdzkdbe"
      fingerprint: "fp_1_3e48ef9d22e096da6838540fb846999890462c8a32730a4f7a5eaee6945315f7"
      path: "/api/arcjet"
      runtime: "node"
      ttl: 0
      conclusion: "ALLOW"
      reason: {"type":"SHIELD","shieldTriggered":false}
      ruleResults: [{"ruleId":"","ttl":0,"state":"RUN","conclusion":"ALLOW","reason":{"type":"SHIELD","shieldTriggered":false}}]
✦Aj LATENCY decideApi: 121.4ms
✦Aj LATENCY remote: 121.436ms
```

morekid: @blaine-arcjet looks good to me, indentation as well, ✦ftw!

morekid (approved):

davidmytton (approved):

PR #870, opened 2024-06-05
docs: Add quick start links & update Bun example

davidmytton:
- Link to quick starts
- Adjust Bun example to match the docs - mainly added more comments, but also switched to token bucket as it's more interesting.
- Remove the Next.js example that was mirroring the Vercel repo because I closed the PR. Will open a new one once Next.js 15 is out and we can show off more functionality.


README.md
```diff
@@ -15,20 +15,22 @@
 </p>
 
 [Arcjet][arcjet] helps developers protect their apps in just a few lines of
-code. Implement rate limiting, bot protection, email verification & defense
+code. Implement rate limiting, bot protection, email verification, and defend
```
blaine-arcjet, line 18: "defend" doesn't sound right here. I think it should be "defense". If we read the sentence (without the list) as "Implement defense against common attacks" it sounds correct, while "Implement defend against common attacks" doesn't.

examples/bun-rl/index.ts
```diff
@@ -16,30 +20,44 @@ const aj = arcjet({
...
     return new Response("Hello world");
   }),
 };
 
 // Or using the `Bun.serve()` API
-// const server = Bun.serve({
-//   port: 3000,
-//   fetch: aj.handler(async (req) => {
-//     const decision = await aj.protect(req);
+/*
```
blaine-arcjet, line 41: The keyboard shortcut for "uncomment all lines" in VSCode works nicer with line comments than block comments.

blaine-arcjet (changes requested):
Commented inline

examples/bun-rl/index.ts
davidmytton: You can use toggle block comment shortcut as well. Removing this style of comment is easier if you don't know the shortcut.

blaine-arcjet (commented):
I just merged the base branch because it added the `runtime` package which needs the logo update.

blaine-arcjet (approved):

PR #897, opened 2024-06-07
feat!: Move all environment lookup into separate package

### LINKED ISSUE #51: Avoid nodejs-specific imports in generic packages

### LINKED ISSUE #885: Remove usage of process.env in generic packages
We use `process.env` pretty liberally throughout our packages; however, the global is not supported in workers so you need to `import * as process from "node:process"` but then that crashes in Vercel edge functions because you can't import node modules.

blaine-arcjet:
This introduces a new `@arcjet/env` package that provides various functions which receive an "env" object (such as `Bun.env` or `process.env`) and look for various keys that we care about. These keys are part of the `Env` type exported by the package to have a centralized definition of the properties.

By separating this logic, we remove all Node.js or other runtime specific logic in our generic packages. The "env" object lookup is then done in each adapter.

Closes #51 
Closes #885 

TODO:
- [x] Tests for env package
- [x] Docs for env package


env/index.ts
```diff
@@ -0,0 +1,78 @@
+export type Env = {
```
davidmytton, line 1: Does it make sense to bring in `ARCJET_KEY` here as well?

blaine-arcjet: Yep. I was planning to add a helper function for it when we added cloudflare, but I can just set that up now.

env/README.md
```diff
@@ -0,0 +1,38 @@
+<a href="https://arcjet.com" target="_arcjet-home">
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="https://arcjet.com/logo/arcjet-dark-lockup-voyage-horizontal.svg">
+    <img src="https://arcjet.com/logo/arcjet-light-lockup-voyage-horizontal.svg" alt="Arcjet Logo" height="128" width="auto">
```
blaine-arcjet, line 4: I need to update this

davidmytton (approved):

PR #929, opened 2024-06-11
chore!: Remove logger dependency from core

### LINKED ISSUE #922: Remove logger dep from arcjet core
Now that loggers are created in adapters, we can remove the dep from arcjet core.

blaine-arcjet:
Closes #922 

Leftover dependency from some refactors.


blaine-arcjet:
This is actually a devDep currently but I think we should drop it and just mock a logger.

davidmytton (approved):

Issue #90, opened 2023-12-22
Remove production check around baseUrl allowlist before 1.0.0


blaine-arcjet:
In #83, we added an allowlist check for the `ARCJET_BASE_URL` environment variable; however, validation is only done in when `process.env["NODE_ENV"] === "production"` to allow for experimentation in non-production environments.

Before we hit 1.0.0, we want to remove this conditional. Users will still be able to use baseUrls not on the allowlist via custom clients.

blaine-arcjet:
@davidmytton with my comment at https://github.com/arcjet/arcjet-js/pull/932#issuecomment-2162941862, do you want to delay this further or do it before beta?

davidmytton:
Let's do it before 1.0.0 and delay it for beta. Using the env var is easier for us :)

PR #932, opened 2024-06-11
chore!: Move client into protocol and rename builders in adapters

### LINKED ISSUE #46: Remove createRemoteClient in `arcjet` and defer to adapters
These are internal details that help us or end-users with debugging, so I don't think these should be able to be changed via options. If that's our decision, we'll probably need to duplicate some code between `arcjet` and `@arcjet/next` for creating the RemoteClient since we won't be able to pass the SDK information as options.

blaine-arcjet:
Closes #46

This moves the `createClient` logic from the `arcjet` core package into the `@arcjet/protocol` package. It is then consumed by adapters, which all export a `createRemoteClient` function that fills in some of the required data. This ensures we don't try to double export these functions (via re-exports) while still allowing users to dive deep into `@arcjet/protocol/client.js` if they need it.


blaine-arcjet:
@davidmytton this cleans up the external APIs but I think it makes it **harder** to swap `baseUrl` and other configuration. So we might want to delay #90 further

davidmytton:
Why is it harder? Is this example how you'd do it?

https://github.com/arcjet/arcjet-js/blob/d78c4aba7ea7e41686032b41625015b905b28d0c/examples/nextjs-14-app-dir-rl/app/api/custom_timeout/route.ts

blaine-arcjet:
> Why is it harder? Is this example how you'd do it?

The example you linked changes the baseUrl but not the transport. But you mentioned that you wanted to tweak transports for some tests. Since I've taken away the ability to change the transport via an adapter, you need to reach into `@arcjet/protocol` to build a client. The `createClient` function in our protocol requires all arguments, so it would look something like:

```ts
import { createClient } from "@arcjet/protocol"
import arcjet, { shield } from "@arcjet/next"
import { createConnectTransport } from "@connectrpc/connect-web";

const baseUrl = "..."

const client = createClient({
    baseUrl,
    timeout: 500,
    sdkStack: "NEXTJS",
    sdkVersion: "1.0.0-alpha.14",
    transport: createConnectTransport({
      baseUrl,
    });
});

const aj = arcjet({
    key: process.env.ARCJET_KEY!,
    client,
    rules: [shield()],
});
```

Issue #987, opened 2024-06-17
Improve DX around missing IP and fallback


davidmytton:
At https://github.com/arcjet/arcjet-js/pull/895#discussion_r1631442244 we discussed using a fallback `ip` of `127.0.0.1` if the IP is missing and we're in development mode. This works so long as `NODE_ENV` (or equivalent) is set, otherwise the IP is empty and the user gets an error:

```
✦Aj WARN generateFingerprint: ip is empty
✦Aj ERROR Encountered problem getting remote decision: [failed_precondition] client IP not provided
```

I've now seen 2 cases (one reported internally by @bendechrai and another with a user who reported it on Discord) where the environment isn't detected because `process.env.NODE_ENV` isn't set. 

For example, Express.js [does not set it by default](https://expressjs.com/en/advanced/best-practice-performance.html#set-node_env-to-production), so for anyone getting started with Express it will be `undefined` and they'll get the client IP error from Arcjet.

You can fix this by setting `NODE_ENV` or by adding `ARCJET_ENV=development` to the `.env` file, but we should provide more help to developers. I will update the Troubleshooting docs with this error, but we should also implement a log warning if we try to get `NODE_ENV` and it's not set, as well as the planned log that the IP is being overwritten:

https://github.com/arcjet/arcjet-js/blob/f275912687a7ca8278dcd4ade4124b1c5869131e/arcjet-bun/index.ts#L174

davidmytton:
Updated the docs in https://github.com/arcjet/arcjet/pull/2387/commits/6f7e910843d73874979d46349448b811dcc112fb

PR #988, opened 2024-06-17
chore(examples): Update .env file with ARCJET_ENV when NODE_ENV isn't set

davidmytton:
Helps with https://github.com/arcjet/arcjet-js/issues/987


blaine-arcjet (changes requested):
I think we need to change anything that is `nodejs-` prefixed because NODE_ENV doesn't actually seem set by node or npm. My guess is that it was probably a webpack thing that everyone started copying 🤔

PR #1000, opened 2024-06-18
chore: Warn when IP is empty, even if we override it in development

### LINKED ISSUE #216: Add logging to arcjet-next
Our `arcjet-next` package doesn't have any logging. This is due to the logger is instantiated inside `arcjet` and not exposed for `arcjet-next`. We don't want 2 instances of the logger in one instance of the SDK so this will take additional work to use a single one.

_Originally posted by @blaine-arcjet in https://github.com/arcjet/arcjet-js/pull/214#discussion_r1483567301_

### LINKED ISSUE #902: Refactor adapters so all functions have access to logger
In #897, the logger is being constructed in each adapter, but much of the adapter implementation doesn't have access to the logger. We need to refactor the helpers to be inside the closure so they can access the logger.

### LINKED ISSUE #987: Improve DX around missing IP and fallback

blaine-arcjet:
@davidmytton Are these the types of logs you were expecting if the IP is empty?

Closes #902 
Closes #987
Closes #216

### DIFF
--- a/arcjet-bun/index.ts
+++ b/arcjet-bun/index.ts
@@ -188,9 +188,12 @@ export default function arcjet<const Rules extends (Primitive | Product)[]>(
       // If the `ip` is empty but we're in development mode, we default the IP
       // so the request doesn't fail.
       if (isDevelopment(env)) {
-        // TODO: Log that the fingerprint is being overridden once the adapter
-        // constructs the logger
+        log.warn("Using 127.0.0.1 as IP address in development mode");
         ip = "127.0.0.1";
+      } else {
+        log.warn(
+          "Empty IP address in non-development mode. Did you mean to set ARCJET_ENV=development?",
+        );
       }
     }
 
--- a/arcjet-bun/index.ts
+++ b/arcjet-bun/index.ts
@@ -192,7 +192,7 @@ export default function arcjet<const Rules extends (Primitive | Product)[]>(
         ip = "127.0.0.1";
       } else {
         log.warn(
-          "Empty IP address in non-development mode. Did you mean to set ARCJET_ENV=development?",
+          `Client IP address is missing. If this is a dev environment set the ARCJET_ENV env var to "development"`,
         );
       }
     }
--- a/arcjet-bun/index.ts
+++ b/arcjet-bun/index.ts
@@ -140,102 +140,6 @@ export interface ArcjetBun<Props extends PlainObject> {
   ) => Response | Promise<Response>;
 }
 
-// This is provided with an `ipCache` where it attempts to lookup the IP. This
-// is primarily a workaround to the API design in Bun that requires access to
-// the `Server` to lookup an IP.
-function toArcjetRequest<Props extends PlainObject>(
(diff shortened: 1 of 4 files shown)


arcjet-bun/index.ts
```diff
@@ -249,12 +153,108 @@ export default function arcjet<const Rules extends (Primitive | Product)[]>(
...
+    );
+    if (ip === "") {
+      // If the `ip` is empty but we're in development mode, we default the IP
+      // so the request doesn't fail.
+      if (isDevelopment(env)) {
+        log.warn("Using 127.0.0.1 as IP address in development mode");
+        ip = "127.0.0.1";
+      } else {
+        log.warn(
+          "Empty IP address in non-development mode. Did you mean to set ARCJET_ENV=development?",
```
davidmytton, line 195: Reworded slightly to make it clearer:

```suggestion
          "Client IP address is missing. If this is a dev environment set ARCJET_ENV=development in your .env.local file.",
```

I'm also wondering if we should also point people to [the docs](https://docs.arcjet.com/troubleshooting#generatefingerprint-ip-is-empty--failed_precondition-client-ip-not-provided) where we have a more detailed expanation. That URL is quite long so we could provide shorter URL redirects for error messages. Next.js does this quite effectively e.g. https://nextjs.org/docs/messages/prerender-error

blaine-arcjet: Since not everyone will use `.env` files (I'm thinking primarily of dockerfiles here), I don't think we should explicitly mention that file. How about:

```suggestion
          `Client IP address is missing. If this is a dev environment set the ARCJET_ENV env var to "development"`,
```

davidmytton: Sounds good - I'm happy with your suggested change.

PR #1009, opened 2024-06-20
chore!: Move generated protobuf to default buf file path

davidmytton:
Following the update to the Buf config v2 in https://github.com/arcjet/arcjet/pull/2435 [the default path](https://buf.build/docs/migration-guides/migrate-v2-config-files#v2-configuration) is now `proto` so the `out` is specified as `../arcjet-js/protocol`


blaine-arcjet (changes requested):
I need to double check, but I believe we need to update the `files` section in the package.json to include the new directory.

blaine-arcjet (approved):
I updated the trunk.yaml to ignore the new proto directory and updated the `files` property to include it in the release.

PR #1096, opened 2024-07-09
feat: detect common free/disposable email providers locally

### LINKED ISSUE #1095: Update WASM to support new email validation interface
The wasm interface for email validation is changing so that it is the same between the sdk and the api. This necessitates a small change to the JS sdk so that it exports stubbed functions.

e-moran:
This PR regenerates the WASM to pull in some updates for our email validation primitive.
Notably it also provides functionality to detect the five most common free email providers locally, preventing a call to the Arcjet API.
This won't result in any behaviour change for email validation beyond potentially fewer api requests.

Closes #1095


arcjet/index.ts
```diff
@@ -554,20 +555,34 @@ export function validateEmail(
         context: ArcjetContext,
         { email }: ArcjetRequestDetails & { email: string },
       ): Promise<ArcjetRuleResult> {
-        if (await analyze.isValidEmail(context, email, analyzeOpts)) {
+        const result = await analyze.isValidEmail(context, email, emailOpts);
+        console.log(result);
```
blaine-arcjet, line 559:
```suggestion
```

arcjet/index.ts
```diff
@@ -554,20 +555,34 @@ export function validateEmail(
...
+          const typedEmailTypes = result.blocked.filter(
+            (type): type is ArcjetEmailType => {
+              return (
+                type === "FREE" ||
+                type === "DISPOSABLE" ||
+                type === "NO_MX_RECORDS" ||
+                type === "NO_GRAVATAR" ||
+                type === "INVALID"
+              );
+            },
```
blaine-arcjet, line 577: This should be a named function near the top of the file. We don't want to be creating a new function on every execution.

analyze/workerd.ts
```diff
@@ -97,14 +123,22 @@ export async function isValidEmail(
...
+    requireTopLevelDomain: true,
+    allowDomainLiteral: false,
+    blockedEmails: [],
+  };
 
   if (typeof analyze !== "undefined") {
-    return analyze.isValidEmail(candidate, options);
+    return analyze.isValidEmail(candidate, optionsOrDefault);
   } else {
     // TODO: Fallback to JS if we don't have WASM?
```
blaine-arcjet, line 137:
```suggestion
```
We're not falling back because we're pretty certain we have Wasm in every environment we target.

analyze/workerd.ts
```diff
@@ -43,6 +44,31 @@ async function init(context: AnalyzeContext) {
...
+    "arcjet:js-req/email-validator-overrides": {
+      isFreeEmail(domain) {
+        if (
+          [
+            "gmail.com",
+            "yahoo.com",
+            "hotmail.com",
+            "aol.com",
+            "hotmail.co.uk",
+          ].includes(domain)
```
blaine-arcjet, line 56: Do we need to check for `@`? 
```suggestion
          [
            "@gmail.com",
            "@yahoo.com",
            "@hotmail.com",
            "@aol.com",
            "@hotmail.co.uk",
          ].includes(domain)
```
Also, please move the email address array outside of the functions. It can be a static list at the top of the file.

analyze/workerd.ts
```diff
@@ -97,14 +123,22 @@ export async function isValidEmail(
...
   candidate: string,
   options?: EmailValidationConfig,
-) {
+): Promise<EmailValidationResult> {
   const analyze = await init(context);
+  const optionsOrDefault = options || {
+    requireTopLevelDomain: true,
+    allowDomainLiteral: false,
+    blockedEmails: [],
+  };
```
blaine-arcjet, line 132: I think this is the incorrect way to default these. I think we need something like:
```suggestion
  const optionsOrDefault = {
    requireTopLevelDomain: true,
    allowDomainLiteral: false,
    blockedEmails: [],
    ...options,
  };
```
I believe this will both in the scenario where `options` is undefined **and** when it does not contain all properties.

blaine-arcjet (changes requested):
I only commented about changes in workerd.ts but those apply to edge-light and index too.

analyze/workerd.ts
e-moran: Nope, there's no need for the `@`, the domain that we're checking against is the one that's returned by the email parser rust lib.

blaine-arcjet (approved):
Excellent! 🎉

PR #1100, opened 2024-07-10
chore(examples): Show dynamic feature flags with LaunchDarkly

bendechrai:
I'd love a review of the approach. Both `arcjet` and `launchdarkly` are singletons and I'm only updating the rules on `arcjet` when I detect a change in rules from LaunchDarkly. I think this is a same implementation...


examples/nodejs-express-launchdarkly/README.md
```diff
@@ -0,0 +1,104 @@
+<a href="https://arcjet.com" target="_arcjet-home">
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="https://arcjet.com/logo/arcjet-dark-lockup-voyage-horizontal.svg">
+    <img src="https://arcjet.com/logo/arcjet-light-lockup-voyage-horizontal.svg" alt="Arcjet Logo" height="128" width="auto">
+  </picture>
+</a>
+
+# Arcjet Dynamic Configuration via LaunchDarkly Feature Flags (with Express for Node.js)
```
blaine-arcjet, line 8:
```suggestion
# Arcjet Dynamic Configuration via LaunchDarkly Feature Flags
```
It makes the permalink worse and doesn't seem worth mentioning when it is already in the description below.

examples/nodejs-express-launchdarkly/lib/arcjet.js
```diff
@@ -0,0 +1,45 @@
...
+// On first load, we'll use the default Arcjet instance without rules
+let ajWithRules = aj;
+
+// We'll store the current LaunchDarkly configuration here
+let currentConfig = {};
+
+// This function will return an Arcjet instance with the latest rules
+const arcjet = async () => {
+  // Get the latest configuration from LaunchDarkly
+  const newConfig = await getArcjetConfig();
```
blaine-arcjet, line 24: This has to fetch the data each time so there's no point in caching the aj client. It is likely that `JSON.stringify` is going to be more costly than using `withRule`

examples/nodejs-express-launchdarkly/lib/arcjet.js
```diff
@@ -0,0 +1,45 @@
...
+  rules: [],
+});
+
+// We don't want to reinitialize Arcjet unless the configuration changes
+// Let's create a global instance of Arcjet with the rules applied
+// On first load, we'll use the default Arcjet instance without rules
+let ajWithRules = aj;
+
+// We'll store the current LaunchDarkly configuration here
+let currentConfig = {};
```
blaine-arcjet, line 19:
```suggestion
```

examples/nodejs-express-launchdarkly/lib/arcjet.js
```diff
@@ -0,0 +1,45 @@
...
+        slidingWindow({
+          mode: currentConfig.slidingWindowMode,
+          max: currentConfig.slidingWindowMax,
+          interval: currentConfig.slidingWindowInterval,
+        })
+      );
+  }
+
+  // Return the Arcjet instance with the latest rules
+  return ajWithRules;
```
blaine-arcjet, line 42:
```suggestion
    // Return the Arcjet instance with the latest rules
    return aj
      .withRule(shield({ mode: currentConfig.shieldMode }))
      .withRule(
        slidingWindow({
          mode: currentConfig.slidingWindowMode,
          max: currentConfig.slidingWindowMax,
          interval: currentConfig.slidingWindowInterval,
        })
      );
```
(the above suggestion needs spacing fixes)

examples/nodejs-express-launchdarkly/lib/launchdarkly.js
```diff
@@ -0,0 +1,37 @@
+import * as ld from "@launchdarkly/node-server-sdk";
+
+// Initialize LaunchDarkly client
+const client = ld.init(process.env.LAUNCHDARKLY_SDK_KEY);
+
+export const getArcjetConfig = async () => {
+  // Wait for the LaunchDarkly client to be initialized
+  await client.waitForInitialization({ timeout: 5000 });
```
blaine-arcjet, line 8: You might want a tighter timeout to be more inline with arcjet SDK (500ms in prod, 1000ms in dev)

blaine-arcjet (changes requested):
I think the arcjet client loading was more complicated than it needed to be. Using `withRule` should be pretty lightweight, especially compared to `JSON.stringify`

examples/nodejs-express-launchdarkly/lib/launchdarkly.js
```diff
@@ -0,0 +1,37 @@
...
+
+export const getArcjetConfig = async () => {
+  // Wait for the LaunchDarkly client to be initialized
+  await client.waitForInitialization({ timeout: 500 });
+
+  // Set the user context for LaunchDarkly - in this example, every user is treated the same.
+  const context = { key: "guest" };
+
+  // Get the latest configuration from LaunchDarkly
+  const shieldMode = await client.variation("shieldMode", context, "LIVE");
```
blaine-arcjet, line 14: Is it worth mentioning that the 3rd argument is the default? Maybe you could make a "defaults" object at the top and pull from it instead of inlining them

examples/nodejs-express-launchdarkly/lib/arcjet.js
```diff
@@ -0,0 +1,28 @@
...
+  return aj.withRule(shield({ mode: currentConfig.shieldMode })).withRule(
+    slidingWindow({
+      mode: currentConfig.slidingWindowMode,
+      max: currentConfig.slidingWindowMax,
+      interval: currentConfig.slidingWindowInterval,
+    })
+  );
+};
+
+export default arcjet;
```
blaine-arcjet, line 28:
```suggestion
import arcjet, { shield, slidingWindow } from "@arcjet/node";
import { getArcjetConfig } from "./launchdarkly.js";

// Initialize Arcjet with your site key and rules
const aj = arcjet({
  // Get your site key from https://app.arcjet.com
  // and set it as an environment variable rather than hard coding.
  // See: https://www.npmjs.com/package/dotenv
  key: process.env.ARCJET_KEY,
  rules: [],
});

// This function will return an Arcjet instance with the latest rules
export default async function() {
  // Get the latest configuration from LaunchDarkly
  const config = await getArcjetConfig();

  // Return the Arcjet instance with the latest rules
  return aj.withRule(shield({ mode: config.shieldMode })).withRule(
    slidingWindow({
      mode: config.slidingWindowMode,
      max: config.slidingWindowMax,
      interval: config.slidingWindowInterval,
    })
  );
}
```
I started making these changes on your blog post but then I thought I should just do it here. Basically, if we are doing a default export, we don't need to name the function so we don't need a special prefix on `arcjet` and I think it looks nicer.

examples/nodejs-express-launchdarkly/lib/arcjet.js
```diff
@@ -0,0 +1,26 @@
+import _arcjet, { shield, slidingWindow } from "@arcjet/node";
+import { getArcjetConfig } from "./launchdarkly.js";
+
+// Initialize Arcjet with your site key and rules
+const aj = _arcjet({
```
blaine-arcjet, line 5:
```suggestion
import arcjet, { shield, slidingWindow } from "@arcjet/node";
import { getArcjetConfig } from "./launchdarkly.js";

// Initialize Arcjet with your site key and rules
const aj = arcjet({
```
Just a nit

blaine-arcjet (changes requested):
@bendechrai Since this is only a nodejs example, you only need to add it to the dependabot config (not the examples CI).

blaine-arcjet (approved):

PR #1176, opened 2024-07-24
fix(analyze): Ensure locally-unknown email information doesn't deny

### LINKED ISSUE #1175: Arcjet incorrectly reports missing MX records for all email addresses
## Description
When using Arcjet for email validation in a newsletter signup form, the system consistently reports that all email addresses are missing MX records, regardless of whether they actually have MX records or not. This behavior is observed both in a custom implementation and on the Arcjet example signup form (https://example.arcjet.com/signup).

## Current Behavior
- All email addresses, including those known to have valid MX records, are flagged as missing MX records.

## Expected Behavior
- The system should correctly identify emails with and without MX records.
- Only email addresses genuinely missing MX records should be flagged as such.

## Steps to Reproduce
1. Implement Arcjet signup form protection as shown in the code snippet below.
2. Submit various email addresses, including those known to have valid MX records.
3. Observe that all submissions result in the "missing MX records" error.

```ts
const aj = arcjet({
  key: process.env.ARCJET_KEY!,
  rules: [
    protectSignup({
      email: {
        mode: 'LIVE',

        block: ['DISPOSABLE', 'INVALID', 'NO_MX_RECORDS'],
      },
      bots: {
        mode: 'LIVE',

        block: ['AUTOMATED'],
      },

      rateLimit: {
        mode: 'LIVE',
        interval: '10m',
        max: 5,
      },
    }),
  ],
})

export async function POST(request: Request) {
  // ... 
  const decision = await aj.protect(request, { email })

  if (decision.isDenied()) {
    if (decision.reason.isEmail()) {
      let message: string
      if (decision.reason.emailTypes.includes('INVALID')) {
        message = 'Your email is invalid.'
      } else if (decision.reason.emailTypes.includes('DISPOSABLE')) {
        message = 'Disposable emails are not allowed.'
      } else if (decision.reason.emailTypes.includes('NO_MX_RECORDS')) {
        message = 'Your email is missing MX records.'
      } else {
        message = 'Your email is just...wrong. So wrong.'
      }

      return NextResponse.json(
        { message, reason: decision.reason },
        { status: 400 },
      )
    }
    // ...
  }
}

```

Any chance there's a setting I've overlooked that's causing this?

I'm using the latest version of Arcjet (as far as I know) with Nextjs. If you need any more info, just give me a shout!

e-moran:
Fixes #1175


blaine-arcjet:
2 bytes seems weird so I'm regenerating again

blaine-arcjet:
I think @e-moran might be using an older version of wasm-tools to generate the component because I'm consistently seeing 1.35kb shaved off our committed files after the recent optimization setting. That would be this PR is adding 0.01kb which is more likely than 2 bytes that it was showing originally.

blaine-arcjet (approved):
Tested this in one of the example apps in the repo and it is working.

PR #1201, opened 2024-07-26
feat: Attempt to warm http2 connection upon SDK startup

blaine-arcjet:
This updates our HTTP2 transport to attempt the `connect()` of the HTTP2 connection upon initialization. It also sets a 600 second timeout on the connection.

This most benefits long-running servers, but improvements can also be seen in warm serverless functions.

### DIFF
--- a/transport/index.ts
+++ b/transport/index.ts
@@ -7,7 +7,7 @@ export function createTransport(baseUrl: string) {
   // We create our own session manager so we can attempt to pre-connect
   const sessionManager = new Http2SessionManager(baseUrl, {
     // AWS ALB doesn't support PING so we use a very high idle timeout
-    idleConnectionTimeoutMs: 4000 * 1000,
+    idleConnectionTimeoutMs: 600 * 1000,
   });
 
   // We ignore the promise result because this is an optimistic pre-connect
--- a/transport/index.ts
+++ b/transport/index.ts
@@ -1,8 +1,21 @@
-import { createConnectTransport } from "@connectrpc/connect-node";
+import {
+  createConnectTransport,
+  Http2SessionManager,
+} from "@connectrpc/connect-node";
 
 export function createTransport(baseUrl: string) {
+  // We create our own session manager so we can attempt to pre-connect
+  const sessionManager = new Http2SessionManager(baseUrl, {
+    // AWS ALB doesn't support PING so we use a very high idle timeout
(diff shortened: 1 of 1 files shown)


blaine-arcjet:
I tested Bun and it doesn't emit any events on the HTTP/2 stream. I spent the afternoon trying to dig into their source code and compare the `@grpc/grpc-js` implementation (which they claim works) with connectrpc and can't figure it out. So I just reverted the bun adapter back to HTTP/1.1 until they make it work correctly.

davidmytton:
Are we not supporting Next.js with this as well?

blaine-arcjet:
> Are we not supporting Next.js with this as well?

Looking at our Next adapter, it uses `@connectrpc/connect-web` dependency. The comment states:
```
// Transport is the HTTP client that the client uses to make requests.
// The Connect Node client doesn't work on edge runtimes: https://github.com/bufbuild/connect-es/pull/589
// so set the transport using connect-web. The interceptor is required for it work in the edge runtime.
```

Node imports are forbidden in the edge runtime, so we'd need to implement near-duplicate files like we do for `@arcjet/analyze` to support different transports inside one package.

davidmytton:
Ah yes, I remember. We definitely want to do this for Next.js as well - do you want to do that as a separate PR?

blaine-arcjet:
@davidmytton this should be ready to review

davidmytton (approved):
Looks good. I've tested Next.js on Vercel with edge middleware and serverless functions. Serverless functions show the expected behavior of the initial connection latency being the highest and then subsequent requests being much faster. There wasn't any improvement in the latency on the edge middleware - latency is the same for every request.

transport/index.ts
```diff
@@ -1,8 +1,21 @@
...
+import {
+  createConnectTransport,
+  Http2SessionManager,
+} from "@connectrpc/connect-node";
 
 export function createTransport(baseUrl: string) {
+  // We create our own session manager so we can attempt to pre-connect
+  const sessionManager = new Http2SessionManager(baseUrl, {
+    // AWS ALB doesn't support PING so we use a very high idle timeout
+    idleConnectionTimeoutMs: 4000 * 1000,
```
blaine-arcjet, line 10:
```suggestion
    idleConnectionTimeoutMs: 600 * 1000,
```

PR #1221, opened 2024-07-30
feat: Abstract transports into package to leverage conditional exports

blaine-arcjet:
This moves our transport logic into its own package. This allows us to leverage [conditional exports](https://nodejs.org/docs/latest-v18.x/api/packages.html#conditional-exports) to support `bun` and `edge-light` which need additional considerations for their transports.

Right now, this package is very thin, as it just calls out to the correct Connect-RPC functions for the correct runtime, but we'll also be able to consolidate our HTTP/2 Session logic from #1201 and build our custom transport in a single place (in the future).

During development, I added console.log statements to verify the correct `createTransport` was being called based on the runtime, but I don't want to thread the logger into here so I just removed them before submitting the PR.


README.md
```diff
@@ -166,8 +166,10 @@ find a specific one through the categories and descriptions below.
 ### Utilities
 
 - [`arcjet`](./arcjet/README.md): JS SDK core.
-- [`@arcjet/protocol`](./protocol/README.md): JS interface
-  into the Arcjet protocol.
+- [`@arcjet/protocol`](./protocol/README.md): JS interface into the Arcjet
+  protocol.
+- [`@arcjet/transport](./transport/README.md): Transport mechanisms for the
```
blaine-arcjet, line 171:
```suggestion
- [`@arcjet/transport`](./transport/README.md): Transport mechanisms for the
```

davidmytton (approved):

Issue #928, opened 2024-06-11
Pre-beta changes


blaine-arcjet:
- [x] https://github.com/arcjet/arcjet-js/issues/46
- [x] https://github.com/arcjet/arcjet-js/issues/191
- [x] https://github.com/arcjet/arcjet-js/issues/922 (only technically breaking since it could be deduped into the tree)
- [x] https://github.com/arcjet/arcjet-js/issues/930
- [x] https://github.com/arcjet/arcjet-js/issues/1397
- [x] https://github.com/arcjet/arcjet-js/issues/992
- [x] https://github.com/arcjet/arcjet-js/issues/1448
- [x] https://github.com/arcjet/arcjet-js/issues/1391
- [x] https://github.com/arcjet/arcjet-js/issues/1801
- [x] https://github.com/arcjet/arcjet-js/issues/1810
- [x] https://github.com/arcjet/arcjet-js/issues/1834
- [x] https://github.com/arcjet/arcjet-js/issues/1835
- [x] https://github.com/arcjet/arcjet-js/issues/1836
- [x] Connect RPC & Buf v2 upgrades
  - [x] https://github.com/arcjet/arcjet-js/pull/2133
  - [x] https://github.com/arcjet/arcjet-js/pull/2286
  - [x] https://github.com/arcjet/arcjet-js/pull/2288
  - [x] https://github.com/arcjet/arcjet-js/pull/2290

davidmytton:
All these make sense to me except #539 unless we also do #782. eslint 9 has version conflicts with other packages so we should avoid forcing a breaking change release just for our internal packages.

blaine-arcjet:
I'm fine to skip the eslint upgrade but I think we're in agreement to decouple the version numbers of our devDeps so we can make breaking changes in the future without affecting the SDK version numbers.

davidmytton:
Yes, agreed.

blaine-arcjet:
We're putting this on hold until we're ready to approach beta again.

e-moran:
We should rename the `block` field in `validateEmail` to align with the allow/deny naming of other rules.

blaine-arcjet:
Remove the ability to construct multiple rules with one constructor: #1397

blaine-arcjet:
@e-moran I think we need to make this backwards compatible with warning logs for deprecated format. Would you be able to do that in the timeline suggested?

blaine-arcjet:
[...]

PR #1300, opened 2024-08-08
feat: add detect sensitive info rule

e-moran:
Adds a new rule that can be used to detect sensitive information being sent in a request when it isn't expected.


analyze/convert.d.ts
blaine-arcjet, line 1: Need to gitignore this file

analyze/convert.js
blaine-arcjet, line 1: Need to gitignore this file

protocol/index.ts
```diff
@@ -50,6 +50,24 @@ export const ArcjetEmailType: ArcjetEnum<ArcjetEmailType> = {
...
+export type IdentifiedEntity<Custom extends string> = {
+  start: number;
+  end: number;
+  identifiedType: SensitiveInfoEntity<Custom>;
+};
+
+export type DetectSensitiveInfoResult<Custom extends string> = {
+  allowed: IdentifiedEntity<Custom>[];
+  denied: IdentifiedEntity<Custom>[];
+};
```
blaine-arcjet, line 69: These aren't user-facing types so they shouldn't exist in the protocol. There's also no real reason to be creating new types for analyze when the generated types work pretty well.

protocol/index.ts
```diff
@@ -108,6 +137,25 @@ export class ArcjetReason {
   }
 }
 
+export class ArcjetSensitiveInfoReason<
+  Custom extends string,
+> extends ArcjetReason {
+  type = "SENSITIVE_INFO" as const;
+
+  denied: IdentifiedEntity<Custom>[];
+  allowed: IdentifiedEntity<Custom>[];
```
blaine-arcjet, line 146: I'm not sure we actually need IdentifiedEntity type on the reason.

protocol/index.ts
```diff
@@ -729,6 +777,26 @@ export interface ArcjetEmailRule<Props extends { email: string }>
...
+export type CustomDetect<Custom extends string> = (
+  tokens: string[],
+) => (SensitiveInfoEntity<Custom> | undefined)[];
+
+export type SensitiveInfoConfig<Custom extends string> = {
+  allow?: SensitiveInfoEntity<Custom>[];
+  deny?: SensitiveInfoEntity<Custom>[];
+  customDetect?: CustomDetect<Custom>;
+  contextWindowSize: number;
+};
```
blaine-arcjet, line 789: These aren't user-facing types so they shouldn't exist in the protocol. There's also no real reason to be creating new types for analyze when the generated types work pretty well.

protocol/index.ts
```diff
@@ -729,6 +777,26 @@ export interface ArcjetEmailRule<Props extends { email: string }>
...
+
+export interface ArcjetSensitiveInfoRule<
+  Props extends {},
+  Custom extends string,
+> extends ArcjetLocalRule<Props> {
+  type: "SENSITIVE_INFO";
+
+  options: SensitiveInfoConfig<Custom>;
```
blaine-arcjet, line 797: Removed in my PR because you don't need to hold onto the config on the rule. The properties on other rules are due to remote serialization.

protocol/index.ts
```diff
@@ -758,11 +826,14 @@ export interface ArcjetLogger {
   error(obj: Record<string, unknown>, msg?: string, ...args: unknown[]): void;
 }
 
+export type GetBody = () => Promise<string>;
```
blaine-arcjet, line 829: Why is this a separate type? Defining it on the ArcjetContext should be enough

analyze/convert.ts
blaine-arcjet, line 1: I question whether we actually need these converter functions. Would it be too much of an ask for the custom `detect` function to produce the objects, like `{ tag: "email" }` instead of returning `"email"`?

analyze/edge-light.ts
```diff
@@ -1,4 +1,10 @@
-import type { ArcjetLogger, ArcjetRequestDetails } from "@arcjet/protocol";
+import type {
+  ArcjetLogger,
+  ArcjetRequestDetails,
+  CustomDetect,
+  DetectSensitiveInfoResult,
+  SensitiveInfoConfig,
```
blaine-arcjet, line 6: Shouldn't pull types from protocol. Analyze is mostly a standalone API that we call into. We probably shouldn't even rely on the ArcjetLogger or ArcjetRequestDetails types.

analyze/edge-light.ts
```diff
@@ -12,6 +18,10 @@ import type {
 import componentCoreWasm from "./wasm/arcjet_analyze_js_req.component.core.wasm?module";
 import componentCore2Wasm from "./wasm/arcjet_analyze_js_req.component.core2.wasm?module";
 import componentCore3Wasm from "./wasm/arcjet_analyze_js_req.component.core3.wasm?module";
+import {
+  ConvertAnalyzeEntitiesToProtocolEntities,
+  ConvertProtocolEntitiesToAnalyzeEntities,
+} from "./convert";
```
blaine-arcjet, line 24: See above

analyze/edge-light.ts
```diff
@@ -40,9 +50,22 @@ async function moduleFromPath(path: string): Promise<WebAssembly.Module> {
...
+async function init<Custom extends string>(
+  context: AnalyzeContext,
+  detect?: CustomDetect<Custom>,
+) {
   const { log } = context;
 
+  const convertedDetect = (tokens: string[]) => {
+    if (detect !== undefined) {
+      return detect(tokens).map(
+        (e) => e && ConvertProtocolEntitiesToAnalyzeEntities,
```
blaine-arcjet, line 62: See above. In my local version, I replaced this completely with a `typeof detect === "function" ? detect : noopDetect`

analyze/edge-light.ts
```diff
@@ -161,3 +187,56 @@ export async function detectBot(
...
+  let entities: core.Entities = {
+    tag: "allow",
+    val: options?.allow?.map(ConvertProtocolEntitiesToAnalyzeEntities) || [],
+  };
+  if (options?.deny) {
+    entities = {
+      tag: "deny",
+      val: options.deny.map(ConvertProtocolEntitiesToAnalyzeEntities),
+    };
+  }
```
blaine-arcjet, line 206: We probably shouldn't care to try to map here and instead should do any transformations in the local rule itself. I've tried to make Analyze mostly just map to WIT after it does the Wasm initialization.

analyze/edge-light.ts
```diff
@@ -161,3 +187,56 @@ export async function detectBot(
...
+    return {
+      allowed,
+      denied,
+    };
+  } else {
+    // Skip the local evaluation of the rule if WASM is not available
+    return {
+      allowed: [],
+      denied: [],
+    };
```
blaine-arcjet, line 240: I'm wondering if this is really bad for a local-only rule. Perhaps we want to throw?

analyze/index.ts
blaine-arcjet, line 1: See edge-light comments

analyze/workerd.ts
blaine-arcjet, line 1: See edge-light comments

arcjet-next/index.ts
```diff
@@ -124,6 +124,18 @@ export interface ArcjetNextRequest {
...
 
+interface ArcjetNextAppRequest extends ArcjetNextRequestCommon {
+  clone: () => Request;
+  body: ReadableStream<Uint8Array> | null;
+}
+
+interface ArcjetNextApiRequest extends ArcjetNextRequestCommon {
+  body: any;
+  clone?: never;
+}
```
blaine-arcjet, line 135: I don't like having these separate interfaces. Instead, I made the ArcjetNextRequest extremely generic and you have to figure out what is available.

arcjet-node/index.ts
```diff
@@ -93,6 +93,8 @@ export interface ArcjetNodeRequest {
   method?: string;
   httpVersion?: string;
   url?: string;
+  clone(): ArcjetNodeRequest;
+  text(): Promise<string>;
```
blaine-arcjet, line 97: Everything here is optional. We have to double check every property.

arcjet/index.ts
```diff
@@ -390,6 +418,7 @@ export type ArcjetRequest<Props extends PlainObject> = Simplify<
     headers?: Headers | Record<string, string | string[] | undefined>;
     cookies?: string;
     query?: string;
+    body?: string;
```
blaine-arcjet, line 421: Not using this, correct? We're using getBody instead.

arcjet/index.ts
```diff
@@ -523,6 +552,86 @@ export function slidingWindow<
   return rules;
 }
 
+export function sensitiveInfo<Custom extends string>(
+  options: SensitiveInfoOptions<Custom>,
+  ...additionalOptions: SensitiveInfoOptions<Custom>[]
+): Primitive<{}> {
+  const rules: ArcjetSensitiveInfoRule<{}, Custom>[] = [];
+
+  // Always create at least one EMAIL rule
```
blaine-arcjet, line 561: not email

arcjet/index.ts
```diff
@@ -523,6 +552,86 @@ export function slidingWindow<
...
+  for (const opt of [
+    options ?? ({} as SensitiveInfoOptions<Custom>),
+    ...additionalOptions,
+  ]) {
+    const mode = opt.mode === "LIVE" ? "LIVE" : "DRY_RUN";
+    // TODO: Filter invalid email types (or error??)
```
blaine-arcjet, line 567: not email

arcjet/index.ts
```diff
@@ -523,6 +552,86 @@ export function slidingWindow<
...
+
+    const redactOpts = {
+      allow: options?.allow || [],
+      deny: options?.deny || [],
+      contextWindowSize: opt.contextWindowSize || 1,
+      customDetect: options?.detect,
```
blaine-arcjet, line 573:
```suggestion
      detect: options?.detect,
```
Can just be named `detect` here too, right?

arcjet/index.ts
```diff
@@ -523,6 +552,86 @@ export function slidingWindow<
...
+    };
+
+    rules.push({
+      type: "SENSITIVE_INFO",
+      priority: Priority.EmailValidation,
+      mode,
+      options: redactOpts,
```
blaine-arcjet, line 580: removed in my PR

arcjet/index.ts
```diff
@@ -911,6 +1020,7 @@ export default function arcjet<
       // body: request.body,
       extra: extraProps(request),
       email: typeof request.email === "string" ? request.email : undefined,
+      body: request.body,
```
blaine-arcjet, line 1023: What's happening here?

blaine-arcjet (changes requested):
Some quick feedback. Chatted with @e-moran about some of it and he said he'll move stuff around and cleanup.

analyze/convert.ts
e-moran: Discussed in slack, agreed to keep it like this to provide a better UX but the convertor functions are now part of the SDK.

body/README.md
```diff
@@ -0,0 +1,56 @@
...
+## Implementation
+
+The implementation of this library is based on the [raw-body][node-raw-body]
+package. Both licensed MIT with licenses included in our source code.
+
+We've chosen to re-implement the logic to read the body from the stream to keep
+the dependency tree for our packages as light as possible. Our implementation is
+as light as possible while still providing the functionality that we need, specifically
+it excludes the functionality to return the stream as a buffer and always parses it
+as a string.
```
blaine-arcjet, line 49: And removes encoding. And only provides an async interface. 😛

body/index.ts
```diff
@@ -0,0 +1,102 @@
+import { Readable } from "stream";
+
+export type GetBodyOpts = {
+  encoding: string;
```
blaine-arcjet, line 4: Needed?

body/index.ts
```diff
@@ -0,0 +1,102 @@
...
+  limit?: number;
```
blaine-arcjet, line 5: Required?

body/index.ts
```diff
@@ -0,0 +1,102 @@
...
+  expectedLength?: number;
+};
+
+export function getBody(stream: Readable, opts: GetBodyOpts): Promise<string> {
+  return new Promise((resolve, reject) => {
+    getBodySync(stream, opts, (err?: Error, buffer?: string) => {
```
blaine-arcjet, line 11: Do we really need to get it sync inside of this async function? I feel like that might just be a shortcut they used but I'm not done reading the implementation.

body/index.ts
```diff
@@ -0,0 +1,102 @@
...
+
+  function onData(chunk: Buffer) {
+    if (complete) return;
+
+    received += chunk.length;
+
+    if (received > limit) {
+      done(new Error("request entity too large"));
+    } else {
+      buffer += decoder.decode(chunk);
```
blaine-arcjet, line 78: TextDecoder can be used in streaming mode: https://nodejs.org/docs/latest-v20.x/api/util.html#textdecoderdecodeinput-options

I think this allows the data to have invalid boundaries; though I don't see this being checked for

body/index.ts
```diff
@@ -0,0 +1,102 @@
...
+      reject(new Error("body reader is in an unexpected state"));
+    });
+  });
+}
+
+type StreamReadCallback = (err?: Error, data?: string) => void;
+export function getBodySync(
+  stream: Readable,
+  opts: GetBodyOpts,
+  callback: StreamReadCallback,
```
blaine-arcjet, line 23: Do we want to keep the callback or would making this promise-returning work better? Unfortunately, we can't make this async and then throw inside the event handlers because the domain on the stream will catch it and tear everything down.

body/test/body.test.ts
blaine-arcjet, line 1: Need to review still

examples/nextjs-14-app-dir-rl/middleware.ts
blaine-arcjet, line 1: please undo changes to this file

examples/nextjs-14-app-dir-si/.env.local.example
blaine-arcjet, line 1: maybe just call this example `nextjs-14-sensitive-info`. I don't think the app vs pages next stuff needs to be there and we want the names to be more clear in the directory listing.

You'll also need to add the example to the dependabot and workflow files.

examples/nextjs-14-app-dir-si/README.md
```diff
@@ -0,0 +1,37 @@
...
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="https://arcjet.com/logo/arcjet-dark-lockup-voyage-horizontal.svg">
+    <img src="https://arcjet.com/logo/arcjet-light-lockup-voyage-horizontal.svg" alt="Arcjet Logo" height="128" width="auto">
+  </picture>
+</a>
+
+# Arcjet Block Sensitive Info with Next.js 14 using the App Router
+
+This example shows how to use Arcjet with a Next.js [route
+handler](https://nextjs.org/docs/app/building-your-application/routing/route-handlers).
```
blaine-arcjet, line 11:
```suggestion
This example shows how to use Arcjet's Sensitive Info rule with a Next.js [route
handler](https://nextjs.org/docs/app/building-your-application/routing/route-handlers) to catch any personally identifying information before executing your business logic.
```
Not sure how this should be worded. cc @bendechrai @davidmytton

examples/nextjs-14-app-dir-si/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+3. Rename `.env.local.example` to `.env.local` and add your Arcjet key.
+
+4. Start the dev server.
+
+   ```bash
+   npm run dev
+   ```
+
+5. Curl `http://localhost:3000/api/arcjet` with some data
```
blaine-arcjet, line 36: Might want an example here.

examples/nextjs-14-app-dir-si/README.md
```diff
@@ -0,0 +1,37 @@
...
+6. If the data you sent contains any blocked types then the api will return a 400.
```
blaine-arcjet, line 37:
```suggestion
6. If the data you sent contains any blocked types then the route will return a 400.
```

examples/nextjs-14-app-dir-si/app/api/arcjet/route.ts
```diff
@@ -0,0 +1,37 @@
...
+        error: "Sensitive Information Identified",
+        reason: decision.reason,
+      },
+      {
+        status: 400,
+      },
+    );
+  }
+
+  console.log(await req.text());
```
blaine-arcjet, line 34: Perhaps add the non-identifying information into the message instead of the log

examples/nextjs-14-app-dir-si/middleware.ts
blaine-arcjet, line 1: You can remove the middleware matcher for this example

examples/nextjs-14-app-dir-si/package.json
```diff
@@ -0,0 +1,29 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "scripts": {
+    "dev": "next dev",
+    "build": "next build",
+    "start": "next start",
+    "lint": "next lint"
+  },
+  "dependencies": {
+    "@arcjet/env": "file:../../env",
```
blaine-arcjet, line 12: Why is this being pulled in?

examples/nextjs-14-app-dir-si/package.json
```diff
@@ -0,0 +1,29 @@
+{
+  "name": "nextjs-14-app-dir-rl",
```
blaine-arcjet, line 2: Wrong name. Change it with renaming the example. Don't forget to regenerate package-lcok

examples/nodejs-express-rl/index.js
blaine-arcjet, line 1: Please undo the change to this file.

examples/nodejs-express-si/.env.local.example
blaine-arcjet, line 1: Change the name of the example to `nodejs-express-sensitive-info` or maybe just `express-sensitive-info`

examples/nodejs-express-si/README.md
```diff
@@ -0,0 +1,42 @@
...
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="https://arcjet.com/logo/arcjet-dark-lockup-voyage-horizontal.svg">
+    <img src="https://arcjet.com/logo/arcjet-light-lockup-voyage-horizontal.svg" alt="Arcjet Logo" height="128" width="auto">
+  </picture>
+</a>
+
+# Arcjet Rate Limit with Express for Node.js
+
+This example shows how to use Arcjet with a Node.js
+[Express](https://expressjs.com/) server.
```
blaine-arcjet, line 11: Better description based on the changed description from the other example

examples/nodejs-express-si/README.md
```diff
@@ -0,0 +1,42 @@
+<a href="https://arcjet.com" target="_arcjet-home">
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="https://arcjet.com/logo/arcjet-dark-lockup-voyage-horizontal.svg">
+    <img src="https://arcjet.com/logo/arcjet-light-lockup-voyage-horizontal.svg" alt="Arcjet Logo" height="128" width="auto">
+  </picture>
+</a>
+
+# Arcjet Rate Limit with Express for Node.js
```
blaine-arcjet, line 8: Wrong title

examples/nodejs-express-si/README.md
```diff
@@ -0,0 +1,42 @@
...
+   npm start
+   ```
+
+   This assumes you're using Node.js 20 or later because the `start` script
+   loads a local environment file with `--env-file`. If you're using an older
+   version of Node.js, you can use a package like
+   [dotenv](https://www.npmjs.com/package/dotenv) to load the environment file.
+
+5. Visit `http://localhost:3000/`.
+6. Refresh the page to trigger the rate limit.
```
blaine-arcjet, line 42: Wrong last steps?

examples/nodejs-express-si/index.js
```diff
@@ -0,0 +1,41 @@
...
+
+app.post('/', async (req, res) => {
+  const decision = await aj.protect(req);
+
+  if (decision.isDenied()) {
+    res.writeHead(429, { "Content-Type": "application/json" });
+    res.end(JSON.stringify({ error: "Too Many Requests" }));
+  } else {
+    res.writeHead(200, { "Content-Type": "application/json" });
+    res.end(JSON.stringify({ message: "Hello World" }));
```
blaine-arcjet, line 33: Probably use the body in this message

.github/release-please-config.json
blaine-arcjet, line 1: This also needs to be added to `.github/.release-please-manifest.json` or it will get confused.

protocol/index.ts
```diff
@@ -50,6 +50,17 @@ export const ArcjetEmailType: ArcjetEnum<ArcjetEmailType> = {
   INVALID: "INVALID",
 };
 
+export type IdentifiedEntity = {
```
blaine-arcjet, line 53:
```suggestion
export type ArcjetIdentifiedEntity = {
```

protocol/index.ts
```diff
@@ -50,6 +50,17 @@ export const ArcjetEmailType: ArcjetEnum<ArcjetEmailType> = {
...
+  start: number;
+  end: number;
+  identifiedType: string;
+};
+
+export type DetectSensitiveInfoResult = {
```
blaine-arcjet, line 59: Where is this used? We should probably remove it.

protocol/index.ts
```diff
@@ -665,7 +716,7 @@ export interface ArcjetRequestDetails {
 }
 
 export type ArcjetRule<Props extends {} = {}> = {
-  type: "RATE_LIMIT" | "BOT" | "EMAIL" | "SHIELD" | string;
+  type: "RATE_LIMIT" | "BOT" | "EMAIL" | "SHIELD" | "DETECT_PII" | string;
```
blaine-arcjet, line 719: Why isn't this named `"SENSITIVE_INFO"`?

protocol/test/client.test.ts
blaine-arcjet, line 1: I believe this needs tests to verify that the SensitiveInfo rule/reason get sent to the backend, right?

blaine-arcjet (changes requested):
This PR is looking excellent. Incredible work @e-moran 💪

There looks like a lot of feedback but most of it is moving stuff around or trimming implementation details. There were a couple of bigger things I flagged that might need some discussion.

I also skipped over the body test until we shake out the API surface.

arcjet/index.ts
```diff
@@ -309,11 +312,51 @@ export type EmailOptions = {
...
+> =
+  | SensitiveInfoOptionsAllow<Detect, CustomEntities>
+  | SensitiveInfoOptionsDeny<Detect, CustomEntities>;
+
 const Priority = {
   Shield: 1,
   RateLimit: 2,
   BotDetection: 3,
   EmailValidation: 4,
+  SensitiveInfo: 5,
```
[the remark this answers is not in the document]
davidmytton: If we have sensitive info then we don't want to do any further processing - the request is likely to be blocked, so we can save time especially when this can be done locally. So I'd move it to `SensitiveInfo: 1`

arcjet-node/index.ts
```diff
@@ -255,7 +259,35 @@ export default function arcjet<
           ExtraProps<Rules>
         >;
 
-        return aj.protect({}, req);
+        const getRequestBody = async () => {
+          try {
+            // If request.body is present then the body was likely read by a package like express' `body-parser`.
+            // If it's not present then we attempt to read the bytes from the IncomingMessage ourselves.
+            if (request.body) {
+              return JSON.stringify(request.body);
```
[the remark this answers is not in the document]
e-moran: Because its an `any` depending on which middleware you used to create it. The most common cases would be JSON or formurlencoded bodies which both get converted to objects which we'd need to stringify. An alternative would be to iterate through the values, check if they are strings and if so then redact those.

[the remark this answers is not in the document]
blaine-arcjet: I think we need to use `typeof` to check it then. If `typeof request.body === "string"` we don't need to serialize it. If `typeof` object we try to serialize, if `typeof` number, I suppose we also have to stringify (??)

Using `typeof` in TS is your best friend because it narrows the type on both side of the if-statement

arcjet/index.ts
```diff
@@ -523,6 +568,163 @@ export function slidingWindow<
...
+      type: "SENSITIVE_INFO",
+      priority: Priority.SensitiveInfo,
+      mode,
+      allow: options.allow || [],
+      deny: options.deny || [],
+
+      validate(
+        context: ArcjetContext,
+        details: ArcjetRequestDetails,
+      ): asserts details is ArcjetRequestDetails {},
```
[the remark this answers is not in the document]
e-moran: Yeah I had originally wanted to put it there, but ran in to that issue 😞

examples/nextjs-14-app-dir-si/package.json
e-moran: Good question, I copied the deps from nextjs-14-app-dir-rl and it was there. But doesn't seem to be needed.

arcjet-node/index.ts
```diff
@@ -255,7 +259,35 @@ export default function arcjet<
           ExtraProps<Rules>
         >;
 
-        return aj.protect({}, req);
+        const getRequestBody = async () => {
```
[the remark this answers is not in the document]
e-moran: That won't work because we import `getBody` from the body package. Could change its naming though.

body/index.ts
e-moran: Not exactly, but I think it feels strange to control the limit inside of body

examples/nextjs-14-app-dir-si/package.json
e-moran: We should probably remove it from that package too, but out of scope for this PR

arcjet/index.ts
```diff
@@ -309,11 +312,51 @@ export type EmailOptions = {
...
+) => (ArcjetSensitiveInfoType | T | undefined)[];
+
+type SensitiveInfoOptionsAllow<
+  Detect extends DetectEntities<CustomEntities>,
+  CustomEntities extends string,
+> = {
+  allow: (
+    | ArcjetSensitiveInfoType
+    | Exclude<ReturnType<Detect>[number], undefined>
+  )[];
```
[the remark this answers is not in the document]
e-moran: Not sure what `Simplify<>` around it does for us here, adding it makes it not be assignable to `string[]` though.

e-moran:
@blaine-arcjet That latest push should address _almost_ everything, its missing a few docs changes and adding tests for testing that we report to the server correctly. It's also failing to run `npm ci`... despite no package-lock changes happening when I run `npm install` locally. I'll look in to all of these tomorrow. But if you get the chance to look over it and let me know if anything else needs my attention that would be great!

e-moran (commented):
Something strange is happening with the github diff ui. If you view the file in the branch it doesn't have the extra newline.

examples/nextjs-14-app-dir-rl/middleware.ts
```diff
@@ -18,4 +18,4 @@ const aj = arcjet({
   ],
 });
 
-export default createMiddleware(aj);
\ No newline at end of file
+export default createMiddleware(aj);
```
e-moran, line 21: Something strange is happening with the github diff ui. If you view the file in the branch it doesn't have the extra newline.

arcjet-bun/index.ts
```diff
@@ -235,7 +235,17 @@ export default function arcjet<
           ExtraProps<Rules>
         >;
 
-        return aj.protect({}, req);
+        const getBody = async () => {
+          try {
+            const clonedRequest = request.clone();
+            return clonedRequest.text();
+          } catch (e) {
+            log.error("failed to get request body", e);
```
blaine-arcjet, line 243: This is the wrong way to use the logger. The string needs to have sprintf-style formatting and the error needs to be turned into a string. I'll push the changes.

arcjet-next/index.ts
```diff
@@ -327,7 +330,21 @@ export default function arcjet<
...
+        const getBody = async () => {
+          try {
+            if (request.clone !== undefined) {
+              const cloned = request.clone();
+              return await cloned.text();
+            } else if (request.body !== undefined) {
+              return JSON.stringify(request.body);
+            }
+          } catch (e) {
+            log.error("failed to get request body", e);
```
blaine-arcjet, line 342: Same issue. I'll push.

arcjet-next/index.ts
```diff
@@ -327,7 +330,20 @@ export default function arcjet<
...
         >;
 
-        return aj.protect({}, req);
+        const getBody = async () => {
+          try {
+            if (request.clone !== undefined) {
+              const cloned = request.clone();
+              return await cloned.text();
+            } else if (request.body !== undefined) {
+              return JSON.stringify(request.body);
```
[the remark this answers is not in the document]
blaine-arcjet: @e-moran this needs an answer

arcjet-next/index.ts
```diff
@@ -327,7 +330,21 @@ export default function arcjet<
           ExtraProps<Rules>
         >;
 
-        return aj.protect({}, req);
+        const getBody = async () => {
+          try {
+            if (request.clone !== undefined) {
```
blaine-arcjet, line 335: typeof

arcjet-next/index.ts
```diff
@@ -327,7 +330,21 @@ export default function arcjet<
...
+              const cloned = request.clone();
+              return await cloned.text();
+            } else if (request.body !== undefined) {
```
blaine-arcjet, line 338: typeof

arcjet-sveltekit/index.ts
```diff
@@ -231,7 +231,16 @@ export default function arcjet<
           ExtraProps<Rules>
         >;
 
-        return aj.protect({}, req);
+        const getBody = async () => {
+          try {
+            const clonedRequest = request.request.clone();
```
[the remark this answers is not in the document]
blaine-arcjet: @e-moran still needs a response

arcjet/index.ts
```diff
@@ -524,6 +570,180 @@ export function slidingWindow<
...
+        }
+
+        if (Array.isArray(opt.deny)) {
+          entitiesTag = "deny";
+          entitiesVal = opt.deny
+            .filter((e) => typeof e !== "undefined")
+            .map(protocolEntitiesToAnalyze);
+        }
+
+        const entities = {
```
blaine-arcjet, line 707: I was looking at the naming around these WIT types and it is really generic. We need to fix them before we ship this as they overlap with so much else we want to define.

body/index.ts
blaine-arcjet, line 1: I streamlined this file. It still needs the original license/copyright and explanation as a comment in the file.

body/test/body.test.ts
blaine-arcjet, line 1: I reworked these tests as they were just hanging with the change away from sync.

@e-moran If you run `npm test -w body`, you'll notice that you need to improve coverage a lot. We target 100% coverage on any libraries we release. We also try to hit 100% on arcjet core but it slips a bit with conversion functions, etc.

examples/express-sensitive-info/README.md
```diff
@@ -0,0 +1,43 @@
...
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="https://arcjet.com/logo/arcjet-dark-lockup-voyage-horizontal.svg">
+    <img src="https://arcjet.com/logo/arcjet-light-lockup-voyage-horizontal.svg" alt="Arcjet Logo" height="128" width="auto">
+  </picture>
+</a>
+
+# Arcjet Sensitive Info Detection with Express for Node.js
+
+This example shows how to use Arcjet with a Node.js
+[Express](https://expressjs.com/) server.
```
blaine-arcjet, line 11: Lost my comment about describing better.

examples/express-sensitive-info/README.md
```diff
@@ -0,0 +1,43 @@
...
+   ```
+
+   This assumes you're using Node.js 20 or later because the `start` script
+   loads a local environment file with `--env-file`. If you're using an older
+   version of Node.js, you can use a package like
+   [dotenv](https://www.npmjs.com/package/dotenv) to load the environment file.
+
+4. Execute to POST request to `http://localhost:3000/` without any sensitive data.
+5. Execute to POST request to `http://localhost:3000/` with some blocked entities in the body
+   and the request should fail.
```
blaine-arcjet, line 43: With curl? Show it.

examples/express-sensitive-info/index.js
```diff
@@ -0,0 +1,41 @@
...
+
+app.post('/', async (req, res) => {
+  const decision = await aj.protect(req);
+
+  if (decision.isDenied() && decision.reason.isSensitiveInfo()) {
+    res.writeHead(400, { "Content-Type": "application/json" });
+    res.end(JSON.stringify({ error: "Too Many Requests", denied: decision.reason.denied }));
+  } else {
+    res.writeHead(200, { "Content-Type": "application/json" });
+    res.end(JSON.stringify({ message: "Hello World" }));
```
blaine-arcjet, line 33: How do I access the body here if I wanted to say something like:
```suggestion
    res.end(JSON.stringify({ message: `You said: ${body}` }));
```

examples/express-sensitive-info/package.json
blaine-arcjet, line 1: Still not added to the dependabot config, I believe

examples/nextjs-14-app-dir-si/README.md
blaine-arcjet, line 1: This project didn't get renamed

examples/nextjs-14-app-dir-si/app/api/arcjet/route.ts
```diff
@@ -0,0 +1,35 @@
...
+        error: "Sensitive Information Identified",
+        reason: decision.reason,
+      },
+      {
+        status: 400,
+      },
+    );
+  }
+
+  return NextResponse.json({ message: "Hello World", reason: decision.reason });
```
blaine-arcjet, line 34:
```suggestion
  return NextResponse.json({ message: `You said: ${await req.text()}` });
```
I meant something like this. We shouldn't send the reason back to users

examples/nextjs-14-app-dir-si/package.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "nextjs-14-app-dir-si",
```
blaine-arcjet, line 2: Rename project directory and change name here and regenerate package lock.

protocol/test/client.test.ts
blaine-arcjet, line 1: I thought I commented on this file that we need client tests for sending the new rule.

examples/express-sensitive-info/index.js
```diff
@@ -0,0 +1,41 @@
...
+    }),
+  ],
+});
+
+app.post('/', async (req, res) => {
+  const decision = await aj.protect(req);
+
+  if (decision.isDenied() && decision.reason.isSensitiveInfo()) {
+    res.writeHead(400, { "Content-Type": "application/json" });
+    res.end(JSON.stringify({ error: "Too Many Requests", denied: decision.reason.denied }));
```
blaine-arcjet, line 30:
```suggestion
    res.end(JSON.stringify({ error: "Sensitive Information Detected", denied: decision.reason.denied }));
```

body/README.md
blaine-arcjet: @e-moran this wasn't addressed

analyze/edge-light.ts
```diff
@@ -161,3 +177,27 @@ export async function detectBot(
     };
   }
 }
+export async function detectSensitiveInfo(
+  context: AnalyzeContext,
+  candidate: string,
+  entities: core.Entities,
+  contextWindowSize: number,
+  detect: typeof ArcjetJsReqSensitiveInformationIdentifier.detect,
```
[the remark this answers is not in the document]
blaine-arcjet: I gave a bad/generic name. I updated to be something more specific.

arcjet-node/index.ts
```diff
@@ -255,7 +259,35 @@ export default function arcjet<
...
+          try {
+            // If request.body is present then the body was likely read by a package like express' `body-parser`.
+            // If it's not present then we attempt to read the bytes from the IncomingMessage ourselves.
+            if (request.body) {
+              return JSON.stringify(request.body);
+            }
+            if (request instanceof IncomingMessage) {
+              const expectedLengthStr = request.headers["content-length"];
```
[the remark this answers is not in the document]
blaine-arcjet: Found a problem with accessing ArcjetHeaders but left a TODO.

arcjet-node/index.ts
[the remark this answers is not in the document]
blaine-arcjet: Renamed

arcjet/index.ts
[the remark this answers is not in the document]
blaine-arcjet: #1391

arcjet/index.ts
```diff
@@ -523,6 +568,163 @@ export function slidingWindow<
...
+          options?.allow
+            ?.filter((e) => e !== undefined)
+            .map(protocolEntitiesToAnalyze) || [];
+
+        if (options?.deny && options?.deny.length > 0) {
+          entitiesTag = "deny";
+          entitiesVal = options.deny
+            .filter((e) => e !== undefined)
+            .map(protocolEntitiesToAnalyze);
+        }
```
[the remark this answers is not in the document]
blaine-arcjet: I noticed the undefined checks near the top but the optional chaining caused me concern and I found that you were using `option` instead of `opt` which meant you were always using the first options for any given rule. Perhaps you need to add tests for configuring multiple rules with one constructor.

That being said. I really dislike the multiple-rules-in-one-constructor API and I think we should discard it for the beta.

body/index.ts
```diff
@@ -0,0 +1,112 @@
...
+      done(new Error("Stream was aborted"));
+    }
+
+    function onData(chunk: Buffer) {
+      received += chunk.length;
+
+      if (received > limit) {
+        done(new Error("request entity too large"));
+      } else {
+        buffer += decoder.decode(chunk, { stream: true });
```
blaine-arcjet, line 77: The hover docs state that:
> Returns the result of running encoding's decoder. The method can be invoked zero or more times with options's stream set to true, and then once without options's stream (or set to false), to process a fragmented input. If the invocation without options's stream (or set to false) has no input, it's clearest to omit both arguments.
>
```
var string = "", decoder = new TextDecoder(encoding), buffer;
while(buffer = next_chunk()) {
  string += decoder.decode(buffer, {stream:true});
}
string += decoder.decode(); // end-of-queue
```
> If the error mode is "fatal" and encoding's decoder returns error, throws a TypeError.

It seems that you need to call `decoder.decode()` once upon completion otherwise you might not get everything that is buffered or an error.

examples/nextjs-14-app-dir-si/.env.local.example
blaine-arcjet: This was missed

examples/nodejs-express-si/index.js
blaine-arcjet: Not addressed how I was thinking but I opened another comment.

protocol/test/client.test.ts
blaine-arcjet: I didn't see this addressed. Opened another comment.

examples/nextjs-14-app-dir-rl/middleware.ts
blaine-arcjet: Taken care of

blaine-arcjet:
>  It's also failing to run npm ci... despite no package-lock changes happening when I run npm install locally.

The problem seemed to be that GitHub Actions creates a merge commit when run on PRs, so it found that the `@rollup/wasm-node` package was different in `@arcjet/body` than the rest of the repo which was a different lockfile. I'm sorry that I told you that I didn't think the dependency changes would be a problem when they were. Anyway, I merged main and updated that dependency and things look to be building.

blaine-arcjet (changes requested):
I commented inline as I was changing things and addressed some stuff. I wanted to get things I could change easily done so @e-moran can touch the bigger stuff and we hopefully can get this wrapped up tomorrow. I re-flagged the big things that weren't addressed or weren't answered.

arcjet/index.ts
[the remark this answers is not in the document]
davidmytton: > That being said. I really dislike the multiple-rules-in-one-constructor API and I think we should discard it for the beta.

@blaine-arcjet Could you add a note about that to #928 please? I can't picture what you mean exactly, so it'd be good to have an example (current, and proposed change) 😊

examples/express-sensitive-info/index.js
e-moran: It should just be in `req.body` because of the `app.use(express.json())` below. I'll change it to use `express.text()` instead.

arcjet/index.ts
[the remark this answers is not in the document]
blaine-arcjet: Added a new issue #1397

Also referenced the issue in #928

arcjet-next/index.ts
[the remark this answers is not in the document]
e-moran: Yeah it is more Node.js-like, but they have a few helpers such as `.body`

https://nextjs.org/docs/pages/building-your-application/routing/api-routes#request-helpers

arcjet-sveltekit/index.ts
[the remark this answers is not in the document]
e-moran: It should be, its listed as a field in their docs without any caveats

https://kit.svelte.dev/docs/types#public-types-requestevent

protocol/test/client.test.ts
e-moran: We don't seem to have tests for sending specific rules, the rule-specific tests are in convert.test.ts. I've added tests for the new rule in there alongside the other ones.

body/test/body.test.ts
```diff
@@ -0,0 +1,342 @@
...
+        } catch (err) {
+          expect(err).toEqual(new Error("request entity too large"));
+        } finally {
+          server.close(done);
+        }
+      });
+    });
+  });
+
+  test("should error if it isnt the exact length specified", (done) => {
```
blaine-arcjet, line 80:
```suggestion
  test("should error if it isn't the exact length specified", (done) => {
```

body/test/body.test.ts
```diff
@@ -0,0 +1,342 @@
...
+
+  test("should error if the `on` function isn't present on the object", (done) => {
+    const server = http.createServer(async (req, res) => {
+      try {
+        const reqNoOn = {
+          removeListener: req.removeListener,
+          readable: req.readable,
+        };
+        const body = await readBody(reqNoOn, { limit: 1024 });
+        console.dir(body);
```
blaine-arcjet, line 125:
```suggestion
```

body/test/body.test.ts
```diff
@@ -0,0 +1,342 @@
...
+
+  test("should error if the `removeListener` function isn't present on the object", (done) => {
+    const server = http.createServer(async (req, res) => {
+      try {
+        const reqNoOn = {
+          on: req.on,
+          readable: req.readable,
+        };
+        const body = await readBody(reqNoOn, { limit: 1024 });
+        console.dir(body);
```
blaine-arcjet, line 163:
```suggestion
```

body/test/body.test.ts
```diff
@@ -0,0 +1,342 @@
...
+  test("should error if the stream is not readable", (done) => {
+    const server = http.createServer(async (req, res) => {
+      try {
+        const reqNoOn = {
+          on: req.on,
+          removeListener: req.removeListener,
+          readable: false,
+        };
+        const body = await readBody(reqNoOn, { limit: 1024 });
+        console.dir(body);
```
blaine-arcjet, line 202:
```suggestion
```

body/test/body.test.ts
```diff
@@ -0,0 +1,342 @@
...
+  test("should error if limit is not present", (done) => {
+    const server = http.createServer(async (req, res) => {
+      try {
+        const reqNoOn = {
+          on: req.on,
+          removeListener: req.removeListener,
+          readable: req.readable,
+        };
+        const body = await readBody(reqNoOn, { limit: undefined as any });
+        console.dir(body);
```
blaine-arcjet, line 241:
```suggestion
```

body/test/body.test.ts
```diff
@@ -0,0 +1,342 @@
...
+
+    try {
+      await readBody(stream, { limit: 100 });
+      throw new Error("this should not return successfully");
+    } catch (e) {
+      expect(e).toEqual(new Error("received no body chunks after 100ms"));
+    }
+
+    // setting test timeout to 105 because the promise should throw after 100ms
+  }, 105);
```
blaine-arcjet, line 286:
```suggestion
  }, 150);
```
Timers in JS can drift. I'd set to at least 150 but maybe even 200 or 250

body/test/body.test.ts
blaine-arcjet, line 1: Some of the tests you added made me want to double check the original tests, so I'm going to do that locally and report back.

examples/express-sensitive-info/README.md
```diff
@@ -0,0 +1,45 @@
...
+1. From the root of the project, install the SDK dependencies.
+
+   ```bash
+   npm ci
+   ```
+
+2. Enter this directory and install the example's dependencies.
+
+   ```bash
+   cd examples/nodejs-express-rl
```
blaine-arcjet, line 24:
```suggestion
   cd examples/express-sensitive-info
```

examples/express-sensitive-info/index.js
```diff
@@ -0,0 +1,41 @@
...
+    res.writeHead(200, { "Content-Type": "application/json" });
+    res.end(JSON.stringify({ message: `You said: ${req.body}` }));
+  }
+})
+
+app.use(express.text());
+
+app.listen(port, () => {
+  console.log(`Example app listening on port ${port}`)
+})
```
blaine-arcjet, line 41:
```suggestion
  console.log(`Example app listening on port ${port}`);
});
```
Nits. We don't format on these files to keep them close to boilerplate.

examples/nextjs-14-pages-wrap/pages/api/arcjet.ts
blaine-arcjet, line 1: I'll reset this file.

examples/nextjs-14-sensitive-info/README.md
```diff
@@ -0,0 +1,38 @@
...
+1. From the root of the project, install the SDK dependencies.
+
+   ```bash
+   npm ci
+   ```
+
+2. Enter this directory and install the example's dependencies.
+
+   ```bash
+   cd examples/nextjs-14-app-dir-rl
```
blaine-arcjet, line 24:
```suggestion
   cd examples/nextjs-14-sensitive-info
```

examples/nextjs-14-sensitive-info/README.md
```diff
@@ -0,0 +1,38 @@
...
+3. Rename `.env.local.example` to `.env.local` and add your Arcjet key.
+
+4. Start the dev server.
+
+   ```bash
+   npm run dev
+   ```
+
+5. Curl `http://localhost:3000/api/arcjet` with some data
+   `curl http://localhost:3000/api/arcjet --data "my email is test@example.com"`
```
blaine-arcjet, line 37: This example will be allowed because you allow email. Should we add a sample that will be blocked?

examples/nextjs-14-sensitive-info/app/api/arcjet/route.ts
```diff
@@ -0,0 +1,36 @@
+import arcjet, { sensitiveInfo } from "@arcjet/next";
+import { NextResponse } from "next/server";
+
+const aj = arcjet({
+  // Get your site key from https://app.arcjet.com
+  // and set it as an environment variable rather than hard coding.
+  // See: https://nextjs.org/docs/app/building-your-application/configuring/environment-variables
+  key: process.env.ARCJET_KEY!,
+  rules: [
```
blaine-arcjet, line 9: Please add shield. We try to add shield in every SDK example because we originally ran it implicitly but now it must be configured.

examples/nextjs-14-sensitive-info/app/api/arcjet/route.ts
```diff
@@ -0,0 +1,36 @@
...
+    // Allows email addresses to be submitted and blocks all other types of sensitive information
+    sensitiveInfo({
+      allow: ["email"], // Will block all sensitive information types other than email.
+      // deny: ["credit-card-number", (tokens) => { return new Array(tokens.length).fill("custom") }], // Will block all sensitive information types other than email.
```
blaine-arcjet, line 13: This commented example is outdated with the new custom detect API.

examples/nextjs-14-sensitive-info/middleware.ts
blaine-arcjet, line 1: Let's remove this middleware from the example

examples/express-sensitive-info/package-lock.json
blaine-arcjet, line 1: Need to regenerate the package-lock in this express example because it has the old name.

.github/workflows/reusable-examples.yml
```diff
@@ -546,3 +546,45 @@ jobs:
       - name: Build
         working-directory: examples/sveltekit
         run: npm run build
+
+  nextjs-14-sensitive-info:
```
blaine-arcjet, line 550: Instead of adding to the end of the file, can you insert it between the 2 examples as found in the directory tree? Whenever I need to audit these, I compare with my directory listing.

blaine-arcjet (changes requested):
Some nits and various cleanup but we're nearly done!

protocol/test/client.test.ts
blaine-arcjet: Interesting. Can you open an issue to add those tests? 🙏

arcjet-sveltekit/index.ts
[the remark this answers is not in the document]
blaine-arcjet: 🤞

examples/nextjs-14-app-dir-si/package.json
blaine-arcjet: Please open an issue or PR that cleans this up

arcjet-next/index.ts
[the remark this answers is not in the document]
blaine-arcjet: I wonder if we need a pages route in your next example so we can test this stuff 🤔

examples/nextjs-14-sensitive-info/app/api/arcjet/route.ts
e-moran: I've replaced this with a much better example

arcjet/index.ts
```diff
@@ -310,11 +318,49 @@ export type EmailOptions = {
   allowDomainLiteral?: boolean;
 };
 
+type DetectEntities<T> = (
```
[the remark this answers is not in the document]
e-moran: I've renamed it for now. Is there any benefit to restricting the type further? I know that in Rust its a best practise to leave types as open as possible and then restrict further in the actual call sites (like how you can technically construct a HashMap<K, V> where V isn't Hash so its completely useless), but not sure if this is the same in TS.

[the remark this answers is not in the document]
blaine-arcjet: I think it's fine here because this type is not exported. If we exported it, we wouldn't want someone to be able to construct the type with a non-string because then it could never be applied to a function that works with the `detect` config option.

examples/express-sensitive-info/package-lock.json
blaine-arcjet: Missed this but I got it with `npm i --package-lock-only`

arcjet-next/index.ts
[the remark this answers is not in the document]
blaine-arcjet: I'm also just realizing that this probably needs `readBody` as well because Next users can disable body parsing. I think we should merge this PR and open an issue to build a couple pages examples and implement `readBody`.

examples/nextjs-14-sensitive-info/app/api/arcjet/route.ts
```diff
@@ -0,0 +1,50 @@
...
+  // See: https://nextjs.org/docs/app/building-your-application/configuring/environment-variables
+  key: process.env.ARCJET_KEY!,
+  rules: [
+    shield({
+      mode: "LIVE", // will block requests. Use "DRY_RUN" to log only
+    }),
+    // allows all pii entities other than email addresses and those containing a dash character.
+    sensitiveInfo({
+      // allow: ["EMAIL"], Will block all sensitive information types other than email.
+      deny: ["EMAIL", "CONTAINS_DASH"], // Will block all sensitive information types other than email.
```
blaine-arcjet, line 25:
```suggestion
      deny: ["EMAIL", "CONTAINS_DASH"], // Will block email and any custom detected values that "contain dash"
```

arcjet-next/index.ts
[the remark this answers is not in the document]
e-moran: https://github.com/arcjet/arcjet-js/issues/1399

blaine-arcjet (approved):
Excellent work here! Excited to get this released next week 💪

protocol/test/client.test.ts
blaine-arcjet: #1398

examples/nextjs-14-app-dir-si/package.json
blaine-arcjet: #1400

PR #1358, opened 2024-08-20
feat: add library to perform redaction of sensitive information

e-moran:
Adds the library to perform redaction on text using the Arcjet sensitive info primitive.

Redaction is performed by the `redact()` function, which takes a `string` and any configuration that you want to use. It returns the redacted text and an `unredact()` function as a `[redacted, unredact]` tuple. Calling `unredact()` with a string will replace the fake data with the original data.


arcjet-redact/.eslintignore
blaine-arcjet, line 1: This directory should just be called `redact`, not `arcjet-redact`. The `arcjet-*` names are for adapters, e.g. arcjet-next is the arcjet adapter for next.

redact-wasm/workerd.ts
```diff
@@ -0,0 +1,91 @@
...
+  };
+
+  if (typeof analyze !== "undefined") {
+    return analyze.detectSensitiveInfo(candidate, options);
+  } else {
+    throw new Error(
+      "detectSensitiveInfo failed to run because wasm is not supported in this enviornment",
+    );
+  }
+}
```
blaine-arcjet, line 91: I don't think you want this implementation inside redact-wasm because you need to implement it multiple times across the various runtimes.

Instead, we probably just want the `redact-wasm` package to provide the `init` function (perhaps we rename it `initialize()`?) and then you consume that in the `redact package.

arcjet-redact/index.ts
```diff
@@ -0,0 +1,221 @@
...
+function performReplacementInText(
+  text: string,
+  replacement: string,
+  start: number,
+  end: number,
+): string {
+  return text.substring(0, start) + replacement + text.substring(end);
+}
+
+export class RedactSession<
```
blaine-arcjet, line 117: Why did you decide on using a class for the API instead of a function? Do you expect users to use `identify()` or is it leaking an implementation detail?

Could `unredact` be a configuration option instead of a function?

e-moran: The class stores the mapping of redacted entities back to their originals. So the intent is that a user would create an instance of the class for each redaction/unredaction pair they want to perform, it doesn't cause any issues if a user wants to just use redact though and discard the class.

I think its cleaner to have a separate function for unredact though because the naming is a little clearer. Otherwise you'd have something like `transform()` that could either be redact or unredact.

redact-wasm/workerd.ts
e-moran: Yeah good idea, is there any reason we don't do that in the sdk?

blaine-arcjet: Because I didn't come up with the pattern until I made the `transport` package and haven't had time to refactor.

arcjet-redact/index.ts
e-moran: Going to change this so that wasm exports a redact function rather than detect sensitive info. This will return an object containing the redacted string and the mapping of original -> replacement. It will look something like this:
```typescript
{
  redacted: string,
  replacedEntities: ReplacedEntity[]
}

type ReplacedEntity = {
  original: string,
  replaced: string
} 
```

`unredact()` will also be provided which will take a string and a `ReplacedEntity[]` to unredact the text.

blaine-arcjet: I think `ReplaceEntity` is an extension of `detected-sensitive-info-entity`

This will probably be the WIT:
```wit
record redacted-sensitive-info-entity {
    // The start index of the detected entity relative to the original text
    start: u32,
    // The end index of the detected entity relative to the original text
    end: u32,
    // The identified sensitive information type
    identified-type: sensitive-info-entity,
    original: string,
    replaced: string,
}
```

blaine-arcjet: > `unredact()` will also be provided which will take a string and a ReplacedEntity[] to unredact the text.

Incorrect, we'll remove the class and instead provide a function like this:
```
const [redacted, unredact] = redact(string, options)
```
The `redacted-sensitive-info-entity` will be available inside the `redact` closure, so the returned `unredact` function will have access to them.

blaine-arcjet:
Random thought: do we want to remove the logger from this library? I'm not sure it adds much and we can streamline the Wasm a bit without it.

e-moran:
After discussing this with @blaine-arcjet further, I am going to rework the design of this so that `redact` returns a tuple containing the unredact function as well as the redacted string. The redacted strings that will replace the original entities in the plaintext will be generate in wasm, but the text replacement will occur in js to avoid an unneeded copy of the string. The replacers object will also be replaced with just a single function that takes in the identified entity type and can optionally return a replacement string for it. If none is returned then a default will be used.

e-moran:
I've pushed the changes there. Just two things that we need to resolve:
1. In your example above you called the custom redact function `redact()`, but that's the name of the list of entities to be redacted. I have called it `replace` but I'm open to suggestions.
2. I don't think its possible for entities to inform the input type of `replace()` because WASM returns custom types as a `string`. I'm not sure how much of an issue this actually is, but its not ideal.
Everything else should be good to go, test coverage isn't 100% but the only uncovered pieces of code are those that throw in a non-wasm environment. I'm not sure how we can actually test these though.

redact-wasm/README.md
```diff
@@ -0,0 +1,78 @@
...
+## Example
+
+```ts
+import { generateFingerprint, isValidEmail } from "@arcjet/analyze";
+
+const fingerprint = generateFingerprint("127.0.0.1");
+console.log("fingerprint: ", fingerprint);
+
+const valid = isValidEmail("hello@example.com");
+console.log("is email valid?", valid);
```
blaine-arcjet, line 40: Wrong example

redact-wasm/README.md
```diff
@@ -0,0 +1,78 @@
...
+```
+
+## Implementation
+
+This package provides pii identification logic implemented as a WebAssembly module which
+will run local analysis on the provided string.
+
+The [arcjet.wasm.js](./wasm/arcjet_analyze_bindings_redact.component.js) file contains the binary inlined as
```
blaine-arcjet, line 48:
```suggestion
The [Wasm](./wasm/arcjet_analyze_bindings_redact.component.js) file contains the binary inlined as
```

redact-wasm/index.ts
```diff
@@ -0,0 +1,69 @@
...
+  RedactSensitiveInfoConfig,
+  SensitiveInfoEntity,
+} from "./wasm/arcjet_analyze_bindings_redact.component.js";
+
+import { wasm as componentCoreWasm } from "./wasm/arcjet_analyze_bindings_redact.component.core.wasm?js";
+import { wasm as componentCore2Wasm } from "./wasm/arcjet_analyze_bindings_redact.component.core2.wasm?js";
+import { wasm as componentCore3Wasm } from "./wasm/arcjet_analyze_bindings_redact.component.core3.wasm?js";
+import { ArcjetRedactCustomRedact } from "./wasm/interfaces/arcjet-redact-custom-redact.js";
+
+// TODO: Do we actually need this wasmCache or does `import` cache correctly?
```
blaine-arcjet, line 14:
```suggestion
```
We don't need this TODO. Since the implementations are split for the runtime, this obviously needs to cache because the fetch-initialization is expensive.

redact-wasm/package.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "@arcjet/redact-wasm",
+  "version": "1.0.0-alpha.21",
```
blaine-arcjet, line 3:
```suggestion
  "version": "1.0.0-alpha.22",
```
We're on a new version now

redact-wasm/package.json
```diff
@@ -0,0 +1,69 @@
...
+  "description": "Arcjet local analysis engine",
```
blaine-arcjet, line 4: Align this description with the descripting in readme.

redact-wasm/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "prepublishOnly": "npm run build",
+    "jco": "jco transpile wasm/arcjet_analyze_bindings_redact.component.wasm --no-wasi-shim --instantiation async -o wasm",
+    "build": "npm run jco; rollup --config rollup.config.js",
+    "lint": "eslint .",
+    "pretest": "npm run build",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest --passWithNoTests"
+  },
+  "sideEffects": [
+    "./wasm/arcjet_analyze_bindings_redact_bg.js"
+  ],
```
blaine-arcjet, line 52:
```suggestion
```
This is old code from the wasm-bindgen days

redact-wasm/package.json
```diff
@@ -0,0 +1,69 @@
...
+  "dependencies": {},
+  "devDependencies": {
+    "@arcjet/eslint-config": "1.0.0-alpha.21",
+    "@arcjet/rollup-config": "1.0.0-alpha.21",
+    "@arcjet/tsconfig": "1.0.0-alpha.21",
```
blaine-arcjet, line 57:
```suggestion
    "@arcjet/eslint-config": "1.0.0-alpha.22",
    "@arcjet/rollup-config": "1.0.0-alpha.22",
    "@arcjet/tsconfig": "1.0.0-alpha.22",
```
out of sync

redact-wasm/package.json
```diff
@@ -0,0 +1,69 @@
...
+    "@bytecodealliance/jco": "1.4.3",
+    "@jest/globals": "29.7.0",
+    "@rollup/wasm-node": "4.20.0",
+    "@types/node": "18.18.0",
+    "jest": "29.7.0",
+    "typescript": "5.5.4"
```
blaine-arcjet, line 63: Some of these are also out of sync, I think

redact-wasm/rollup.config.js
blaine-arcjet, line 1: This file didn't get changed between analyze and this package? I think we can move it into the rollup-config now. I'll do that in a separate PR.

redact-wasm/wasm/interfaces/arcjet-redact-logger.d.ts
```diff
@@ -0,0 +1,3 @@
+export namespace ArcjetRedactLogger {
+  export function debug(msg: string): void;
+}
```
blaine-arcjet, line 3: I think this can be removed since you removed the logger, right?

redact-wasm/workerd.ts
```diff
@@ -0,0 +1,56 @@
...
+    return componentCore2Wasm;
+  }
+  if (path === "arcjet_analyze_bindings_redact.component.core3.wasm") {
+    return componentCore3Wasm;
+  }
+
+  throw new Error(`Unknown path: ${path}`);
+}
+
+function noOpFn(): any {}
```
blaine-arcjet, line 27:
```suggestion
```

redact/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+# `arcjet`
+
+<p>
+  <a href="https://www.npmjs.com/package/arcjet">
+    <picture>
+      <source media="(prefers-color-scheme: dark)" srcset="https://img.shields.io/npm/v/arcjet?style=flat-square&label=%E2%9C%A6Aj&labelColor=000000&color=5C5866">
+      <img alt="npm badge" src="https://img.shields.io/npm/v/arcjet?style=flat-square&label=%E2%9C%A6Aj&labelColor=ECE6F0&color=ECE6F0">
+    </picture>
+  </a>
```
blaine-arcjet, line 16: Wrong links

redact/README.md
```diff
@@ -0,0 +1,60 @@
+<a href="https://arcjet.com" target="_arcjet-home">
+  <picture>
+    <source media="(prefers-color-scheme: dark)" srcset="https://arcjet.com/logo/arcjet-dark-lockup-voyage-horizontal.svg">
+    <img src="https://arcjet.com/logo/arcjet-light-lockup-voyage-horizontal.svg" alt="Arcjet Logo" height="128" width="auto">
+  </picture>
+</a>
+
+# `arcjet`
```
blaine-arcjet, line 8: Wrong name

redact/README.md
```diff
@@ -0,0 +1,60 @@
...
+    </picture>
+  </a>
+</p>
+
+[Arcjet][arcjet] helps developers protect their apps in just a few lines of
+code. Implement rate limiting, bot protection, email verification, and defense
+against common attacks.
+
+This is the [Arcjet][arcjet] TypeScript and JavaScript sensitive information
+detection and redaction library.
```
blaine-arcjet, line 24:
```suggestion
This is the [Arcjet][arcjet] TypeScript and JavaScript sensitive information redaction library.
```

redact/README.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Getting started
+
+Visit [docs.arcjet.com](https://docs.arcjet.com) to get started.
```
blaine-arcjet, line 28:
```suggestion
```

redact/README.md
```diff
@@ -0,0 +1,60 @@
...
+const session = new RedactSession({ redact: ["email", "phone-number"] });
+const text = "Hi, my name is John and my email adress is john@example.com";
+const redacted = await session.redact(text);
+console.log(redacted);
+// Hi, my name is John and my email address is <REDACTED INFO #0>
+
+const unredacted = await session.unredact(
+  "Your email address is <REDACTED INFO #0>",
+);
+console.log(unredacted); // Your email address is john@example.com
```
blaine-arcjet, line 48: Wrong example

redact/index.ts
```diff
@@ -0,0 +1,191 @@
+import { initializeWasm, type SensitiveInfoEntity } from "@arcjet/redact-wasm";
```
blaine-arcjet, line 1:
```suggestion
import { initializeWasm } from "@arcjet/redact-wasm";
import type { SensitiveInfoEntity } from "@arcjet/redact-wasm";
```
While you *can* import types like that and work with verbatimModuleSyntax, I want us to keep them as separate imports.

redact/index.ts
```diff
@@ -0,0 +1,191 @@
...
+  redact: SensitiveInfoEntities<Detect, CustomEntities>;
+  contextWindowSize?: number;
+  detect?: Detect;
+  replace?: Replace;
+};
+
+function userEntitiesToWasm<Custom extends string>(
+  entity: ArcjetSensitiveInfoType | Custom,
+) {
+  // TODO: Make these the same as analyze.js
```
blaine-arcjet, line 36: This needs to check if the type is a string otherwise throw.

redact/index.ts
```diff
@@ -0,0 +1,191 @@
...
+function performReplacementInText(
+  text: string,
+  replacement: string,
+  start: number,
+  end: number,
+): string {
+  return text.substring(0, start) + replacement + text.substring(end);
+}
+
+function noOpFn(..._args: any[]): any {}
```
blaine-arcjet, line 88: We should make individual functions and type them correctly.

redact/index.ts
```diff
@@ -0,0 +1,191 @@
...
+
+async function callRedactWasm<
+  Detect extends DetectSensitiveInfoEntities<CustomEntities>,
+  CustomEntities extends string,
+>(
+  candidate: string,
+  options: RedactOptions<Detect, CustomEntities>,
+): Promise<RedactedSensitiveInfoEntity[]> {
+  let convertedDetect = noOpFn;
+  if (typeof options.detect !== "undefined") {
```
blaine-arcjet, line 106:
```suggestion
  if (typeof options.detect === "function") {
```
What if it was passed as a string?

redact/index.ts
```diff
@@ -0,0 +1,191 @@
...
+    const detect = options.detect;
+    convertedDetect = (tokens: string[]) => {
+      return detect(tokens)
+        .filter((e) => typeof e !== "undefined")
+        .map((e) => userEntitiesToWasm(e));
+    };
+  }
+
+  let convertedRedact = noOpFn;
+  if (typeof options.replace !== "undefined") {
```
blaine-arcjet, line 116:
```suggestion
  if (typeof options.replace === "function") {
```
What if it was passed as a string?

redact/index.ts
```diff
@@ -0,0 +1,191 @@
...
+    const replace = options.replace;
+    convertedRedact = (identifiedType: SensitiveInfoEntity) => {
+      return replace(wasmEntitiesToString(identifiedType));
+    };
+  }
+
+  const wasm = await initializeWasm(convertedDetect, convertedRedact);
+
+  const skipCustomDetect = options.detect === undefined;
+  const skipCustomRedact = options.redact === undefined;
```
blaine-arcjet, line 126: What if these were passed as a string?

redact/index.ts
```diff
@@ -0,0 +1,191 @@
...
+
+  const config = {
+    entities: options.redact.map(userEntitiesToWasm),
+    contextWindowSize: options.contextWindowSize,
+    skipCustomDetect,
+    skipCustomRedact,
+  };
```
blaine-arcjet, line 133: Build the config inside the "if wasm" statement.

redact/index.ts
```diff
@@ -0,0 +1,191 @@
...
+  if (typeof wasm !== "undefined") {
+    return wasm.redact(candidate, config).map((e) => {
+      return {
+        ...e,
+        identifiedType: wasmEntitiesToString(e.identifiedType),
+      };
+    });
+  } else {
+    throw new Error(
+      "redact failed to run because wasm is not supported in this enviornment",
```
blaine-arcjet, line 144:
```suggestion
      "redact failed to run because Wasm is not supported in this environment",
```

redact/index.ts
```diff
@@ -0,0 +1,191 @@
...
+
+interface RedactedSensitiveInfoEntity {
+  original: string;
+  redacted: string;
+  start: number;
+  end: number;
+  identifiedType: string;
+}
+
+async function callRedactWasm<
```
blaine-arcjet, line 98: Why did you decide to make this a separate function and not just have the logic in the `redact` function?

redact/index.ts
```diff
@@ -0,0 +1,191 @@
...
+    throw new Error(
+      "redact failed to run because wasm is not supported in this enviornment",
+    );
+  }
+}
+
+type Unredact = (input: string) => string;
+
+export async function redact<
+  Detect extends DetectSensitiveInfoEntities<CustomEntities>,
```
blaine-arcjet, line 152: Pretty sure this needs another type variable for the redact function.

redact/index.ts
```diff
@@ -0,0 +1,191 @@
...
+    candidate = performReplacementInText(
+      candidate,
+      redaction.redacted,
+      redaction.start + extraOffset,
+      redaction.end + extraOffset,
+    );
+
+    // We need to track an extra offset because the indices are relative to the original text.
+    // When we make our first replacement we shift all subsequent replacements, and this needs to be accounted for.
+    extraOffset += redaction.redacted.length - redaction.original.length;
```
blaine-arcjet, line 171: Can we work in reverse to avoid this?

redact/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@arcjet/redact",
+  "version": "1.0.0-alpha.21",
```
blaine-arcjet, line 3:
```suggestion
  "version": "1.0.0-alpha.22",
```
keep in sync

redact/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "description": "Arcjet PII Redaction Library",
```
blaine-arcjet, line 4: Align this description and the readme description

redact/package.json
```diff
@@ -0,0 +1,60 @@
...
+  "license": "Apache-2.0",
+  "homepage": "https://arcjet.com",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/arcjet/arcjet-js.git",
+    "directory": "arcjet-redact"
```
blaine-arcjet, line 10:
```suggestion
    "directory": "redact"
```
You renamed the directory

redact/package.json
```diff
@@ -0,0 +1,60 @@
...
+  ],
+  "scripts": {
+    "prepublishOnly": "npm run build",
+    "build": "rollup --config rollup.config.js",
+    "lint": "eslint .",
+    "pretest": "npm run build",
+    "test": "NODE_OPTIONS=--experimental-vm-modules jest"
+  },
+  "dependencies": {
+    "@arcjet/redact-wasm": "1.0.0-alpha.21"
```
blaine-arcjet, line 43:
```suggestion
    "@arcjet/redact-wasm": "1.0.0-alpha.22"
```
keep in sync

redact/package.json
```diff
@@ -0,0 +1,60 @@
...
+  },
+  "devDependencies": {
+    "@arcjet/eslint-config": "1.0.0-alpha.21",
+    "@arcjet/rollup-config": "1.0.0-alpha.21",
+    "@arcjet/tsconfig": "1.0.0-alpha.21",
```
blaine-arcjet, line 48:
```suggestion
    "@arcjet/eslint-config": "1.0.0-alpha.22",
    "@arcjet/rollup-config": "1.0.0-alpha.22",
    "@arcjet/tsconfig": "1.0.0-alpha.22",
```
Keep in sync

redact/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "@edge-runtime/jest-environment": "3.0.1",
```
blaine-arcjet, line 49:
```suggestion
```
Unused?

redact/package.json
```diff
@@ -0,0 +1,60 @@
...
+    "@jest/globals": "29.7.0",
+    "@rollup/wasm-node": "4.20.0",
+    "@types/node": "18.18.0",
+    "jest": "29.7.0",
+    "typescript": "5.5.4"
```
blaine-arcjet, line 54: Some of these need to be synced, I think

redact/index.ts
e-moran: Yeah that's a great idea!

package-lock.json
```diff
@@ -6663,6 +6697,299 @@
...
+        "@rollup/wasm-node": "4.20.0",
+        "@types/node": "18.18.0",
+        "jest": "29.7.0",
+        "typescript": "5.5.4"
+      },
+      "engines": {
+        "node": ">=18"
+      }
+    },
+    "redact-wasm/node_modules/@arcjet/eslint-config": {
```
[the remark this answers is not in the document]
e-moran: Yeah npm rm'd them after I updated to alpha.22

redact-wasm/README.md
e-moran: I just removed the example instead now that it just provides initialisation. Do you think it makes any sense to keep one around?

redact-wasm/package.json
e-moran: Worth removing it from `analyze` as well?

redact-wasm/rollup.config.js
e-moran: Yeah its the same

redact/index.ts
e-moran: Just for readability/separation of concerns really, there's no technical reason for it. Everything in `callRedactWasm` is converting to/from wasm types and checking inputs, everything in `redact` is performing manipulations on the string.

redact/index.ts
e-moran: I added one, though it then requires an `as` when we call replace because the Wasm custom entity type is typed as `string`, so the type can't be informed the whole way though.

redact-wasm/README.md
blaine-arcjet: I think having an example still makes sense, especially showing the "if defined" check, etc.

redact-wasm/README.md
blaine-arcjet: I actually realized that this section is totally out of date because we use rollup to output in the _virtual directory. Let me think through rewording it.

redact-wasm/package.json
blaine-arcjet: Yeah. Please open a PR.

redact/index.ts
blaine-arcjet: Constant string returns would be a subtype of `string` so it should still work. Would you like me to work on the types?

redact-wasm/package.json
e-moran: https://github.com/arcjet/arcjet-js/pull/1444/files

redact-wasm/README.md
```diff
@@ -0,0 +1,62 @@
...
+<p>
+  <a href="https://www.npmjs.com/package/@arcjet/redact-wasm">
+    <picture>
+      <source media="(prefers-color-scheme: dark)" srcset="https://img.shields.io/npm/v/%40arcjet%2Fredact-wasm?style=flat-square&label=%E2%9C%A6Aj&labelColor=000000&color=5C5866">
+      <img alt="npm badge" src="https://img.shields.io/npm/v/%40arcjet%2Fredact-wasm?style=flat-square&label=%E2%9C%A6Aj&labelColor=ECE6F0&color=ECE6F0">
+    </picture>
+  </a>
+</p>
+
+This is the [Arcjet][arcjet] Redaction library PII detection engine.
```
blaine-arcjet, line 19:
```suggestion
[Arcjet][arcjet] sensitive information redaction detection engine.
```

redact-wasm/README.md
```diff
@@ -0,0 +1,62 @@
...
+
+## Installation
+
+```shell
+npm install -S @arcjet/redact-wasm
+```
+
+## Implementation
+
+This package provides pii identification logic implemented as a WebAssembly module which
```
blaine-arcjet, line 61:
```suggestion
This package provides sensitive information identification and redaction logic implemented as a WebAssembly module which
```

redact-wasm/README.md
```diff
@@ -0,0 +1,62 @@
...
+will run local analysis on the provided string.
+
+The [Wasm](./wasm/arcjet_analyze_bindings_redact.component.js) file contains the binary inlined as
+a base64 [Data URL][mdn-data-url] with the `application/wasm` MIME type.
```
blaine-arcjet, line 65:
```suggestion
The generated `_virtual/arcjet_analyze_bindings_redact.component.js` file contains the binary inlined as
a base64 [Data URL][mdn-data-url] with the `application/wasm` MIME type.
```

redact-wasm/README.md
```diff
@@ -0,0 +1,62 @@
...
+ways based on the platform or bundler a user is targeting. One example being
+that Next.js requires special `asyncWebAssembly` webpack config to load our
+Wasm file if we don't do this.
+
+In the future, we hope to do away with this workaround when all bundlers
+properly support consistent asset bundling techniques.
+
+## API
+
+In progress.
```
blaine-arcjet, line 85:
```suggestion
```
Remove it or write one?

redact-wasm/edge-light.ts
```diff
@@ -0,0 +1,55 @@
...
+  ImportObject,
+  RedactedSensitiveInfoEntity,
+  RedactSensitiveInfoConfig,
+  SensitiveInfoEntity,
+} from "./wasm/arcjet_analyze_bindings_redact.component.js";
+
+import componentCoreWasm from "./wasm/arcjet_analyze_bindings_redact.component.core.wasm?module";
+import componentCore2Wasm from "./wasm/arcjet_analyze_bindings_redact.component.core2.wasm?module";
+import componentCore3Wasm from "./wasm/arcjet_analyze_bindings_redact.component.core3.wasm?module";
+import type { ArcjetRedactCustomRedact } from "./wasm/interfaces/arcjet-redact-custom-redact.js";
```
blaine-arcjet, line 12: Nit: move this near the other type/core imports and not near the wasm module imports

redact-wasm/package.json
```diff
@@ -0,0 +1,66 @@
+{
+  "name": "@arcjet/redact-wasm",
+  "version": "1.0.0-alpha.22",
+  "description": "Arcjet PII identification and redaction engine.",
```
blaine-arcjet, line 4:
```suggestion
  "description": "Arcjet sensitive information redaction detection engine",
```
Aligm but I remove the periods on these

redact-wasm/index.ts
```diff
@@ -0,0 +1,68 @@
...
+  ImportObject,
+  RedactedSensitiveInfoEntity,
+  RedactSensitiveInfoConfig,
+  SensitiveInfoEntity,
+} from "./wasm/arcjet_analyze_bindings_redact.component.js";
+
+import { wasm as componentCoreWasm } from "./wasm/arcjet_analyze_bindings_redact.component.core.wasm?js";
+import { wasm as componentCore2Wasm } from "./wasm/arcjet_analyze_bindings_redact.component.core2.wasm?js";
+import { wasm as componentCore3Wasm } from "./wasm/arcjet_analyze_bindings_redact.component.core3.wasm?js";
+import type { ArcjetRedactCustomRedact } from "./wasm/interfaces/arcjet-redact-custom-redact.js";
```
blaine-arcjet, line 12: Nit: move this near the other type/core imports and not near the wasm module imports

redact-wasm/workerd.ts
```diff
@@ -0,0 +1,54 @@
+import * as core from "./wasm/arcjet_analyze_bindings_redact.component.js";
+import type {
+  ImportObject,
+  RedactedSensitiveInfoEntity,
+  RedactSensitiveInfoConfig,
+  SensitiveInfoEntity,
+} from "./wasm/arcjet_analyze_bindings_redact.component.js";
+import componentCoreWasm from "./wasm/arcjet_analyze_bindings_redact.component.core.wasm";
```
blaine-arcjet, line 8:
```suggestion

import componentCoreWasm from "./wasm/arcjet_analyze_bindings_redact.component.core.wasm";
```
Nit: I'd like to keep a space between the "normal" imports and the wasm module imports

redact-wasm/workerd.ts
```diff
@@ -0,0 +1,54 @@
...
+import componentCore2Wasm from "./wasm/arcjet_analyze_bindings_redact.component.core2.wasm";
+import componentCore3Wasm from "./wasm/arcjet_analyze_bindings_redact.component.core3.wasm";
+import type { ArcjetRedactCustomRedact } from "./wasm/interfaces/arcjet-redact-custom-redact.js";
```
blaine-arcjet, line 11: Nit: move this near the other type/core imports and not near the wasm module imports

redact-wasm/README.md
e-moran: Added it back!

redact/README.md
```diff
@@ -0,0 +1,54 @@
...
+console.log(redacted);
+// Hi, my name is John and my email address is <Redacted email #0>
+
+const unredacted = unredact("Your email address is <Redacted email #0>");
+console.log(unredacted); // Your email address is john@example.com
+```
+
+## API
+
+TODO
```
blaine-arcjet, line 47:
```suggestion
```
I think we should either write it or remove it.

redact-wasm/README.md
```diff
@@ -0,0 +1,94 @@
...
+// If WebAssembly isn't available in the environment then it will be undefined.
+if (typeof wasm !== "undefined") {
+  const config = {
+    entities: [],
+    contextWindowSize: 1,
+    skipCustomDetect: true,
+    skipCustomRedact: true,
+  };
+
+  wasm.redact("I am a string", config);
```
blaine-arcjet, line 51:
```suggestion
  const entities = wasm.redact("I am a string", config);
  // Do something with entities that should be redacted.
```

redact/index.ts
```diff
@@ -0,0 +1,208 @@
...
+      candidate,
+      redaction.redacted,
+      redaction.start,
+      redaction.end,
+    );
+  }
+
+  function unredact(input: string): string {
+    for (const redaction of redactions) {
+      const position = input.indexOf(redaction.redacted);
```
blaine-arcjet, line 193: Should you be looping while you can find the redaction text? I would imagine a response could have the same redaction string multiple times.

redact/index.ts
```diff
@@ -0,0 +1,208 @@
...
+>(
+  candidate: string,
+  options: RedactOptions<Detect, Replace, CustomEntities>,
+): Promise<[string, Unredact]> {
+  const redactions = await callRedactWasm(candidate, options);
+
+  // Need to apply the redactions in reverse order so that the offsets aren't changed
+  // when we redact with strings that are longer/shorter than the original.
+  redactions.reverse();
+  for (const redaction of redactions) {
```
blaine-arcjet, line 182:
```suggestion

  for (const redaction of redactions) {
```
Nit: having a space here seems like I'd look at the comment and reverse line above.

redact/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@arcjet/redact",
+  "version": "1.0.0-alpha.22",
+  "description": "Arcjet sensitive information redaction library.",
```
blaine-arcjet, line 4:
```suggestion
  "description": "Arcjet sensitive information redaction library",
```
Nit: I remove the periods here.

blaine-arcjet (changes requested):
Some more nits, but 1 big question that I wanted to ask about unredact. Still need to review the tests and work on the type signatures of the custom redact function.

redact/index.ts
e-moran: Yeah good catch! I'll fix this and add a test to ensure that this works correctly

redact/README.md
e-moran: My plan for this was to copy what we do in the arcjet package and link to the docs. Since we don't have the docs written yet we can't do that. I'll remove the section for now and open an issue to add it back once we have published the docs.

e-moran: https://github.com/arcjet/arcjet-js/issues/1445

redact/index.ts
e-moran: The issue I was having is [here](https://github.com/arcjet/arcjet-js/pull/1358/files#diff-2e25abfb099bd46d41b989f26d6fb2fa881c7a9fde1c97d2af801df3ff1d0f90R134) where we need to pass a `string` in to a function with a more precise type. We know from our own understanding of the code that it should always be valid, but I couldn't find a way of expressing it in TypeScript because we just get a `string` in the generated bindings. If you know a way of making that work properly that would be great.

blaine-arcjet: In exploring the type inference on this, I found that the sensitiveInfo inference doesn't work correctly either. We need to fix them.

redact-wasm/rollup.config.js
blaine-arcjet: We'll keep it for now and I'll move it later.

blaine-arcjet: #1447

redact/index.ts
blaine-arcjet: The sample that I've been playing with is:

```ts
const [redacted, unredact] = await redact("abc, 123", {
  // entities: ["credit-card-number", "abc"],
  // detect: (tokens: string[]) => {
  //   return tokens.map((token) => {
  //     if (token == "abc") {
  //       return "abc";
  //     }
  //   });
  // },
  replace: (entity) => {
    // if (entity === "abc") return "123456";
    if (entity === "email") return "foobarbaz";
    // if (entity === "") return "123";
  },
});
```

Essentially, we should be able to comment/uncomment the various pieces and the type inference should be good. I have inlay hints enabled for all my TS so I can see how it is inferring things.

I'm going to push up my changes, but be aware that they have the same caveat as https://github.com/arcjet/arcjet-js/pull/1446#issuecomment-2319084446 and I'm not good enough at TypeScript to know how to make both ways work.

blaine-arcjet:
Copying to the root of the PR so it doesn't get buried in comments:

The sample that I've been playing with is:

```ts
const [redacted, unredact] = await redact("abc, 123", {
  // entities: ["credit-card-number", "abc"],
  // detect: (tokens: string[]) => {
  //   return tokens.map((token) => {
  //     if (token == "abc") {
  //       return "abc";
  //     }
  //   });
  // },
  replace: (entity) => {
    // if (entity === "abc") return "123456";
    if (entity === "email") return "foobarbaz";
    // if (entity === "") return "123";
  },
});
```

Essentially, we should be able to comment/uncomment the various pieces and the type inference should be good. I have inlay hints enabled for all my TS so I can see how it is inferring things.

I'm going to push up my changes, but be aware that they have the same caveat as https://github.com/arcjet/arcjet-js/pull/1446#issuecomment-2319084446 and I'm not good enough at TypeScript to know how to make both ways work.

redact/index.ts
```diff
@@ -0,0 +1,212 @@
...
+> = Array<
+  ArcjetSensitiveInfoType | Exclude<ReturnType<Detect>[number], undefined>
+>;
+
+type RedactOptions<
+  Detect extends DetectSensitiveInfoEntities<CustomEntities>,
+  Replace extends ReplaceSensitiveInfoEntities<CustomEntities>,
+  CustomEntities extends string,
+> = {
+  allow?: never;
```
blaine-arcjet, line 30: Unusued.

redact/index.ts
```diff
@@ -0,0 +1,212 @@
...
+  redact: SensitiveInfoEntities<Detect, CustomEntities>;
```
blaine-arcjet, line 31: I think we should rename this to `entities`

blaine-arcjet:
I've also just discovered that the arguments of `detect` must be typed as `string[]` otherwise the inference is lost. I need think more about how to solve that.

redact/test/index.test.ts
```diff
@@ -0,0 +1,159 @@
+import { describe, expect, test } from "@jest/globals";
+import { redact } from "../index";
+
+describe("ArcjetRedact", () => {
+  describe("redact()", () => {
+    test("it will do nothing if no entities are configured", async () => {
+      const text = "email test@example.com phone 011234567 ip 10.12.234.2";
+      const [redacted] = await redact(text, { redact: [] });
```
blaine-arcjet, line 8: Don't know feel like this logic should be swapped? I feel like you should be able to use this library like `const [redacted] = await redact(text);` and have it apply all our default redactions.

blaine-arcjet (changes requested):
Big question on whether we should be defaulting to "no redactions" in the redact library.

redact-wasm/workerd.ts
blaine-arcjet: #1448

redact/test/index.test.ts
davidmytton: Do nothing, but log a warning to say that no redactions were configured?

e-moran: What if we default to redact all if `entities` (renamed from `redact`) is undefined? If they pass in an empty list then we should warn and redact nothing.

blaine-arcjet: What is the purpose of using a redact library and redacting nothing? You should just remove the call to the library in that case. I think we'd throw on an empty list.

davidmytton: Throw or log a warning, either is fine by me. There is no purpose of using a redact library and redacting nothing so the goal is to guard against user error by letting them know they made a mistake.

redact/index.ts
```diff
@@ -0,0 +1,210 @@
...
+    const skipCustomRedact = typeof options?.replace !== "function";
+
+    if (options?.entities?.length === 0) {
+      console.warn("no entities configured for redaction");
+    }
+
+    const config = {
+      entities: Array.isArray(options?.entities)
+        ? options.entities.map(userEntitiesToWasm)
+        : undefined,
```
blaine-arcjet, line 147: Now that you are checking the length, we can also do the `Array.isArray` check above (before you check the length). You need to check it's an array before you check the length because objects, strings, etc can all have length properties.

redact/test/index.test.ts
e-moran: Makes sense, I'll update it to throw if someone passes an empty list. It currently redacts all if the list isn't provided at all.

redact/index.ts
e-moran: I've tidied this up a little bit and made it throw

redact/index.ts
```diff
@@ -0,0 +1,213 @@
...
+    } else if (options?.entities?.length === 0) {
+      throw new Error("no entities configured for redaction");
+    }
+
+    const config = {
+      entities: options?.entities?.map(userEntitiesToWasm),
+      contextWindowSize: options?.contextWindowSize,
+      skipCustomDetect,
+      skipCustomRedact,
+    };
```
blaine-arcjet, line 154: All of this is super sloppy. Let's move it into a function. I wrote this:

```ts
function getWasmOptions<
  const Detect extends DetectSensitiveInfoEntities<CustomEntities> | undefined,
  const CustomEntities extends string,
>(options?: RedactOptions<Detect>) {
  if (typeof options === "object" && options !== null) {
    if (Array.isArray(options.entities)) {
      if (options.entities.length < 1) {
        throw new Error("no entities configured for redaction");
      }
    } else {
      throw new Error("entities must be an array");
    }

    return {
      entities: options.entities.map(userEntitiesToWasm),
      contextWindowSize: options.contextWindowSize || 1,
      skipCustomDetect: typeof options.detect !== "function",
      skipCustomRedact: typeof options.replace !== "function",
    };
  } else {
    return {
      entities: undefined,
      contextWindowSize: 1,
      skipCustomDetect: true,
      skipCustomRedact: true,
    };
  }
}
```

e-moran: Yeah good suggestion, this is a lot cleaner!

blaine-arcjet: Good catch on the undefined

blaine-arcjet (approved):
I think this is good to go! Great work @e-moran 🎉 I think this API is really great.

@e-moran I pushed up some changes to get to 100% code coverage. Please review before you merge.

redact-wasm/index.ts
```diff
@@ -50,7 +50,9 @@ export async function initializeWasm(
   };
 
   try {
-    return core.instantiate(moduleFromPath, coreImports);
+    // Await the instantiation to catch the failure
+    const api = await core.instantiate(moduleFromPath, coreImports);
+    return api;
```
blaine-arcjet, line 55: We need to fix this in analyze too.

redact/jest.config.js
```diff
@@ -9,6 +9,7 @@ const config = {
   // for this context". Tracking in
   // https://github.com/vercel/edge-runtime/issues/250
   coverageProvider: "v8",
+  collectCoverageFrom: ["index.js"],
```
blaine-arcjet, line 12: Avoids it running coverage on redact-wasm

redact-wasm/index.ts
blaine-arcjet: #1454

PR #1377, opened 2024-08-21
deps: Add turborepo as a root devDependency

blaine-arcjet:
This adds turborepo as a devDependency at the root of our monorepo. This will allow us to build the packages in order, which fixes an issue @e-moran encountered with the redact packages.

I chose turborepo because we already depend on some vercel packages so adding another doesn't expand our supply chain dependence.

Also, I've aliased all the npm scripts in our root to the turbo commands to make sure nothing changes from a usability standpoint.

Unfortunately, turborepo doesn't have a Wasm build so we need to allow optional dependencies again.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -134,3 +134,6 @@ dist
 
 # macOS
 .DS_Store
+
+# turborepo
+.turbo/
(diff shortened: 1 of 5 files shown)


e-moran (approved):
Solves the problem very cleanly, great work!

Issue #72, opened 2023-12-15
Consolidate examples


blaine-arcjet:
Create one example for Next 13 and one for Next 14 that don't have default files.

> Yeh I've been thinking about this. I've copied the idea from how Next.js does it at https://github.com/vercel/next.js/tree/canary/examples - they have more lightweight examples without the default files e.g. images, CSS, etc, so I'd like to do that. We should probably have just a single example app for each major version and also merge in app / pages dirs as well. Then we can more easily manage the ignore list.

_Originally posted by @davidmytton in https://github.com/arcjet/arcjet-js/issues/62#issuecomment-1857827037_

wooorm-arcjet:
This is an older issue, with not too much information.
The title is broad, but the text is only about Next.js examples.
Now halfway through 2025, what is the idea? Can this issue be split, refined? How can it be solved?

davidmytton:
We have individual examples for each use case we want to test. Each one is a distinct installation of Next.js with most of the default assets. They could be made more lightweight with just the required code to test the use case. Sometimes they might need to be independent, such as to test specific versions of Next.js. Otherwise we could combine them into a single example - this would make it easier to maintain.

PR #1437, opened 2024-08-28
feat!: Rework bot detection rule with allow/deny configuration

### LINKED ISSUE #39: TS SDK: Additional validation on detectBot options
Currently, the `detectBot` rule passes `patterns.add` and `patterns.remove` directly to the rule constructor. We likely want more validation.

blaine-arcjet:
This reworks the bot detection rule to be configured like the sensitive info rule, which is to say you can either `allow` or `deny` list of bots. I've also reworked the bot detection to look for almost 600 well-known bots.

Closes #39 - we've changed the configuration format and have validation on `allow` and `deny`.


arcjet/index.ts
```diff
@@ -835,64 +791,57 @@ export function validateEmail(
...
-    const block = Array.isArray(opt.block)
-      ? opt.block
-      : [ArcjetBotType.AUTOMATED];
-    // TODO: Does this avoid prototype pollution by putting in a Map first?
-    const addMap = new Map();
-    for (const [key, value] of Object.entries(opt.patterns?.add ?? {})) {
-      addMap.set(key, value);
+    if (typeof opt.allow !== "undefined" && typeof opt.deny !== "undefined") {
+      throw new Error(
+        "Both allow and deny cannot be provided to sensitiveInfo",
```
davidmytton, line 805:
```suggestion
        "Both allow and deny cannot be provided to detectBot",
```

davidmytton (changes requested):

blaine-arcjet:
Should be ready to test. I've added an example explicitly for bots.

davidmytton (approved):
👍 🤖 🙅‍♂️

arcjet/index.ts
```diff
@@ -836,131 +792,114 @@ export function validateEmail(
...
+        if (typeof details.headers === "undefined") {
+          throw new Error("bot detection requires `headers` to be set");
+        }
+        if (typeof details.headers.has !== "function") {
+          throw new Error(
+            "bot detection requires `headers` to extend `Headers`",
+          );
+        }
+        if (!details.headers.has("user-agent")) {
+          throw new Error("bot detection requires user-agent header");
```
e-moran, line 867: How would this be presented to the user?
I feel like they'd want to handle this differently from the two cases above since would both most likely be triggered by misuse of the SDK, but its completely valid that they'd receive a request with no user-agent header and might want to then decide whether to allow it or not.

examples/express-bots/index.js
```diff
@@ -0,0 +1,42 @@
...
+app.get('/', async (req, res) => {
+  const decision = await aj.protect(req);
+
+  if (decision.isDenied() && decision.reason.isBot()) {
+    res.writeHead(400, { "Content-Type": "application/json" });
+    res.end(JSON.stringify({
+      error: "You are a bot",
+      detected: decision.reason.denied[0]
+    }));
+  } else {
```
e-moran, line 34: Since we're adding shield as a rule as well should we do something with a potential shield response in the example?

e-moran (commented):
Looks great! This is going to be a huge improvement from the current bot detection, just pointing out two small things that I noticed while looking through it.

arcjet/index.ts
blaine-arcjet: This gets turned into an ArcjetErrorReason. If we don't have a user agent, we can't detect bots so it's an error reason and we fail open.

examples/express-bots/index.js
blaine-arcjet: I really don't want to. Adding branching for shield takes away from what the example is showing.

We add shield to every example because we originally "auto added" shield and so we keep it in every example to mirror that behavior.

arcjet/index.ts
e-moran: Yeah but how would a user use that? I would imagine that an error like the above two would be turned in to a 500, where not sending a user-agent header would suit a 400 better. If it became a 5xx that would be wrong since the error was caused by the client not sending required headers, not a server error.

examples/express-bots/index.js
e-moran: Would it be worth even adding a comment indicating that you should add handling for shield? My concern is that a user would just copy the example, assume that its set up to handle both rules since they're both in there, then be confused as to why shield does nothing

blaine-arcjet: These examples are not boilerplates to copy, they are just for testing specific functionality works and continues to work in the SDK during development.

Please add feedback to #72 which discusses how we need to create consolidated examples that show everything.

davidmytton: We started these as examples to copy, but they became more part of our test suite. We still refer to them in some places as examples, but we're going to deprioritize that and focus on the docs code samples and other "full" example apps. Should probably rename them from examples as well, but this is for #72.

arcjet/index.ts
e-moran: Yeah I think that structured errors would solve it quite nicely, but they are obviously out of scope for this. It might just be worth documenting it when it comes to updating the docs.

e-moran (approved):

arcjet/index.ts
blaine-arcjet: https://github.com/arcjet/arcjet-docs/issues/45

PR #1446, opened 2024-08-29
fix(arcjet): Infer types when no detect function is specified

blaine-arcjet:
While trying to improve the type inference on #1358, I found that we weren't providing the correct type inference on the `sensitiveInfo` rule when a user **does not** specify a `detect()` function. This is likely the most common scenario so we want to have proper type inference here.

To solve this, I made `Detect` very generic throughout the types and then used some techniques (commented inline) to figure out if the function was specified in the options.

### DIFF
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -336,38 +336,39 @@ type DetectSensitiveInfoEntities<T> = (
   tokens: string[],
 ) => Array<ArcjetSensitiveInfoType | T | undefined>;
 
-type SensitiveInfoOptionsAllow<
-  Detect extends DetectSensitiveInfoEntities<CustomEntities>,
-  CustomEntities extends string,
-> = {
-  allow: Array<
-    ArcjetSensitiveInfoType | Exclude<ReturnType<Detect>[number], undefined>
-  >;
+type ValidEntities<Detect> = Array<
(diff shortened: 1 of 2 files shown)


blaine-arcjet:
There's a really annoying behavior with this type inference. When the `detect` property is specified, it needs to use a colon and can't be the object function shorthand syntax. Such that:

```ts
const [rule] = sensitiveInfo({
        mode: "LIVE",
        allow: ["custom", "CREDIT_CARD_NUMBER"],
// WORKS
        detect: (tokens: string[]) => {
            if (token === "test@example.com") {
              return "custom";
            }
          });
        },
        contextWindowSize: 1,
});

const [rule] = sensitiveInfo({
        mode: "LIVE",
        allow: ["custom", "CREDIT_CARD_NUMBER"],
// FAILS
        detect(tokens: string[]) {
            if (token === "test@example.com") {
              return "custom";
            }
          });
        },
        contextWindowSize: 1,
});
```

I can't figure out how to make TypeScript support both so that's just a caveat here.

blaine-arcjet:
I think this has the other caveat I found in #1358:

> I've also just discovered that the arguments of detect must be typed as string[] otherwise the inference is lost. I need think more about how to solve that.

e-moran (approved):

PR #1776, opened 2024-09-26
feat: Add NestJS adapter

### LINKED ISSUE #781: NestJS adapter
We have a NestJS example, but we can create an adapter the provides Arcjet as a guard, etc.

blaine-arcjet:
This adds a NestJS adapter to our SDK. It is definitely just a *first* implementation while we wait for feedback from users.

NestJS is a very different framework than adapters for previous frameworks, but I tried to make it feel natural for users of NestJS. This includes the `ArcjetModule` export which they can import into their `@Module` decorators, the `ArcjetGuard` export which can be used with `@UseGuards(ArcjetGuard)`, and the `@WithArcjetRules` decorator which can provide additional rules to the ArcjetGuard that is run before the handler.

Unlike previous adapters, the design of NestJS doesn't allow us to leverage our type inference to help guide users, so we rely more heavily on runtime errors if things are misconfigured, such as trying to use a `validateEmail` rule with `ArcjetGuard`.

One benefit of the adapter is that we support NestJS on top of both Express and Fastify!

I've revamped the NestJS example to highlight various ways the adapter can be used and even created a NestJS + Fastify example to ensure we work with both.

Closes #781


blaine-arcjet:
@SocketSecurity ignore npm/fast-content-type-parse@1.1.0
    @SocketSecurity ignore npm/light-my-request@5.13.0
    @SocketSecurity ignore npm/fast-json-stringify@5.16.1
    @SocketSecurity ignore npm/avvio@8.4.0

All fastify packages.

blaine-arcjet:
We should consider using https://docs.nestjs.com/fundamentals/execution-context#current-application-context to make sure we're in an http context but let's wait for user feedback.

blaine-arcjet:
I've created an example that uses LaunchDarkly to configure the rules at NestJS application startup. This demonstrates that the ArcjetModule can be configured via async logic by using `ArcjetModule.forRootAsync` with a class/factory/etc.

examples/nestjs-launchdarkly/src/app.module.ts
```diff
@@ -0,0 +1,35 @@
+import { Module } from '@nestjs/common';
+import { ConfigModule } from '@nestjs/config';
+import { ArcjetGuard, ArcjetModule, shield } from '@arcjet/nest';
```
davidmytton, line 3:
```suggestion
import { ArcjetGuard, ArcjetModule } from '@arcjet/nest';
```

davidmytton (approved):

davidmytton (commented):
Great to see these examples!

PR #1782, opened 2024-09-27
feat: Add Deno adapter

### LINKED ISSUE #758: Support for Deno
Allow us to support applications built with Deno.

blaine-arcjet:
This adds a Deno adapter to our SDK. The implementation is very close to the Bun implementation with the platform-specific APIs swapped for Deno.

The work we did to make our core platform-agnostic has really paid off.

I created an example that uses `sensitiveInfo` to verify that our body reading logic works correctly. You may notice a bunch of relative `imports` in the `deno.json` file, which is a workaround for loading all arcjet-js packages from the monorepo. I believe these won't be required once we load this via the `npm:@arcjet/deno` specifier, but we won't know for certain until it is published.

Closes #758


e-moran (approved):
Looks great!

PR #1785, opened 2024-09-29
chore!: Add options validation for all rules

### LINKED ISSUE #992: Missing `refillRate` on TokenBucket configuration causes error
Might be the lack of the required `refillRate` or the inclusion of an unknown `fillRate`, but the following config:

```js
    tokenBucket({
      mode: "LIVE",
      capacity: 40,
      fillRate: 10,
      interval: 60,
    }),
```

Caused the following in DataDog:

```
Key: 'TokenBucketOptions.RefillRate' Error:Field validation for 'RefillRate' failed on the 'required' tag
```

Should probably return an error to the app, not the server.

blaine-arcjet:
This adds option validation on all of our rules. We've relied too heavily on TypeScript types to ensure that all options are passed correctly, but that has lead to issues when TypeScript is configured less strictly or JavaScript is used.

I've written some very simple validation machinery in our core package and implemented for all our rules.

Closes #992

### DIFF
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -800,13 +800,17 @@ export function sensitiveInfo<
     typeof options.allow !== "undefined" &&
     typeof options.deny !== "undefined"
   ) {
-    throw new Error("`sensitiveInfo` options error: `allow` and `deny` cannot be provided together");
+    throw new Error(
+      "`sensitiveInfo` options error: `allow` and `deny` cannot be provided together",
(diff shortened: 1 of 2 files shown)


blaine-arcjet:
I want to add tests for validation on all options to each rule.

davidmytton (approved):

e-moran (approved):
Very nice!

PR #1838, opened 2024-10-02
feat: Use `waitUntil` for Report call if available

### LINKED ISSUE #884: waitUntil on ArcjetAdapterContext
We should attach the `waitUntil` function (if it is available) to the ArcjetAdapterContext.

See https://vercel.com/docs/functions/functions-api-reference#waituntil
And https://developers.cloudflare.com/pages/functions/api-reference/#eventcontext

blaine-arcjet:
This attaches a `waitUntil` function to the SDK context and uses it on the call to Report if it is available. I've added the lookup of the `waitUntil` function on Vercel in Arcjet core, since Vercel supports many frameworks.

Closes #884

### DIFF
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -216,6 +216,35 @@ function toString(value: unknown) {
   return "<unsupported value>";
 }
 
+// This is the Symbol that Vercel defines in their infrastructure to access the
+// Context (where available). The Context can contain the `waitUntil` function.
+// https://github.com/vercel/vercel/blob/930d7fb892dc26f240f2b950d963931c45e1e661/packages/functions/src/get-context.ts#L6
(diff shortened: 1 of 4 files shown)


davidmytton:
Vercel [supports other frameworks](https://vercel.com/docs/frameworks/more-frameworks) including Remix and SvelteKit, so I'd go ahead and add it to the other adapters. It's doing a lookup to see if it exists anyway, so won't cause any issues if it's not there.

blaine-arcjet:
@davidmytton I've moved this into the core arcjet package so as to not duplicate the lookup logic. It still can be overridden in an adapter if that adapter has a different way to "wait until".

davidmytton (approved):

PR #1854, opened 2024-10-03
fix!: Update Wasm with phone-number fix and tokenizer update

e-moran:
Regenerates the wasm bindings to add some more supported phone number formats to the sensitive info identifier.


blaine-arcjet (approved):

blaine-arcjet:
I'm going to label this as a breaking fix since it changes the tokens that users could be receiving in custom detect functions.

PR #1988, opened 2024-10-21
chore(protocol)!: Remove received_at and decision fields from Report

blaine-arcjet:
This updates our protocol package to the latest protobuf definition which removes the `decision` from `ReportResponse` to send less data down the wire which also reduces the amount of data we need to deserialize.

This means that the decision `id` field in the report log will be undefined in older SDKs so I've marked this as breaking.

It also removed `received_at` because our usage was inconsistent. We may revisit this in the future.

### DIFF
--- a/protocol/client.ts
+++ b/protocol/client.ts
@@ -1,6 +1,5 @@
 import type { Transport } from "@connectrpc/connect";
 import { createPromiseClient } from "@connectrpc/connect";
-import { Timestamp } from "@bufbuild/protobuf";
 import {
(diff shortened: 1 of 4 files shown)


davidmytton (approved):
The req ID is still available for local decisions, so long as you update to this version of the SDK, right?

blaine-arcjet:
> The req ID is still available for local decisions, so long as you update to this version of the SDK, right?

Yes, I changed it to the `decision.id` in the closure just like other values. I don't remember why I didn't use that for the id originally.

PR #1991, opened 2024-10-22
feat(nextjs): Support Next.js Server Actions

### LINKED ISSUE #1200: Support for Next.js server actions
[Server actions](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations) are the recommended way to do things like [handling forms in Next.js](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations#forms). They don't act like route handlers and there is no `request` object to pass to the Arcjet `protect()` function.

We can get the headers using the [headers](https://nextjs.org/docs/app/api-reference/functions/headers) function, but the main request details currently needs to be constructed manually. I hacked together a workaround for a form handler, but we should support this in a nicer way

```ts
import arcjet, { validateEmail } from "@arcjet/next";
import { headers } from "next/headers";
//import { ipAddress } from '@vercel/functions';

// Utility function to get the IP address of the client
// From https://nextjs.org/docs/app/api-reference/functions/headers#ip-address
function IP() {
    const FALLBACK_IP_ADDRESS = "127.0.0.1";

    // Uncomment if running on Vercel
    //const ip = ipAddress(request)
    //return ip ?? FALLBACK_IP_ADDRESS;

    const forwardedFor = headers().get("x-forwarded-for");

    if (forwardedFor) {
        return forwardedFor.split(",")[0] ?? FALLBACK_IP_ADDRESS;
    }

    return headers().get("x-real-ip") ?? FALLBACK_IP_ADDRESS;
}

// Arcjet client defined outside of the component for example purposes. You
// probably want to put this somewhere else in your app.
const aj = arcjet({
    key: process.env.ARCJET_KEY!, // Get your site key from https://app.arcjet.com
    rules: [
        validateEmail({
            mode: "LIVE", // will block requests. Use "DRY_RUN" to log only
            block: ["DISPOSABLE", "NO_MX_RECORDS"], // block disposable email addresses
        }),
    ],
});

// Test form component
export default function Form() {
    async function testForm(formData: FormData) {
        "use server";

        const headersList = headers();
        const host = headersList.get("host");

        // Construct the request needed by Arcjet because we're not in a request
        // handler here.
        const path = new URL(`http://${host}`);
        const req = {
            ip: IP(),
            method: "POST",
            host: headersList.get("host"),
            url: path.href,
            headers: headersList,
        };

        const email = formData.get("email") as string;

        const decision = await aj.protect(req, {
            email,
        });

        console.log("Decision: ", decision);
    }

    return (
        <form action={testForm}>
            <input type="email" name="email" />
            <button type="submit">Submit</button>
        </form>
    );
}
```

blaine-arcjet:
This adds support for Next.js Server Actions. It primarily adds an async `request()` function that builds a minimal request object with just Headers and Cookies, but was enabled by reducing the details we need in our Service.

I've also built an example on [Next.js 15](https://nextjs.org/blog/next-15) to show that `characteristics` must be set on the SDK if the IP is not available in the Headers. I chose to set a UUID cookie when the page is initially loaded and use it to fingerprint, which makes fingerprinting consistent between development and production.

Closes #1200


blaine-arcjet:
@SocketSecurity ignore npm/detect-libc@2.0.3
    @SocketSecurity ignore npm/sharp@0.33.5
    @SocketSecurity ignore npm/next@15.0.0

Issue #1904, opened 2024-10-13
Review changes to Next.js 15


davidmytton:
Review [the changes in Next.js](https://nextjs.org/docs/canary/app/building-your-application/upgrading/version-15) for anything we may need to adjust.

From a quick look, we should review:

- [Previously synchronous Dynamic APIs that rely on runtime information are now asynchronous](https://nextjs.org/docs/canary/app/building-your-application/upgrading/version-15#async-request-apis-breaking-change)
- [The geo and ip properties on NextRequest have been removed as these values are provided by your hosting provider](https://nextjs.org/docs/canary/app/building-your-application/upgrading/version-15#nextrequest-geolocation)

davidmytton:
[RC2 is now out](https://nextjs.org/blog/next-15-rc2). Additional things to note:

- [Further details on the dynamic APIs](https://nextjs.org/blog/next-15-rc2#async-request-apis-breaking-change)
- [New Form component](https://nextjs.org/docs/canary/app/api-reference/components/form). Not a breaking change, but something we want to support.

PR #2018, opened 2024-10-22
feat(ip)!: Accept Request or IncomingMessage directly

blaine-arcjet:
This changes the `findIP` API so it can accept a `Request` object directly. This is achieved by requiring the `headers` field on the first argument. I've combined the separate "request-like" and "headers" argument into one.

This is primarily being done so `@arcjet/ip` can be used with Next.js 15 which [removed the `ip` property from `NextRequest`](https://nextjs.org/docs/canary/app/building-your-application/upgrading/version-15#nextrequest-geolocation).

Ref #1904

### DIFF
--- a/ip/index.ts
+++ b/ip/index.ts
@@ -551,6 +551,8 @@ export interface RequestLike {
   info?: PartialInfo;
 
   requestContext?: PartialRequestContext;
+
+  headers: Headers;
 }
 
 export type Platform = "cloudflare" | "fly-io" | "vercel";
@@ -580,11 +582,7 @@ export interface Options {
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
 // SOFTWARE.
-function findIP(
-  request: RequestLike,
-  headers: Headers,
-  options: Options = {},
-): string {
+function findIP(request: RequestLike, options: Options = {}): string {
   // Prefer anything available via the platform over headers since headers can
   // be set by users. Only if we don't have an IP available in `request` do we
   // search the `headers`.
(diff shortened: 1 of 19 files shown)


ip/index.ts
```diff
@@ -618,13 +626,13 @@ function findIP(
 
   if (platform === "cloudflare") {
```
davidmytton, line 627: Do we want to add a [Vercel platform check](https://vercel.com/docs/projects/environment-variables/system-environment-variables#VERCEL) because we know they set [`x-forwarded-for`](https://vercel.com/docs/edge-network/headers/request-headers#x-forwarded-for)?

blaine-arcjet: We need to support `x-forwarded-for` and `x-real-ip` for nginx and other proxies so those are handled by generic checks; however, that presents the question: if we can detect the platform, should we return an empty string if we can't detect the global IP from the platform-specific header? I think we should, which would mean we need to do an additional check for those 2 headers for vercel.

davidmytton: > should we return an empty string if we can't detect the global IP from the platform-specific header?

I think so. It _should_ always be there, but we can fall back to our other logic if it isn't for whatever reason.

blaine-arcjet: I just opened https://github.com/arcjet/arcjet-js/pull/2021. Let's land that and then I'll add the Vercel checks in another PR to have a changelog entry.

blaine-arcjet: Followed up on that with #2022 and then I'll rebase this.

davidmytton (approved):

blaine-arcjet:
I just realize that this changes makes the signature incompatible with Node's IncomingRequest

davidmytton:
Is there a test for that?

blaine-arcjet:
> Is there a test for that?

Nope. We only inherently supported it due to the type signature, but it would be nice to fully support it with these changes. There will be a test once I finish my updates.

ip/test/ipv4.test.ts
```diff
@@ -5,122 +5,171 @@ import { describe, expect, test } from "@jest/globals";
...
+  test("returns empty string if headers is not object", () => {
+    expect(
+      ip({
+        // @ts-expect-error
+        headers: "",
+      }),
+    ).toEqual("");
+  });
+
+  // Support for Node.js IncomingRequest
```
blaine-arcjet, line 38:
```suggestion
  // Support for Node.js IncomingMessage
```

ip/test/ipv4.test.ts
```diff
@@ -5,122 +5,171 @@ import { describe, expect, test } from "@jest/globals";
...
+    const request = {
+      headers: {
+        // Node.js lowercases the header keys
+        "x-real-ip": "1.1.1.1",
+      },
+    };
+    expect(ip(request)).toEqual("1.1.1.1");
+  });
+
+  // Support for Node.js IncomingRequest
```
blaine-arcjet, line 49:
```suggestion
  // Support for Node.js IncomingMessage
```

PR #2019, opened 2024-10-22
fix!: Ensure performance metrics are scoped to same call

### LINKED ISSUE #1865: Timing functions are still broken
Even after #1858, there are still bugs with the `log.time()` and `log.timeEnd()` calls.

Since we run things async, one `protect` call might call `log.time()` and another might call `log.timeEnd()`.

I think we need some unique ID to ensure the timing is correct. Otherwise we'll need to implement timing logic ourselves, using `performance.now()` and probably returning an end function.

blaine-arcjet:
This implements performance logging with our own functions instead of using `console.time()` and `console.timeEnd()` so we can keep them correctly scoped to the `protect()` call. Prior to this, if 2 calls to `protect()` were racing, one might call `time()` first and the other may call `timeEnd()` which would mess up all the labels.

There are opportunities for performance improvements here, especially if the loglevel isn't `debug`, but we should also check if using closures is more performant than building a map with unique symbols.

Fixes #1865

### DIFF
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -200,6 +200,24 @@ function isEmailType(type: string): type is ArcjetEmailType {
   );
 }
 
+class Performance {
+  log: ArcjetLogger;
+
+  constructor(logger: ArcjetLogger) {
+    this.log = logger;
+  }
+
+  // TODO: We should no-op this if loglevel is not `debug` to do less work
+  measure(label: string) {
+    const start = performance.now();
+    return () => {
+      const end = performance.now();
+      const diff = end - start;
+      this.log.debug("LATENCY %s: %sms", label, diff.toFixed(3));
+    };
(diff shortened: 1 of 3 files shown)


davidmytton (approved):

arcjet/index.ts
```diff
@@ -200,6 +200,24 @@ function isEmailType(type: string): type is ArcjetEmailType {
...
 }
 
+class Performance {
+  log: ArcjetLogger;
+
+  constructor(logger: ArcjetLogger) {
+    this.log = logger;
+  }
+
+  // TODO: We should no-op this if loglevel is not `debug` to do less work
```
blaine-arcjet, line 210:
```suggestion
  // TODO(#2020): We should no-op this if loglevel is not `debug` to do less work
```

Issue #2153, opened 2024-11-05
Failure when deployed on Netlify Edge Functions


blaine-arcjet:
It seems that undici is bundled when trying to deploy to Netlify Edge Functions; however, they use an old version of Deno that seems to crash on `node:` specifiers (specifically, `node:assert`).

This should be using deno fetch directly but I need to figure out how it is resolved.

blaine-arcjet:
I don't think we want to fix this with a workaround because the problem will either:
1. Go away when Netlify updates the version of Deno they use
2. No longer be an issue when we write our own tranport and drop the ConnectRPC dependency.

PR #2155, opened 2024-11-05
feat(transport): Make SvelteKit adapter deployable on Netlify Edge Functions

blaine-arcjet:
Netlify Edge Funcitons run on a super old version of Deno that fails on some `node:*` imports. This is a really nasty workaround that allows Arcjet-protected SvelteKit apps to be deployed there.

### DIFF
--- a/transport/svelte.ts
+++ b/transport/svelte.ts
@@ -57,10 +57,8 @@ export function createTransport(baseUrl: string) {
       let client;
       try {
         client = await node;
-        console.log("node client");
       } catch {
         client = await web;
-        console.log("web client");
       }
 
       return client.unary(...args);
--- a/transport/svelte.ts
+++ b/transport/svelte.ts
@@ -1,0 +1,81 @@
+// This entire workaround is to make SvelteKit run on Netlify Edge Functions
+// which execute in an old version of Deno that doesn't support some `node:*`
+// imports, such as `node:assert`. This makes the undici import fail when we try
+// to use `@connectrpc/connect-node`.
+//
+// By making these dynamic imports, we can await them inside of the `unary` or
+// `stream` function calls to figure out which client we are using.
+
+export function createTransport(baseUrl: string) {
+  const web = import("@connectrpc/connect-web").then(
+    ({ createConnectTransport }) => {
+      // The Connect Node client doesn't work on edge runtimes: https://github.com/bufbuild/connect-es/pull/589
+      // so set the transport using connect-web. The interceptor is required for it work in the edge runtime.
+      return createConnectTransport({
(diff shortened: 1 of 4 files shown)


transport/svelte.ts
```diff
@@ -0,0 +1,81 @@
...
+      ...args: [any, any, any, any, any, any, any]
+    ): Promise<unknown> {
+      let client;
+      try {
+        client = await node;
+      } catch {
+        client = await web;
+      }
+
+      return client.unary(...args);
```
blaine-arcjet, line 78:
```suggestion
      return client.stream(...args);
```

transport/svelte.ts
```diff
@@ -0,0 +1,81 @@
...
+  );
+
+  return {
+    async unary(
+      ...args: [any, any, any, any, any, any, any]
+    ): Promise<unknown> {
+      let client;
+      try {
+        client = await node;
+        console.log("node client");
```
blaine-arcjet, line 60:
```suggestion
        console.log("node client");
```

transport/svelte.ts
```diff
@@ -0,0 +1,81 @@
...
+      } catch {
+        client = await web;
+        console.log("web client");
```
blaine-arcjet, line 63:
```suggestion
        console.log("web client");
```

transport/svelte.ts
```diff
@@ -0,0 +1,79 @@
...
+        baseUrl,
+        httpVersion: "2",
+        sessionManager,
+      });
+    },
+  );
+
+  return {
+    async unary(
+      ...args: [any, any, any, any, any, any, any]
```
e-moran, line 55: Can we get some concrete types for these?

transport/svelte.ts
```diff
@@ -0,0 +1,79 @@
...
+      try {
+        client = await node;
+      } catch {
+        client = await web;
+      }
+
+      return client.unary(...args);
+    },
+    async stream(
+      ...args: [any, any, any, any, any, any, any]
```
e-moran, line 67: Same as above

e-moran (changes requested):

transport/svelte.ts
blaine-arcjet: I believe these types come from connectrpc directly and we don't want to add another dependency to get the types. There may be a way to use utility types to pull them out but I'm not sure.

blaine-arcjet:
I'm going to close this because it was a proof-of-concept to see if #2153 could be solved with the current architecture. While it can, we'll have a better solution once we remove ConnectRPC (or if Netlify upgrades the version of Deno they are using).

PR #2156, opened 2024-11-05
fix(sveltekit): Load env from `node:process`

### LINKED ISSUE #2154: Netlify Edge Functions use different global than process.env
When deploying to Netlify Edge functions, the [Netlify.env](https://docs.netlify.com/functions/environment-variables/#edge-functions-2) needs to be used to access the environment. This causes a problem with adapters because they expect to access environment in a specific way.

blaine-arcjet:
This loads `env` from the `node:process` import. We want to do this because some runtimes, like Netlify Edge Functions, don't provide the global `process.env` but instead will shim the `node:process` import to provide the environment.

Closes #2154

### DIFF
--- a/arcjet-sveltekit/index.ts
+++ b/arcjet-sveltekit/index.ts
@@ -15,6 +15,7 @@ import { baseUrl, isDevelopment, logLevel, platform } from "@arcjet/env";
 import { Logger } from "@arcjet/logger";
 import { createClient } from "@arcjet/protocol/client.js";
 import { createTransport } from "@arcjet/transport";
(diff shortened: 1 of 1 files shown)


davidmytton:
Should we do this everywhere and not just in the SvelteKit adapter?

blaine-arcjet:
> Should we do this everywhere and not just in the SvelteKit adapter?

SvelteKit is the only framework that I know about that has an explicit "netlify edge" adapter, so let's keep it scoped to this until we hear differently.

davidmytton (approved):

PR #2237, opened 2024-11-14
feat: Create nosecone package for creating secure headers

blaine-arcjet:
This implements our `nosecone` package and 2 adapters, Next.js and SvelteKit. These 2 frameworks have some of the best support for nonce-based CSPv3—although Next.js has the caveat of it only working in dynamic mode.

Runtimes like Bun, Deno, and Node.js can use Nosecone directly to set headers on the responses, while adapters are needed for deeper integration. Using middleware works really well for Next.js because we can force the headers to be forwarded and it even detects the nonce from the `script-src` directive, which it adds to each `<script>` tag that webpack generates. For SvelteKit, we need to provide `csp` in the config so it'll add the CSP header, but we also use a hook to add our additional secure headers.

Notably missing:
- Removing X-Powered-By—the frameworks don't give us access to this, we should instead just recommend it be removed.
- A Remix adapter—it needs a lot of logic to thread nonces throughout the application.
- An Express adapter—most people are already using Helmet.


davidmytton (commented):
> All that being said, I really don't like the generalized concept here and I think this makes more sense as a series of packages crafted for specific frameworks. 

What do you think about having a central package that defines all the headers and their default values which is imported by other framework-specific packages? Then they can be natively integrated for each framework and the config can be overridden if necessary.

blaine-arcjet:
> What do you think about having a central package that defines all the headers and their default values which is imported by other framework-specific packages?

Yeah, I was thinking about it this morning and it'd be cool to provide functions that produce "good" values for use by any packages. I just need to decide on the API; my initial thought being that we could return an instance of `Headers` containing all of them.

examples/nextjs-app-dir-validate-email/app/layout.tsx
```diff
@@ -9,11 +10,14 @@ export const metadata: Metadata = {
   description: "Generated by create next app",
 };
 
-export default function RootLayout({
+export default async function RootLayout({
   children,
 }: {
   children: React.ReactNode;
 }) {
+  // Opt-out of static generation for every page so the CSP nonce can be applied
```
davidmytton, line 18: Isn't this quite a big change in how Sveltekit works? Could this happen at build time?

nosecone-sveltekit/index.ts
```diff
@@ -0,0 +1,64 @@
...
+    return options;
+  }
+
+  return {
+    ...options,
+    contentSecurityPolicy: false,
+  };
+}
+
+// Setting specific headers is the way that Next implements middleware
```
davidmytton, line 18: Are these valid in Sveltekit?

examples/nextjs-app-dir-validate-email/app/layout.tsx
blaine-arcjet: This is actually a Next.js example.

The problem with Next.js is that it doesn't hash files that it produces statically (SvelteKit does this!). The result is that if you don't opt-out of static generation, then you need to set  `'self' 'unsafe-inline' 'unsafe-eval'` on the CSP as per their docs at https://nextjs.org/docs/app/building-your-application/configuring/content-security-policy#without-nonces

This defeats the much of the purpose of the CSP, since untrusted JavaScript can run on the page. Here's the warnings and errors from https://csp-evaluator.withgoogle.com/:
Screenshot 2024-11-18 at 8 26 44 AM

I tried getting access to metadata about static/dynamic pages but it's not available at runtime. I believe this makes it impossible for us to figure out if the unsafe headers or nonce should be used and I opted for the secure nonce-based headers since we're trying to push safe defaults.

davidmytton: Makes sense. We can document this reasoning and explain the tradeoffs.

nosecone-sveltekit/index.ts
blaine-arcjet: Removed.

blaine-arcjet:
I've updated & cleaned up all the implementations, reached 100% code coverage on the core `nosecone` package, and created simple readmes for everything. I've also updated the body of the root PR with the changes/decisions made.

examples/nextjs-app-dir-rate-limit/middleware.ts
```diff
@@ -18,4 +18,4 @@ const aj = arcjet({
   ],
 });
 
-export default createMiddleware(aj);
\ No newline at end of file
+export default createMiddleware(aj);
```
davidmytton, line 21: How can we use this if we already have some other middleware e.g. Clerk/Auth.js & Arcjet? How can they be chained?

davidmytton (commented):
Will you include the `@arcjet/nosecone` package in this PR?

examples/nextjs-app-dir-rate-limit/middleware.ts
blaine-arcjet: Like any other middleware. Next.js middleware is nowhere near as nice as Express middleware or SvelteKit hooks. I've been thinking about a potentially cool package for dealing with Next.js middleware but that is not the responsibility of this package.

blaine-arcjet:
> Will you include the @arcjet/nosecone package in this PR?

I was going to hold off for this first release. That's mostly because I couldn't think of a good directory name for it and I expect we'll need another release quickly to deal with any issues we encounter with our integration.

davidmytton (approved):
Looks good! I'd suggest adding an example of how to configure the headers in the READMEs, but good for us to test internally.

PR #2375, opened 2024-11-29
feat: show isSpoofed() in bot examples

e-moran:
This PR updates the examples for bot detection to show the proper use of `.isSpoofed()` to detect if a bot is not who they say they are.


davidmytton (changes requested):
~~To fully document this feature, can you add examples to all of the framework guides and an explanation in the references for each one please?~~

Nm, mis-read where this PR was (thought it was the docs).

davidmytton (approved):

PR #2380, opened 2024-11-29
chore(nosecone): Add JSDoc comments to NoseconeOptions

blaine-arcjet:
This adds JSDoc comments to every configuration option in `NoseconeOptions`. I needed these descriptions, links, etc in the reference guide so it made sense to add them here and migrate over.

### DIFF
--- a/nosecone/index.ts
+++ b/nosecone/index.ts
@@ -145,7 +145,7 @@ export interface PermittedCrossDomainPoliciesConfig {
 export interface NoseconeOptions {
   /**
    * Configure the `Content-Security-Policy` header, which helps mitigate a
-   * large number of * attacks, such as cross-site scripting.
+   * large number of attacks, such as cross-site scripting.
    *
    * Enable with defaults by specifying `true`, disable by specifying `false`,
    * or provide {@link ContentSecurityPolicyConfig} to configure individual
--- a/nosecone/index.ts
+++ b/nosecone/index.ts
@@ -143,18 +143,168 @@ export interface PermittedCrossDomainPoliciesConfig {
 }
 
 export interface NoseconeOptions {
+  /**
+   * Configure the `Content-Security-Policy` header, which helps mitigate a
+   * large number of * attacks, such as cross-site scripting.
+   *
+   * Enable with defaults by specifying `true`, disable by specifying `false`,
+   * or provide {@link ContentSecurityPolicyConfig} to configure individual
+   * items.
+   *
+   * See also:
+   * - https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP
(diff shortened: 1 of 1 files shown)


nosecone/index.ts
```diff
@@ -143,18 +143,168 @@ export interface PermittedCrossDomainPoliciesConfig {
 }
 
 export interface NoseconeOptions {
+  /**
+   * Configure the `Content-Security-Policy` header, which helps mitigate a
+   * large number of * attacks, such as cross-site scripting.
```
davidmytton, line 148:
```suggestion
   * large number of attacks, such as cross-site scripting.
```

davidmytton: Not sure if this was intentional - looks like a comment artifact.

nosecone/index.ts
```diff
@@ -143,18 +143,168 @@ export interface PermittedCrossDomainPoliciesConfig {
...
+   * Enable with defaults by specifying `true`, disable by specifying `false`,
+   * or provide {@link PermittedCrossDomainPoliciesConfig} to configure
+   * individual items.
+   *
+   * See also:
+   * - https://owasp.org/www-project-secure-headers/#x-permitted-cross-domain-policies
+   */
   xPermittedCrossDomainPolicies?: PermittedCrossDomainPoliciesConfig | boolean;
+  /**
+   * Disable the `X-XSS-Protection` header, which could introduce a browser
```
davidmytton, line 296: This is deprecated and doesn't work consistently across browsers. Should we add some note about that?

davidmytton (changes requested):
The main question this leaves is: what are the defaults? I think we should document those in the docs, but probably also in the `README` because that's what shows up on NPM etc. Then link to the docs in these comments. What were you planning @blaine-arcjet ?

nosecone/index.ts
davidmytton: Yeah fine if it's disabled by default.

Would love to see if any browsers are still using this! Must be years old at this point:

image

nosecone/index.ts
```diff
@@ -143,18 +143,168 @@ export interface PermittedCrossDomainPoliciesConfig {
...
+   * side-channel if enabled.
```
blaine-arcjet, line 297:
```suggestion
   * side-channel in legacy browsers if enabled.
```

blaine-arcjet:
> The main question this leaves is: what are the defaults? I think we should document those in the docs, but probably also in the README because that's what shows up on NPM etc. Then link to the docs in these comments. What were you planning @blaine-arcjet ?

I agree, showing the defaults is going to be an important piece of documenting this. Adding to the README and our docs site seems good; and then I can link from the JSDoc to keep them from growing too large.

davidmytton (approved):

PR #2479, opened 2024-12-06
chore: Switch most test harnesses to node:test

blaine-arcjet:
This migrates most of our test harnesses to `node:test`. This will eventually allow us to remove Jest completely.

I still need to figure out how to make custom environments to test the `@arcjet/runtime` package and run tests with the [edge-runtime](https://edge-runtime.vercel.app/packages/runtime) package but this is a step in the right direction.


davidmytton (approved):
Nice. One less dependency!

PR #2565, opened 2024-12-12
chore(runtime): Replace Jest with Node test harness

### LINKED ISSUE #9: Replace jest so we don't need NODE_OPTIONS
We want to add Windows to our CI and required env vars like NODE_OPTIONS for testing make it difficult.

Need to investigate node's built-in test harness.

blaine-arcjet:
This completes the migration away from Jest as a test harness.

I added a utility function that allows us to import our module with a different global to test the branches for various platforms.

The only downside to this design is that Node's experimental coverage support doesn't apply since we are create new modules with their own globals. I think this downside is worth it to drop hundreds of dependencies from our lockfile.

Closes #9


e-moran (approved):
+1,505 -4,473 🚀

PR #2661, opened 2024-12-20
feat: support `allow` or `deny` config in validateEmail & deprecate `block` config

### LINKED ISSUE #1834: Rename validateEmail options to align with detectBot and sensitiveInfo options
The `validateEmail` currently uses `block` to indicate which email types to deny. We should change this to `deny` like `detectBot` and `sensitiveInfo` now use.

Also, should we add an `allow` variant here? cc @davidmytton

e-moran:
This PR updates the `validateEmail` rule to support the allow/deny config.

Closes #1834


arcjet/index.ts
```diff
@@ -521,13 +522,24 @@ type BotOptionsDeny = {
 
 export type BotOptions = BotOptionsAllow | BotOptionsDeny;
 
-export type EmailOptions = {
+type EmailOptionsAllow = {
```
blaine-arcjet, line 525:
```suggestion
export type EmailOptionsAllow = {
```
Need to export this to make documentation easier

arcjet/index.ts
```diff
@@ -521,13 +522,24 @@ type BotOptionsDeny = {
...
   mode?: ArcjetMode;
-  block?: ArcjetEmailType[];
+  allow: ArcjetEmailType[];
+  deny?: never;
+  requireTopLevelDomain?: boolean;
+  allowDomainLiteral?: boolean;
+};
+
+type EmailOptionsDeny = {
```
blaine-arcjet, line 533:
```suggestion
export type EmailOptionsDeny = {
```
Need to export this to make documentation easier

arcjet/index.ts
```diff
@@ -946,22 +958,65 @@ export function validateEmail(
   validateEmailOptions(options);
```
blaine-arcjet, line 958: You didn't update this validator

arcjet/index.ts
```diff
@@ -946,22 +958,65 @@ export function validateEmail(
...
-  const emailOpts = {
-    requireTopLevelDomain,
-    allowDomainLiteral,
-    blockedEmails: block,
+  let config: EmailValidationConfig = {
+    tag: "allow-email-validation-config",
+    val: {
+      requireTopLevelDomain,
+      allowDomainLiteral,
+      allow: [],
```
blaine-arcjet, line 987: Is "allow none" the right default here? I ask because we have "inverted" types, such as `NO_GRAVATAR`

blaine-arcjet (changes requested):
Accidentally submitted the other comments as single comments.

arcjet/index.ts
e-moran: I had set it to "allow none" here so it would align with bot detection/sensitive info. I don't think the inverted types change that because `allow: []` still means "block everything that we can detect" in all cases. However that default makes less sense with emails because of how restrictive it is.

Do you think we should prioritize alignment with other rules, or follow what people are more likely to want from an email validation rule (in which case we might also consider `block: ["INVALID"]` as the default).

arcjet/index.ts
e-moran: Fixed this, is there any reason this isn't the case for `BotOptionsAllow` and `BotOptionsDeny`?

blaine-arcjet: Those need to be exported (among others). I discovered this while working on the `<DisplayType />` component in our docs. Do you want to open an issue?

arcjet/index.ts
```diff
@@ -462,6 +462,8 @@ const validateEmailOptions = createValidator({
   validations: [
     { key: "mode", required: false, validate: validateMode },
     { key: "block", required: false, validate: validateEmailTypes },
```
blaine-arcjet, line 464: Block needs to be removed, right?

arcjet/index.ts
blaine-arcjet: I believe the idea (from long ago) was that primitives (e.g. `detectBot`/`validateEmail`) would be configured and not make assumptions about defaults, while products (e.g. `protectSignup`) would be preconfigured for common use cases.

We've strayed quite far from that because we never added more products and we've defaulted primitives to be the most open. For example, the current `validateEmail` blocks nothing if you don't set `block`.

I think your change to "deny none" keeps us in line with the current behavior but I'm not sure it is fundamentally ideal. Perhaps we open an issue/discussion for feedback.

arcjet/index.ts
```diff
@@ -968,16 +982,33 @@ export function validateEmail(
...
     );
   }
+  if (
+    typeof options.allow !== "undefined" &&
+    typeof options.block !== "undefined"
+  ) {
+    throw new Error(
+      "`validateEmail` options error: `allow` and `block` cannot be provided together",
+    );
+  }
```
blaine-arcjet, line 992: Shouldn't there be `allow && deny` check too?

blaine-arcjet: nvm, I was looking at a single commit.

arcjet/test/arcjet.test.ts
blaine-arcjet, line 1: Please add tests that use `// @ts-expect-error` and validates config errors when allow/deny/block specified together or not specified.

blaine-arcjet (changes requested):
This looks great. It just needs additional tests to ensure our validation works when TypeScript is ignored or not used.

blaine-arcjet (approved):
🎉 :shipit: 📧

Issue #2639, opened 2024-12-19
Combine createMiddleware with additional auth middleware


brandanking-decently:
I was wondering how I can combine the createMiddleware method with other middleware such as NextAuth

davidmytton:
Hey @brandanking-decently - our `createMiddleware` function takes another middleware as the second parameter, so you can pass it through directly. For example with Auth.js you could do this:

```ts
// This example is for Auth.js 5, the successor to NextAuth 4
import arcjet, { createMiddleware, shield } from "@arcjet/next";
import NextAuth from "next-auth";
import GitHub from "next-auth/providers/github";
// @ts-ignore
import type { NextAuthConfig, NextAuthRequest } from "next-auth";

export const config = {
  // matcher tells Next.js which routes to run the middleware on.
  // This runs the middleware on all routes except for static assets.
  matcher: ["/((?!_next/static|_next/image|favicon.ico).*)"],
};
const aj = arcjet({
  key: process.env.ARCJET_KEY!, // Get your site key from https://app.arcjet.com
  rules: [
    // Protect against common attacks with Arcjet Shield
    shield({
      mode: "LIVE", // will block requests. Use "DRY_RUN" to log only
    }),
  ],
});

export const authConfig = {
  providers: [GitHub],
} satisfies NextAuthConfig;

const { auth } = NextAuth(authConfig);

export const authMiddleware = auth(async (req: NextAuthRequest) => {
  if (!req.auth) {
    // If the user is not authenticated, return a 401 Unauthorized response. You
    // may wish to redirect to a login page instead.
    return Response.json({ message: "Unauthorized" }, { status: 401 });
  }
});

export default createMiddleware(aj, authMiddleware);
```

For multiple middleware then you may want to use a helper library like https://nemo.rescale.build

Have you tried either of these approaches? Did you have any problems?

davidmytton:
Ah, for Nosecone you would need to call it within your own custom middleware function. If you're using Auth.js with Next.js then you could create a `middleware.ts` file like this:

```ts
import { type NoseconeOptions, createMiddleware, defaults } from "@nosecone/next";
import { auth } from "auth";

// Nosecone security headers configuration
// https://docs.arcjet.com/nosecone/quick-start
const noseconeOptions: NoseconeOptions = {
  ...defaults,
};

const securityHeaders = createMiddleware(noseconeOptions);

export default auth(async (req) => {
[...]

PR #2662, opened 2024-12-22
feat(example): Add Nosecone example with custom router

davidmytton:
Created to help with https://github.com/arcjet/arcjet-js/issues/2639 so making this an official example.


examples/nosecone-nextjs-router/middleware.ts
```diff
@@ -0,0 +1,75 @@
...
+    ): Promise<NextResponse | Response> => {
+        const path = request.nextUrl.pathname || "/";
+        const addedHeaders = new Headers();
+
+        for (const [matchFunc, middlewareFuncs] of middleware) {
+            const m = matchFunc(path);
+            if (m) {
+                for (const fn of middlewareFuncs) {
+                    const resp = await fn(request, event);
+                    // TODO: better response guards
```
davidmytton, line 59: @blaine-arcjet could you review my fix in https://github.com/arcjet/arcjet-js/issues/2639#issuecomment-2558514321 please? Seems like we should handle redirects and auth errors by just returning the response, but that might leave the security headers unset.

blaine-arcjet: Is there a reason we can't add the headers to returned response before returning it from the middleware?

It looks like the implementation for `NextResponse.redirect()` isn't complex:
```js
static redirect(url, init) {
        const status = typeof init === "number" ? init : (init == null ? void 0 : init.status) ?? 307;
        if (!REDIRECTS.has(status)) {
            throw new RangeError('Failed to execute "redirect" on "response": Invalid status code');
        }
        const initObj = typeof init === "object" ? init : {};
        const headers = new Headers(initObj == null ? void 0 : initObj.headers);
        headers.set("Location", (0, _utils.validateURL)(url));
        return new NextResponse(null, {
            ...initObj,
            headers,
            status
        });
}
```

So we could return a `Response` instantly that sets the status code & Location header.

Also, to make this resilient, we probably want to handle `NextResponse.rewrite()`, which looks like it sets a different header than `NextResponse.next()`:

```js
static rewrite(destination, init) {
        const headers = new Headers(init == null ? void 0 : init.headers);
        headers.set("x-middleware-rewrite", (0, _utils.validateURL)(destination));
        handleMiddlewareField(init, headers);
        return new NextResponse(null, {
            ...init,
            headers
        });
}
```

davidmytton: Ideally we'd add the headers to all responses. So this is really about extending how we handle the redirect and rewrites to make sure they're captured correctly.

blaine-arcjet:
As per a discussion today, we are going to close this and instead implement a fully-featured middleware routing library. See https://github.com/arcjet/arcjet-js/issues/3131

PR #2684, opened 2024-12-24
chore: Deprecate Arcjet enum-like objects

### LINKED ISSUE #2621: Deprecate Enum-like objects
This issue is related to #1833.

We must first deprecate the Enum-like objects before we remove them though.

We should add a warning log to advice users to move to the string unions soon.

e-moran:
Closes #2621


blaine-arcjet (changes requested):
These should have deprecation messages.

blaine-arcjet (approved):

PR #2715, opened 2025-01-02
chore!: deprecate `reason` for `ArcjetAllowDecision`

e-moran:
This PR makes `reason` optional for `ArcjetAllowDecision` and deprecates it.

We are making this change because the rule that is accessible through the `reason` field is not obvious to our users and has the potential to cause confusion.

Instead we recommend iterating through the `ruleResults` list to find rule you are looking for.


davidmytton:
Can you add the migration example code to this PR description please? It should be documented in the changelog as well, but it'll be helpful to have it in this PR in case someone comes across it.

decorate/index.ts
```diff
@@ -208,7 +208,7 @@ export function setRateLimitHeaders(
       .sort(sortByLowestMax)
       .map(toPolicyString)
       .join(", ");
-  } else {
+  } else if (typeof decision.reason !== "undefined") {
```
blaine-arcjet, line 211: Why doesn't `isRateLimitReason` just taking an optional reason?

examples/nextjs-bot-categories/app/api/arcjet/route.ts
```diff
@@ -34,14 +34,15 @@ export async function GET(req: Request) {
   }
 
   const headers = new Headers();
-  if (decision.reason.isBot()) {
+  const botReason = decision.results.map((rule) => rule.reason).find((reason) => reason.isBot());
```
blaine-arcjet, line 37: Probably should be a `reduce` or something instead of `find`. Using `find` can result in the wrong code if multiple rules are defined.

examples/nextjs-clerk-rate-limit/app/api/arcjet/route.ts
```diff
@@ -69,13 +69,17 @@ export async function GET(req: NextRequest) {
...
-  
+
   let reset: Date | undefined;
   let remaining: number | undefined;
 
-  if (decision.reason.isRateLimit()) {
-    reset = decision.reason.resetTime;
-    remaining = decision.reason.remaining;
-  }
+  const rateLimitReason = decision.results.map((rule) => rule.reason).find((reason) => reason.isRateLimit());
```
blaine-arcjet, line 76: Don't `map` and `find`. See above, but also you are creating multiple intermediate arrays.

examples/nextjs-openai/app/api/chat/route.ts
```diff
@@ -59,8 +59,9 @@ export async function POST(req: Request) {
   const decision = await aj.protect(req, { requested: estimate });
   console.log("Arcjet decision", decision.conclusion);
 
-  if (decision.reason.isRateLimit()) {
-    console.log("Requests remaining", decision.reason.remaining);
+  const rateLimitReason = decision.results.map((rule) => rule.reason).find((reason) => reason.isRateLimit());
```
blaine-arcjet, line 62: See above.

protocol/test/convert.test.ts
```diff
@@ -519,7 +520,7 @@ describe("convert", () => {
           results: [],
           reason: new ArcjetReason(),
           ip: new ArcjetIpDetails(),
-        }),
+        }) as ArcjetDecisionWithReason,
```
blaine-arcjet, line 523: Don't cast things. TypeScript just allows any cast and accepts it.

protocol/index.ts
```diff
@@ -591,7 +591,7 @@ export abstract class ArcjetDecision {
   ip: ArcjetIpDetails;
 
   abstract conclusion: ArcjetConclusion;
-  abstract reason: ArcjetReason;
+  abstract reason: ArcjetReason | undefined;
```
blaine-arcjet, line 594: This stinks. We might want to remove these abstract fields.

protocol/index.ts
```diff
@@ -627,15 +627,20 @@ export abstract class ArcjetDecision {
   }
 }
 
+export type ArcjetDecisionWithReason = ArcjetDecision & {
+  reason: NonNullable<ArcjetDecision["reason"]>;
+};
```
blaine-arcjet, line 632: I don't like this.

blaine-arcjet (changes requested):

examples/nextjs-bot-categories/app/api/arcjet/route.ts
e-moran: I've tried a few variations of this but I haven't found one that I'm happy with yet.

Option 1:
```
let rateLimitReason = decision.results.reduce<ArcjetRateLimitReason | undefined>((result, rule) => {
    if (!result && rule.reason.isRateLimit()) {
      return rule.reason;
    }
    return result;
  }, undefined);
 ```

Option 2:
```
const rateLimitReason = decision.results.find((rule): rule is Omit<ArcjetRuleResult, 'reason'> & { reason: ArcjetRateLimitReason } => rule.reason.isRateLimit())?.reason;
```

Option 3:
```
const rateLimitReason = decision.results.find((rule): rule is ArcjetRuleResult & { reason: ArcjetRateLimitReason } => rule.reason.isRateLimit())?.reason;
```

Out of these I prefer the two using `.find()`, they seem a bit neater. I don't like that we have to specify the final type in all of them though.

The difference between 2 and 3 is that in 2 `rateLimitReason` is typed as `ArcjetRateLimitReason | undefined` where in 3 it is `(ArcjetRateLimitReason & ArcjetReason) | undefined`, which is effectively the same but a bit grosser.

We could hide a lot of this by expanding `@arcjet/decorate`, use that as default in our examples/docs, and provide a snippet in the docs that shows usage without decorate for users who don't want another dep.

What do you think?

File: protocol/index.ts (hunk 7 above)
Comment (e-moran) [reply protocol/index.ts:632]: Do you not like its existence, or just how its defined?
I could change it to just `ArcjetDecision & { reason: ArcjetReason}`, or else we'd have to use that definition in every function that requires a reason.

File: protocol/index.ts (hunk 6 above)
Comment (e-moran) [reply protocol/index.ts:594]: They could be useful if someone wants to type-check on a generic ArcjetDecision to see if it has a reason or not.

File: examples/nextjs-clerk-rate-limit/app/api/arcjet/route.ts [hunk 8]
```diff
@@ -69,13 +111,40 @@ export async function GET(req: NextRequest) {
...
+        typeof reason.max !== "number" ||
+        typeof reason.window !== "number" ||
+        typeof reason.remaining !== "number" ||
+        typeof reason.reset !== "number"
+      ) {
+        console.error(format("Invalid rate limit encountered: %o", reason));
+      }
+
+      policies.set(reason.max, reason.window);
+    }
```
blaine-arcjet, line 141: Policies isn't doing anything? Did you want to break out of the logic or something?

examples/nextjs-clerk-rate-limit/package.json
```diff
@@ -10,6 +10,7 @@
   },
   "dependencies": {
     "@arcjet/next": "file:../../arcjet-next",
+    "@arcjet/sprintf": "file:../../sprintf",
```
blaine-arcjet, line 13: Might not be needed if you don't keep the policies stuff

examples/nextjs-openai/app/api/chat/route.ts
```diff
@@ -59,8 +101,40 @@ export async function POST(req: Request) {
...
+      if (
+        typeof reason.max !== "number" ||
+        typeof reason.window !== "number" ||
+        typeof reason.remaining !== "number" ||
+        typeof reason.reset !== "number"
+      ) {
+        console.error(format("Invalid rate limit encountered: %o", reason));
+      }
+
+      policies.set(reason.max, reason.window);
```
blaine-arcjet, line 128: see above

examples/nextjs-openai/package.json
```diff
@@ -11,6 +11,7 @@
   "dependencies": {
     "@ai-sdk/openai": "^1.0.8",
     "@arcjet/next": "file:../../arcjet-next",
+    "@arcjet/sprintf": "file:../../sprintf",
```
blaine-arcjet, line 14: see above

protocol/index.ts
blaine-arcjet: I don't think we need to provide it because they could write their own intersection. Right?

protocol/index.ts
```diff
@@ -627,21 +627,30 @@ export abstract class ArcjetDecision {
...
-    reason: ArcjetReason;
+    reason?: ArcjetReason;
     ip?: ArcjetIpDetails;
   }) {
     super(init);
 
     this.reason = init.reason;
   }
+
+  hasReason(): this is ArcjetAllowDecision & { reason: ArcjetReason } {
```
blaine-arcjet, line 651: 🤔

examples/nextjs-openai/app/api/chat_userid/route.ts
```diff
@@ -64,8 +106,40 @@ export async function POST(req: Request) {
...
+      if (
+        typeof reason.max !== "number" ||
+        typeof reason.window !== "number" ||
+        typeof reason.remaining !== "number" ||
+        typeof reason.reset !== "number"
+      ) {
+        console.error(format("Invalid rate limit encountered: %o", reason));
+      }
+
+      policies.set(reason.max, reason.window);
```
blaine-arcjet, line 133: See above

examples/nextjs-bot-categories/app/api/arcjet/route.ts
```diff
@@ -33,20 +53,27 @@ export async function GET(req: Request) {
...
-    headers.set("X-Arcjet-Bot-Denied", decision.reason.denied.join(", "))
-
-    // We need to check that the bot is who they say they are.
-    if (decision.reason.isSpoofed()) {
-      return NextResponse.json(
-        { error: "You are pretending to be a good bot!" },
-        { status: 403, headers },
-      );
-    }
+  const allowedBots = decision.results.reduce<string[]>(reduceAllowedBots, []);
```
blaine-arcjet, line 56: I noticed you are still using map/filter in the other examples (for rate limit). We should keep it uniform. The reason to use a reduce is to avoid creating extra arrays, but the implementation of the reduction function is creating a new array on each iteration which is worse than a single filter then a map.

protocol/client.ts
```diff
@@ -127,7 +128,7 @@ export function createClient(options: ClientOptions): Client {
     report(
       context: ArcjetContext,
       details: ArcjetRequestDetails,
-      decision: ArcjetDecision,
+      decision: ArcjetDecisionWithReason,
```
blaine-arcjet, line 131: It doesn't need to have a reason, right?

protocol/index.ts
blaine-arcjet, line 1: What happened with also deprecating `reason` on ArcjetErrorDecision?

protocol/index.ts
```diff
@@ -627,21 +627,30 @@ export abstract class ArcjetDecision {
...
 }
 
+export type ArcjetDecisionWithReason = ArcjetDecision & {
+  reason: NonNullable<ArcjetDecision["reason"]>;
+};
+
 export class ArcjetAllowDecision extends ArcjetDecision {
   conclusion = "ALLOW" as const;
-  reason: ArcjetReason;
+  /** @deprecated */
```
blaine-arcjet, line 636: needs a message

protocol/index.ts
blaine-arcjet: I need to double check, but I don't think this needs to be abstract to do that.

blaine-arcjet (changes requested):

examples/nextjs-clerk-rate-limit/app/api/arcjet/route.ts
e-moran: Its used to print an error if two rate limit policies share the same limit, like we do in decorate. The difference is that decorate returns an error in this case. We could respond with a 500 here to have the same behavior. What do you think?

protocol/index.ts
e-moran: Yeah they could! I've removed it

protocol/index.ts
e-moran: I was using this to see if I could come up with a better solution for the types when using reduce, and it stuck around. I'll remove it though since you can just check that reason is not undefined.

examples/nextjs-bot-categories/app/api/arcjet/route.ts
e-moran: I've changed this to a reduce. This removes checking that a limit isn't duplicated which is discussed in another comment though.

blaine-arcjet:
Converting to draft while we discuss this further.

blaine-arcjet:
Closing this since deprecating this value is not a current priority and provides fast-path to users that want to vary their response status codes or messages.

PR #2727, opened 2025-01-06
perf(analyze): Compile WebAssembly upon module load

e-moran:
This change will pre-compile the Wasm files for long-running process when the analyze-wasm module is loaded.

By keeping the Promise results of the compilation as variables at the top of the module, we no longer need a module cache map.

### DIFF
--- a/analyze-wasm/index.ts
+++ b/analyze-wasm/index.ts
@@ -21,6 +21,10 @@ type DetectSensitiveInfoFunction =
 
 const wasmCache = new Map<string, WebAssembly.Module>();
 
+const componentCoreWasmPromise = componentCoreWasm();
+const componentCore2WasmPromise = componentCore2Wasm();
+const componentCore3WasmPromise = componentCore3Wasm();
+
 async function moduleFromPath(path: string): Promise<WebAssembly.Module> {
   const cachedModule = wasmCache.get(path);
   if (typeof cachedModule !== "undefined") {
@@ -28,17 +32,17 @@ async function moduleFromPath(path: string): Promise<WebAssembly.Module> {
   }
 
   if (path === "arcjet_analyze_js_req.component.core.wasm") {
-    const mod = await componentCoreWasm();
+    const mod = await componentCoreWasmPromise;
     wasmCache.set(path, mod);
     return mod;
   }
   if (path === "arcjet_analyze_js_req.component.core2.wasm") {
-    const mod = await componentCore2Wasm();
+    const mod = await componentCore2WasmPromise;
     wasmCache.set(path, mod);
     return mod;
   }
   if (path === "arcjet_analyze_js_req.component.core3.wasm") {
-    const mod = await componentCore3Wasm();
+    const mod = await componentCore3WasmPromise;
     wasmCache.set(path, mod);
     return mod;
   }
--- a/analyze-wasm/index.ts
+++ b/analyze-wasm/index.ts
@@ -19,32 +19,19 @@ import { wasm as componentCore3Wasm } from "./wasm/arcjet_analyze_js_req.compone
 type DetectSensitiveInfoFunction =
   typeof ArcjetJsReqSensitiveInformationIdentifier.detect;
 
-const wasmCache = new Map<string, WebAssembly.Module>();
-
 const componentCoreWasmPromise = componentCoreWasm();
 const componentCore2WasmPromise = componentCore2Wasm();
 const componentCore3WasmPromise = componentCore3Wasm();
 
 async function moduleFromPath(path: string): Promise<WebAssembly.Module> {
-  const cachedModule = wasmCache.get(path);
-  if (typeof cachedModule !== "undefined") {
-    return cachedModule;
-  }
-
   if (path === "arcjet_analyze_js_req.component.core.wasm") {
-    const mod = await componentCoreWasmPromise;
-    wasmCache.set(path, mod);
-    return mod;
+    return componentCoreWasmPromise;
   }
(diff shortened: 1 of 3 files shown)


analyze/index.ts
```diff
@@ -83,6 +83,17 @@ function createCoreImports(detect?: DetectSensitiveInfoFunction): ImportObject {
   };
 }
 
+/**
+ * Pre-warms the wasm cache to improve cold start performance.
+ * @returns Whether the cache was successfully warmed or not.
+ */
+export async function prewarmWasmCache(): Promise<boolean> {
+  const coreImports = createCoreImports();
```
blaine-arcjet, line 91: This doesn't work because you can configure a custom detect function

blaine-arcjet (changes requested):

analyze/index.ts
blaine-arcjet: Initialization is not expensive, the compilation is the expensive part. So you want to call `componentCoreWasm()`, `componentCore2Wasm()`, and `componentCore3Wasm()` when setting up `analyze-wasm/index.ts` and then await their promises inside `moduleFromPath`.

I believe this would allow us to drop the `wasmCache` map because the promises will always be there holding/caching the wasm modules.

e-moran: I've moved the loading to the approach that you've suggested here. Do you think its worth keeping the `.prewarmWasmCache()` export from analyze though? It would allow us to know how long the load is taking for logs, though I would change it to call Promise.all() on the new promises instead of calling `initializeWasm`

analyze-wasm/index.ts
```diff
@@ -21,24 +21,28 @@ type DetectSensitiveInfoFunction =
...
 
+const componentCoreWasmPromise = componentCoreWasm();
+const componentCore2WasmPromise = componentCore2Wasm();
+const componentCore3WasmPromise = componentCore3Wasm();
+
 async function moduleFromPath(path: string): Promise<WebAssembly.Module> {
   const cachedModule = wasmCache.get(path);
   if (typeof cachedModule !== "undefined") {
     return cachedModule;
   }
```
blaine-arcjet, line 32:
```suggestion
```

analyze-wasm/index.ts
```diff
@@ -21,24 +21,28 @@ type DetectSensitiveInfoFunction =
...
 
   if (path === "arcjet_analyze_js_req.component.core.wasm") {
-    const mod = await componentCoreWasm();
+    const mod = await componentCoreWasmPromise;
     wasmCache.set(path, mod);
     return mod;
```
blaine-arcjet, line 37:
```suggestion
    return componentCoreWasmPromise;
```

analyze-wasm/index.ts
```diff
@@ -21,24 +21,28 @@ type DetectSensitiveInfoFunction =
...
   }
   if (path === "arcjet_analyze_js_req.component.core2.wasm") {
-    const mod = await componentCore2Wasm();
+    const mod = await componentCore2WasmPromise;
     wasmCache.set(path, mod);
     return mod;
```
blaine-arcjet, line 42:
```suggestion
    return componentCore2WasmPromise;
```

analyze-wasm/index.ts
```diff
@@ -21,24 +21,28 @@ type DetectSensitiveInfoFunction =
...
   }
   if (path === "arcjet_analyze_js_req.component.core3.wasm") {
-    const mod = await componentCore3Wasm();
+    const mod = await componentCore3WasmPromise;
     wasmCache.set(path, mod);
     return mod;
```
blaine-arcjet, line 47:
```suggestion
    return componentCore3WasmPromise;
```

analyze/index.ts
blaine-arcjet: I think we remove it for now. Since we've removed logging from the wasm due to performance, it's not available in `coreImports` so we are making a clumsy API to try to facilitate this.

The compilation will happen immediately upon import with my requested changes so I don't think tracing is strictly necessary right now.

arcjet/index.ts
```diff
@@ -1240,6 +1240,8 @@ export default function arcjet<
   }
   const log = options.log;
 
+  const wasmPrewarm = analyze.prewarmWasmCache();
```
blaine-arcjet, line 1243:
```suggestion
```

arcjet/index.ts
```diff
@@ -1298,6 +1300,15 @@ export default function arcjet<
...
 
+    const logWasmPrewarmPerf = perf.measure("wasm prewarm");
+    const prewarmed = await wasmPrewarm;
+    if (!prewarmed) {
+      log.debug(
+        "Wasm prewarm failed. This may indicate that Wasm is not supported on this platform",
+      );
+    }
+    logWasmPrewarmPerf();
+
```
blaine-arcjet, line 1311:
```suggestion
```

analyze/index.ts
blaine-arcjet, line 91:
```suggestion
```

analyze/index.ts
```diff
@@ -83,6 +83,17 @@ function createCoreImports(detect?: DetectSensitiveInfoFunction): ImportObject {
...
+  const analyze = await initializeWasm(coreImports);
+
+  return typeof analyze !== "undefined";
+}
+
```
blaine-arcjet, line 96:
```suggestion
```

blaine-arcjet (changes requested):
The cache is not needed since a promise caches indefinitely.

blaine-arcjet (approved):
I cleaned up the code as per my previous review.

PR #2750, opened 2025-01-07
chore(examples): Next.js + Better Auth

davidmytton:
Docs: https://github.com/arcjet/arcjet-docs/pull/356


examples/nextjs-better-auth/auth.ts
```diff
@@ -0,0 +1,137 @@
...
+        github: {
+            clientId: process.env.GITHUB_CLIENT_ID!,
+            clientSecret: process.env.GITHUB_CLIENT_SECRET!,
+        }
+    },
+    hooks: {
+        // Run on every auth request
+        before: createAuthMiddleware(async (ctx) => {
+            let decision: ArcjetDecision;
+            const req = await request(); // Gets the request object
```
davidmytton, line 38: There may be a better way to do this @blaine-arcjet, but [using `ctx.request`](https://www.better-auth.com/docs/concepts/hooks#ctx) didn't match the types so I used our `request` helper instead.

blaine-arcjet: I don't think this is correct. Remember that `request()` is an extremely lacking API that really only exists for server actions which are super limited.

examples/nextjs-better-auth/auth.ts
```diff
@@ -0,0 +1,137 @@
...
+
+            // If the user is logged in we'll use their ID as the identifier. This
+            // allows limits to be applied across all devices and sessions (you could
+            // also use the session ID). Otherwise, fall back to the IP address.
+            let userId: string;
+            if (ctx.context.session?.user.id) {
+                userId = ctx.context.session?.user.id;
+            } else {
+                userId = req.ip || "127.0.0.1"; // Fall back to local IP if none
```
blaine-arcjet, line 47: This will never work correctly because you are using `await request()` which will never have the `ip` field.

examples/nextjs-better-auth/components/SignInGitHub.tsx
```diff
@@ -0,0 +1,16 @@
+'use client'
+
+import { authClient } from "@/auth-client"
+
+export default function SignInButton() {
+    const handleSignIn = async () => {
+        await authClient.signIn.social({
```
blaine-arcjet, line 7: What is the failure case on this? Without a try/catch, this will just crash

examples/nextjs-better-auth/components/SignOut.tsx
```diff
@@ -0,0 +1,20 @@
...
+
+import { authClient } from "@/auth-client"
+import { useRouter } from 'next/navigation'
+
+export default function SignOutButton() {
+    const router = useRouter()
+
+    return (
+        <button onClick={async () => {
+            await authClient.signOut({
```
blaine-arcjet, line 11: Thrown case not handled

examples/nextjs-better-auth/components/SignUp.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { useRouter } from 'next/navigation';
+
+export default function SignUp() {
+    const [email, setEmail] = useState('');
+    const [password, setPassword] = useState('');
+    const [name, setName] = useState('');
+    const router = useRouter();
+
+    const signUp = async () => {
+        const { data, error } = await authClient.signUp.email({
```
blaine-arcjet, line 13: Failure case not handled

examples/nextjs-better-auth/package.json
```diff
@@ -0,0 +1,31 @@
...
+  "dependencies": {
+    "@arcjet/next": "file:../../arcjet-next",
+    "better-auth": "^1.1.10",
+    "better-sqlite3": "^11.7.2",
+    "next": "15.1.0",
+    "react": "^19",
+    "react-dom": "^19"
+  },
+  "devDependencies": {
+    "@types/better-sqlite3": "^7.6.12",
```
blaine-arcjet, line 20: Why is this version number so different? Did they start publishing their own types and this is not needed anymore?

blaine-arcjet (changes requested):
There are some fundamental problems here. I commented inline. Going to follow up if I should prioritize making this work correctly in production environments.

examples/nextjs-better-auth/app/api/auth/[...all]/route.ts
```diff
@@ -0,0 +1,135 @@
...
+  // https://arcjet.com/bot-list
+  allow: [], // prevents bots from submitting the form
+} satisfies BotOptions;
+
+const rateLimitOptions = {
+  // uses a sliding window rate limit
+  mode: "LIVE",
+  interval: "2m", // counts requests over a 10 minute sliding window
+  max: 5, // allows 5 submissions within the window
+} satisfies SlidingWindowRateLimitOptions<[]>;
```
Comment (blaine-arcjet) [inline examples/nextjs-better-auth/app/api/auth/[...all]/route.ts:48]: This requires a fix to export the options. I'm going to cherry-pick the commit into a PR so there is a changelog entry.

examples/nextjs-better-auth/package.json
```diff
@@ -0,0 +1,32 @@
...
+  "dependencies": {
+    "@arcjet/next": "file:../../arcjet-next",
+    "better-auth": "^1",
+    "better-sqlite3": "^11",
+    "next": "^15",
+    "react": "^19",
+    "react-dom": "^19"
+  },
+  "devDependencies": {
+    "@better-auth/cli": "^1",
```
blaine-arcjet, line 20: I added the cli so it can be locked via the lockfile when running the `npx` command.

examples/nextjs-better-auth/package.json
```diff
@@ -0,0 +1,32 @@
...
+    "@types/better-sqlite3": "^7",
+    "@types/node": "^20",
+    "@types/react": "19.0.1",
+    "@types/react-dom": "^19",
+    "autoprefixer": "^10",
+    "eslint": "^8",
+    "eslint-config-next": "^15",
+    "postcss": "^8",
+    "tailwindcss": "^3.4.16",
```
blaine-arcjet, line 29: I'll also apply the tailwind 4 updates

examples/nextjs-better-auth/app/api/auth/[...all]/route.ts
Comment (blaine-arcjet) [reply examples/nextjs-better-auth/app/api/auth/[...all]/route.ts:48]: We already did this. I just need to merge main 🤦

examples/nextjs-better-auth/auth.ts
blaine-arcjet: Moved into the handler

examples/nextjs-better-auth/README.md
```diff
@@ -0,0 +1,50 @@
...
+1. From the root of the project, install the SDK dependencies.
+
+   ```bash
+   npm ci
+   ```
+
+2. Enter this directory and install the example's dependencies.
+
+   ```bash
+   cd examples/nextjs-app-dir-rate-limit
```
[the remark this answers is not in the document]
blaine-arcjet: Applied

examples/nextjs-better-auth/README.md
```diff
@@ -0,0 +1,50 @@
...
+
+   ```bash
+   npm run dev
+   ```
+
+5. Run the database migration. A SQLite database is used for this example.
+
+   ```bash
+   npx @better-auth/cli migrate
+   ```
```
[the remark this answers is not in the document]
blaine-arcjet: Applied

examples/nextjs-better-auth/auth.ts
blaine-arcjet: Moved into the handler.

examples/nextjs-better-auth/components/SignInGitHub.tsx
blaine-arcjet: It'd be better to try/catch and handle errors, but this is just an example so we can leave it.

examples/nextjs-better-auth/components/SignOut.tsx
blaine-arcjet: It'd be better to try/catch and handle errors, but this is just an example so we can leave it.

examples/nextjs-better-auth/components/SignUp.tsx
blaine-arcjet: Response failures are handled via the configuration so I'm not sure what the promise rejecting case would be. It'd be better to try/catch and handle errors, but this is just an example so we can leave it.

examples/nextjs-better-auth/package.json
blaine-arcjet: I couldn't answer this. It looks like no one actually works on the types.

examples/nextjs-better-auth/app/api/auth/[...all]/route.ts
```diff
@@ -0,0 +1,135 @@
...
+  if (session?.user.id) {
+    userId = session.user.id;
+  } else {
+    userId = ip(req) || "127.0.0.1"; // Fall back to local IP if none
+  }
+
+  // If this is a signup then use the special protectSignup rule
+  // See https://docs.arcjet.com/signup-protection/quick-start
+  if (req.nextUrl.pathname.startsWith("/api/auth/sign-up")) {
+    // Better-Auth doesn't clone the body, so we need to clone the request preemptively
```
Comment (blaine-arcjet) [inline examples/nextjs-better-auth/app/api/auth/[...all]/route.ts:76]:
```suggestion
    // Better-Auth reads the body, so we need to clone the request preemptively
```

blaine-arcjet (approved):
This seems to be working now but leaving to @davidmytton to review my changes to ensure it does what he wants.

davidmytton:
@SocketSecurity ignore-all

davidmytton:
Looks good, thanks @blaine-arcjet !

PR #2855, opened 2025-01-14
fix(protocol)!: Improve deprecation message on enum-like field usage

blaine-arcjet:
While reviewing #2715 locally, I noticed that `/** @deprecated */` wasn't being marked in my editor. This is because `/** */` is just a block comment and not a doc comment.

Doc comments need to be in the form as I've changed this PR. Additionally, deprecating the top-level export didn't seem to mark anything in the editor, so each field needed to be marked as deprecated. Furthermore, overriding the type as `ArcjetEnum<>` seemed to lose the doc comments.

I also removed the exports locally and made everything compile once they are removed. They still exist for now but I want the removal to be seamless.


arcjet/test/arcjet.test.ts
```diff
@@ -1302,17 +1299,9 @@ describe("Primitive > validateEmail", () => {
   });
 
   test("allows specifying EmailTypes to deny", async () => {
-    const options = {
```
e-moran, line 1305: Do we want to keep some tests for the old enums until they're actually removed?

e-moran (approved):
Thanks for this! Just have one question but its non-blocking

PR #2863, opened 2025-01-15
fix!: Correctly handle urlencoded data when tokenizing characters

e-moran:
This PR regenerates the wasm to include a fix to how we handle `+` characters in the tokenizer for our sensitive info detection rule.

This introduces a breaking change when using a custom detect function. If the body contains plus symbols it may be called twice for the same characters, once with the `+` characters treated as spaces, and once where they aren't.

Release-As: 1.0.0-beta.1


blaine-arcjet (approved):
LGTM. I'm trying to add `Release-As: x.x.x` to the PR body to see if Release Please will pick it up on merge.

PR #2992, opened 2025-01-27
feat: Implement Astro integration

### LINKED ISSUE #1075: Support for Astro
Astro allows middleware and provides the request information in the `context`. [Per the docs](https://docs.astro.build/en/reference/api-reference/#astrorequest), this is a standard [Request](https://developer.mozilla.org/en-US/docs/Web/API/Request) object.

If we try to pass this through to `aj.protect` there is a type error:

```
Argument of type 'Request' is not assignable to parameter of type 'ArcjetNodeRequest'.
  Types of property 'headers' are incompatible.
    Type 'Headers' is not assignable to type 'Record<string, string | string[] | undefined>'.
      Index signature for type 'string' is missing in type 'Headers'.ts(2345)
(property) AstroSharedContext<Record<string, any>, Params>.request: Request
```

Note that Astro supports [several rendering modes](https://docs.astro.build/en/guides/server-side-rendering/#enable-on-demand-server-rendering), including static (where middleware doesn't make sense) as well as multiple [runtime adapters](https://docs.astro.build/en/guides/deploy/). So we'll need to make it clear that we require server mode otherwise you get this warning:

```
15:09:15 [WARN] `Astro.request.headers` is unavailable in "static" output mode, and in prerendered pages 
within "hybrid" and "server" output modes. If you need access to request headers, make sure that `output` 
is configured as either `"server"` or `output: "hybrid"` in your config file, and that the page accessing the 
headers is rendered on-demand.
```

blaine-arcjet:
This implements an Arcjet adapter for Astro. We do this via an integration that generates a virtual module, such that `arcjet()` gets added to the `integrations` array in `astro.config.mjs` with the normal SDK configuration. The integration then generates a client and types as the `arcjet:client` virtual module. This ensures there is a singleton client in the Astro module graph.

Closes #1075


arcjet-astro/index.ts
```diff
@@ -0,0 +1,292 @@
...
+
+export function protectSignup<Characteristics extends readonly string[]>(
+  options: ProtectSignupOptions<Characteristics>,
+) {
+  return { type: "protectSignup", options } as const;
+}
+
+export default function arcjet<Characteristics extends readonly string[]>({
+  rules,
+}: ArcjetIntegrationOptions<Characteristics>): AstroIntegration {
```
blaine-arcjet, line 158: I need to support more configuration options here.

arcjet-astro/internal.ts
```diff
@@ -0,0 +1,295 @@
...
+      });
+
+  function toArcjetRequest<Props extends PlainObject>(
+    request: Request,
+    props: Props,
+  ): ArcjetRequest<Props> {
+    const clientAddress = Reflect.get(request, ipSymbol);
+    if (!clientAddress) {
+      // TODO: Handle this better
+      throw new Error("Arcjet used in prerendered page");
```
blaine-arcjet, line 212: Do we though? Failing when trying to prerender seems fine. We might add an integration-level fancy logging thing as a feature but this is probably good enough for beta.

blaine-arcjet:
@SocketSecurity ignore npm/emoji-regex-xs@1.0.0
@SocketSecurity ignore npm/ofetch@1.4.1
@SocketSecurity ignore npm/tinyexec@0.3.

arcjet-astro/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+Try an Arcjet protected app live at [https://example.arcjet.com][example-url]
+([source code][example-source]).
+
+## Installation
+
+```shell
+npx astro add @arcjet/astro
+```
+
```
blaine-arcjet, line 39: I still need to document usage/configuration of the integration.

arcjet-astro/index.ts
```diff
@@ -0,0 +1,482 @@
...
+import type { AstroIntegration } from "astro";
+import { z, type ZodType, type ZodTypeDef } from "astro/zod";
+import fs from "node:fs/promises";
+
+const resolvedVirtualClientId = "\0ARCJET_VIRTUAL_CLIENT";
+
+const validateMode = z.enum(["LIVE", "DRY_RUN"]);
+const validateProxies = z.array(z.string());
+const validateCharacteristics = z.array(z.string());
+const validateClientOptions = z
```
e-moran, line 20: Do you think it makes sense to try and re-use the validators from the `arcjet` package here where possible? I'm just asking because it would be nice to minimize the number of packages that require code changes when doing feature work, currently we don't really need to update the adapters all that much.

An alternative could be to move those adapters in to a separate adapters package so that they don't need to be publicly exported from the main sdk. Or if we called it an sdk utils package we could put that `errorMessage()` function in there too.

e-moran (commented):
Looks great! Just have one question

arcjet-astro/README.md
```diff
@@ -0,0 +1,203 @@
...
+## License
+
+Licensed under the [Apache License, Version 2.0][apache-license].
+
+[arcjet]: https://arcjet.com
+[astro]: https://astro.build/
+[example-url]: https://example.arcjet.com
+[quick-start]: https://docs.arcjet.com/get-started/astro
```
blaine-arcjet, line 199: Need to write this because it's 404 currently. Depends upon publishing this.

e-moran (approved):

blaine-arcjet:
> Also, when I started the server without a key I got quite a generic/scary error. If you look closely it explains the problem, but I was wondering if there's a more user friendly error we can throw instead? Or is this all Astro provides?

I think we could enable [`.env.validateSecrets`](https://docs.astro.build/en/reference/configuration-reference/#envvalidatesecrets) via our adapter, but that might be too heavy-handed for all users? Perhaps we just document that it should be set.

blaine-arcjet:
Setting this configuration produces this error at build time:
```
15:51:41 [ERROR] [config] [EnvInvalidVariables] The following environment variables defined in `env.schema` are invalid:

- ARCJET_KEY is missing

  Error reference:
    https://docs.astro.build/en/reference/errors/env-invalid-variables/
```

davidmytton:
Overriding the default is probably not a good idea, so let's document it as a recommendation and also add the error description into our Troubleshooting docs. That way if someone searches, it's easy to find the answer.

examples/astro-integration/README.md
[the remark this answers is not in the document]
blaine-arcjet: Done

davidmytton (approved):

PR #3013, opened 2025-01-29
fix(arcjet): Ensure Characteristics are readonly type on protect signup options

blaine-arcjet:
Cherry-picked from #2992 

This fixes a problem I found where the definition of `Characteristics` was different on `protectSignup`. By aligning the type as `readonly`, the same type can be used everywhere.


e-moran (approved):
Looks good to me, unsure why trunk failed though

PR #3097, opened 2025-02-05
chore: Update trunk, linters, and allowed workflow endpoints

blaine-arcjet:
This updates trunk and the linters. I've also fixed some CI stuff related to Harden Runner's allowed endpoints.

### DIFF
--- a/.github/workflows/reusable-examples.yml
+++ b/.github/workflows/reusable-examples.yml
@@ -187,6 +187,7 @@ jobs:
             fonts.gstatic.com:443
             github.com:443
             registry.npmjs.org:443
+            decide.arcjet.com:443
 
       # Checkout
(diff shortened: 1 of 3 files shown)


.github/workflows/reusable-examples.yml
```diff
@@ -224,6 +230,7 @@ jobs:
             fonts.gstatic.com:443
             github.com:443
             registry.npmjs.org:443
+            decide.arcjet.com:443
```
blaine-arcjet, line 233: I'm not sure why this is be accessed during Next/SvelteKit builds but I've added to each example that needs it. Perhaps it is a url validator or something (like the fonts urls?)

davidmytton (approved):

PR #3130, opened 2025-02-07
chore(examples): TRPC + rate limiting

danni-popova:
Adds an example app of how Arcjet Rate Limiting can be used with TRPC middleware.


examples/nextjs-trpc/package.json
```diff
@@ -0,0 +1,46 @@
...
+    "check": "next lint && tsc --noEmit",
+    "dev": "next dev --turbo",
+    "lint": "next lint",
+    "lint:fix": "next lint --fix",
+    "preview": "next build && next start",
+    "start": "next start",
+    "typecheck": "tsc --noEmit"
+  },
+  "dependencies": {
+    "@arcjet/next": "^1.0.0-beta.2",
```
davidmytton, line 17:
```suggestion
    "@arcjet/next": "file:../../arcjet-next",
```

For these examples we use relative packages because we're already in the repo and we use the latest files to test new versions against all examples before we do a release.

examples/nextjs-trpc/package.json
```diff
@@ -0,0 +1,46 @@
+{
+	"name": "nextjs-trpc",
+	"version": "0.1.0",
+	"private": true,
+	"type": "module",
+	"scripts": {
+		"build": "next build",
+		"check": "next lint && tsc --noEmit",
+		"dev": "next dev --turbo",
```
e-moran, line 9:
```suggestion
		"dev": "next dev",
```

I wasn't able to get this running unless I removed the `--turbo` flag, it was failing to import `@arcjet/next`. It seems that relative imports are known to be broken when using next dev with turbo. 

https://github.com/vercel/next.js/issues/69426

I don't think the solution they propose will work either because its for importing `.ts` files, but its probably fine to just not use turbo here. Curious to hear if @blaine-arcjet has any better solutions though.

e-moran (commented):
This looks really great! I just have a few questions/suggestions

examples/nextjs-trpc/src/server/api/trpc.ts
```diff
@@ -0,0 +1,135 @@
...
+});
+
+/**
+ * Arcjet Rate Limiting Middleware
+ */
+const rateLimitMiddleware = t.middleware(async ({ ctx, next }) => {
+	// Access the request object
+	const request = ctx.req;
+
+	if (request) {
```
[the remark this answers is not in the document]
danni-popova: I cannot access the NextRequest in the "server-only" executed code. Check [trpc/server.ts](https://github.com/arcjet/arcjet-js/blob/feat/nextjs-trpc-example/examples/nextjs-trpc/src/trpc/server.ts) where the other instance of `createContext` is.

examples/nextjs-trpc/package.json
danni-popova: Removed the flag for now.

blaine-arcjet (changes requested):
I still need to pull this down locally and test it out, but please follow https://github.com/arcjet/arcjet-js/blob/main/CONTRIBUTING.md#examples which outlines the need for CI and dependabot additions.

.github/workflows/reusable-examples.yml
```diff
@@ -808,6 +808,48 @@ jobs:
...
+      - name: Harden Runner
+        uses: step-security/harden-runner@cb605e52c26070c328afc4562f0b4ada7618a84e # v2.10.4
+        with:
+          disable-sudo: true
+          egress-policy: block
+          allowed-endpoints: >
+            fonts.googleapis.com:443
+            fonts.gstatic.com:443
+            github.com:443
+            registry.npmjs.org:443
```
blaine-arcjet, line 827:
```suggestion
            registry.npmjs.org:443
            decide.arcjet.com:443
```

it's unclear why Next.js is calling Arcjet endpoints during build but this is just what we have to allow for now.

.github/dependabot.yml
```diff
@@ -721,6 +721,40 @@ updates:
...
+      - dependency-name: "@types/node"
+        versions: [">18.18"]
+      # TODO(#539): Upgrade to eslint 9
+      - dependency-name: eslint
+        versions: [">=9"]
+      # Ignore React during the `rc` period
+      - dependency-name: react
+        versions: ["*"]
+      - dependency-name: react-dom
+        versions: ["*"]
```
blaine-arcjet, line 756: 19 is released. I'll get this cleaned up in the server actions example too
```suggestion
```

examples/nextjs-trpc/package.json
```diff
@@ -0,0 +1,46 @@
...
+		"@arcjet/next": "file:../../arcjet-next",
+		"@t3-oss/env-nextjs": "^0.10.1",
+		"@tanstack/react-query": "^5.50.0",
+		"@trpc/client": "^11.0.0-rc.446",
+		"@trpc/react-query": "^11.0.0-rc.446",
+		"@trpc/server": "^11.0.0-rc.446",
+		"geist": "^1.3.0",
+		"next": "^15.0.1",
+		"react": "^18.3.1",
+		"react-dom": "^18.3.1",
```
blaine-arcjet, line 26: React stuff should be upgraded to 19 since it's using next 15

examples/nextjs-trpc/package.json
```diff
@@ -0,0 +1,46 @@
...
+		"server-only": "^0.0.1",
+		"superjson": "^2.2.1",
+		"zod": "^3.23.3"
+	},
+	"devDependencies": {
+		"@types/eslint": "^8.56.10",
+		"@types/node": "^20.14.10",
+		"@types/react": "^18.3.3",
+		"@types/react-dom": "^18.3.0",
```
blaine-arcjet, line 35: React 19

examples/nextjs-trpc/package.json
```diff
@@ -0,0 +1,46 @@
...
+		"@t3-oss/env-nextjs": "^0.10.1",
+		"@tanstack/react-query": "^5.50.0",
+		"@trpc/client": "^11.0.0-rc.446",
+		"@trpc/react-query": "^11.0.0-rc.446",
+		"@trpc/server": "^11.0.0-rc.446",
+		"geist": "^1.3.0",
+		"next": "^15.0.1",
+		"react": "^18.3.1",
+		"react-dom": "^18.3.1",
+		"server-only": "^0.0.1",
```
blaine-arcjet, line 27: Is `server-only` still needed with Next 15/React 19?

examples/nextjs-trpc/src/app/api/trpc/[trpc]/route.ts
```diff
@@ -0,0 +1,35 @@
...
+		headers: req.headers,
+	});
+};
+
+const handler = (req: NextRequest) =>
+	fetchRequestHandler({
+		endpoint: '/api/trpc',
+		req,
+		router: appRouter,
+		createContext: () => createContext(req),
```
Comment (blaine-arcjet) [inline examples/nextjs-trpc/src/app/api/trpc/[trpc]/route.ts:24]: Is this supposed to be wrapped in a function?

examples/nextjs-trpc/src/lib/arcjet.ts
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+import arcjet, { slidingWindow } from '@arcjet/next';
+
+export const aj = arcjet({
+	key: process.env.ARCJET_KEY!,
```
blaine-arcjet, line 6: Why is this using `process.env` instead of the env helper stuff from t3-app?

examples/nextjs-trpc/src/lib/arcjet.ts
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
blaine-arcjet, line 2:
```suggestion
```
See below, I'm guessing these can go away if you use the env helper stuff from t3 app but I don't know enough about it.

examples/nextjs-trpc/src/server/api/routers/post.ts
```diff
@@ -0,0 +1,44 @@
...
+		.mutation(async ({ input }) => {
+			const post: Post = {
+				id: posts.length + 1,
+				name: input.name,
+			};
+			posts.push(post);
+			return post;
+		}),
+
+	getLatest: publicProcedure.query(() => {
```
blaine-arcjet, line 41: I noticed that you used prefetch on `getLatest` in the application. Prefetch is a footgun when you are protecting with Arcjet because it'll essentially issue requests twice. However, `getLatest` isn't protected by Arcjet so the prefetch isn't going to cause exhaustion quickly.

examples/nextjs-trpc/src/server/api/trpc.ts
```diff
@@ -0,0 +1,141 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
```
blaine-arcjet, line 3: Why disable these?

examples/nextjs-trpc/src/server/api/trpc.ts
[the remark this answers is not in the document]
blaine-arcjet: This seems very confusing/concerning for me. Server code should be the only the place request would be available. It is not available on the client (and if it were, it could be faked).

examples/nextjs-trpc/src/trpc/react.tsx
```diff
@@ -0,0 +1,76 @@
+"use client";
+
+import { QueryClientProvider, type QueryClient } from "@tanstack/react-query";
+import { loggerLink, unstable_httpBatchStreamLink } from "@trpc/client";
+import { createTRPCReact } from "@trpc/react-query";
+import { type inferRouterInputs, type inferRouterOutputs } from "@trpc/server";
```
blaine-arcjet, line 6:
```suggestion
import type { inferRouterInputs, inferRouterOutputs } from "@trpc/server";
```
Just some nice cleanup

examples/nextjs-trpc/src/trpc/react.tsx
```diff
@@ -0,0 +1,76 @@
...
+import { useState } from "react";
+import SuperJSON from "superjson";
+
+import { type AppRouter } from "~/server/api/root";
```
blaine-arcjet, line 10:
```suggestion
import type { AppRouter } from "~/server/api/root";
```

blaine-arcjet (changes requested):
Some thoughts. I'm going to pull this down to try to understand why a request would not be available server side.

blaine-arcjet:
Alright, so I think I found the problem with the context. Create T3 App is not structured correctly for the Next.js App Router. If we look at the [tRPC App Dir Example](https://github.com/trpc/examples-next-app-dir), we can see that they separate the context into the [server directory](https://github.com/trpc/examples-next-app-dir/blob/main/src/server/context.ts). This is much the same file structure as the T3 App's [Pages Router structure](https://create.t3.gg/en/folder-structure-pages?packages=trpc).

I think we want to either create this example as a Pages Router example or restructure the App Router example like the official [tRPC example](https://github.com/trpc/examples-next-app-dir), even if it is marked as experimental.

blaine-arcjet:
Diving deeper: there does seem to be a similar file called [rq-server.ts](https://github.com/trpc/examples-next-app-dir/blob/20f6ea082501b459204f119941a94d623eebf7db/src/trpc/rq-server.ts). However, I still am suspicious of this pattern because trpc brags about the type safety.

examples/nextjs-trpc/src/app/api/trpc/[trpc]/route.ts
Comment (blaine-arcjet) [inline examples/nextjs-trpc/src/app/api/trpc/[trpc]/route.ts:24]: I figured this question out. We should be doing something like this:
```suggestion
import {
  type FetchCreateContextFnOptions,
  fetchRequestHandler,
} from "@trpc/server/adapters/fetch";
import { type NextRequest } from 'next/server';

import { env } from '~/env';
import { appRouter } from '~/server/api/root';
import { createTRPCContext } from '~/server/api/trpc';

/**
 * This wraps the `createTRPCContext` helper and provides the required context for the tRPC API when
 * handling a HTTP request (e.g. when you make requests from Client Components).
 */
const createContext = async (opts: FetchCreateContextFnOptions) => {
  return createTRPCContext({
    req: opts.req,
    headers: opts.req.headers,
  });
};

const handler = (req: NextRequest) =>
	fetchRequestHandler({
		endpoint: '/api/trpc',
		req,
		router: appRouter,
		createContext,
```

examples/nextjs-trpc/src/server/api/trpc.ts
blaine-arcjet:
```suggestion
```
I didn't encounter build errors when I removed these comments

examples/nextjs-trpc/src/server/api/trpc.ts
```diff
@@ -0,0 +1,141 @@
...
+ * These allow you to access things when processing a request, like the database, the session, etc.
+ *
+ * This helper generates the "internals" for a tRPC context. The API handler and RSC clients each
+ * wrap this and provides the required context.
+ *
+ * @see https://trpc.io/docs/server/context
+ */
+export const createTRPCContext = async (opts: {
+	headers: Headers;
+	req?: NextRequest;
```
blaine-arcjet, line 25:
```suggestion
	req?: Request;
```

examples/nextjs-trpc/src/server/api/trpc.ts
```diff
@@ -0,0 +1,141 @@
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
+/* eslint-disable @typescript-eslint/no-unsafe-call */
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+
+import { initTRPC, TRPCError } from '@trpc/server';
+import superjson from 'superjson';
+import { ZodError } from 'zod';
+import { type NextRequest } from 'next/server';
```
blaine-arcjet, line 8:
```suggestion
```

examples/nextjs-trpc/src/server/api/trpc.ts
```diff
@@ -0,0 +1,141 @@
...
+	console.log(`[TRPC] ${path} took ${end - start}ms to execute`);
+
+	return result;
+});
+
+/**
+ * Arcjet Rate Limiting Middleware
+ */
+const rateLimitMiddleware = t.middleware(async ({ ctx, next }) => {
+	// Access the request object
```
blaine-arcjet, line 101:
```suggestion
	if (!ctx.req) {
		throw new TRPCError({
			code: "BAD_REQUEST",
		});
	}

	// Access the request object
```

examples/nextjs-trpc/src/server/api/trpc.ts
```diff
@@ -0,0 +1,141 @@
...
+	const request = ctx.req;
+
+	if (request) {
```
blaine-arcjet, line 104:
```suggestion
```

examples/nextjs-trpc/src/server/api/trpc.ts
```diff
@@ -0,0 +1,141 @@
...
+					code: 'TOO_MANY_REQUESTS',
+					message: "You've hit a rate limit!",
+				});
+			} else {
+				throw new TRPCError({
+					code: 'BAD_REQUEST',
+				});
+			}
+		}
+	}
```
blaine-arcjet, line 126:
```suggestion
```

.github/workflows/reusable-examples.yml
```diff
@@ -808,6 +808,48 @@ jobs:
...
+      - name: Install dependencies
+        run: npm ci
+
+      - name: Install example dependencies
+        working-directory: examples/nextjs-trpc
+        run: npm ci
+
+      - name: Build
+        working-directory: examples/nextjs-trpc
+        run: npm run build
```
blaine-arcjet, line 851:
```suggestion
        run: npm run build
        env:
          # Skip the T3 environment validation during our test builds
          SKIP_ENV_VALIDATION: true
```

examples/nextjs-trpc/src/env.js
```diff
@@ -0,0 +1,42 @@
...
+import { z } from 'zod';
+
+export const env = createEnv({
+	/**
+	 * Specify your server-side environment variables schema here. This way you can ensure the app
+	 * isn't built with invalid env vars.
+	 */
+	server: {
+		NODE_ENV: z.enum(['development', 'test', 'production']),
+		ARCJET_KEY: z.string().startsWith('ajkey_').optional(),
```
blaine-arcjet, line 11:
```suggestion
		ARCJET_KEY: z.string().startsWith('ajkey_'),
```
This doesn't need to be optional if you set the `SKIP_ENV_VALIDATION` environment variable in CI.

examples/nextjs-trpc/src/lib/arcjet.ts
blaine-arcjet: My suggestion seems to work. Commenting above.
```suggestion
	key: env.ARCJET_KEY,
```

examples/nextjs-trpc/src/lib/arcjet.ts
```diff
@@ -0,0 +1,14 @@
...
+import arcjet, { slidingWindow } from '@arcjet/next';
+
```
blaine-arcjet, line 4:
```suggestion

import { env } from "~/env";

```

blaine-arcjet:
I'm going to collapse my comments about the tRPC Context but I do want to follow up to say I really hate the way tRPC tries to make Context global across client and server—it just leads to sloppy code that doesn't clear indicate **where** the code is running. Since this is just something the library forces, we have to accept the pattern but I made some suggestions that make things a tiny bit safer.

blaine-arcjet:
Regarding the Socket warnings:
1. tRPC is super sketchy in that they are actively abusing manifest confusion to write hashes into their package.json version but publishing without them. This is scary because they could inject an attack via manifest confusion and it could slip through since their package already abuses it.
2. ljharb is a good maintainer so those are fine
3. You need to investigate the OFL-1.1 license. I don't have experience to know if it is okay to include deps with that license.

blaine-arcjet:
@SocketSecurity ignore npm/async-function@1.0.0

examples/nextjs-trpc/src/lib/arcjet.ts
blaine-arcjet: @danni-popova is there any reason you didn't accept the changes to this file?

danni-popova: No reason, I haven't pushed my latest commit

blaine-arcjet:
I've marked this as Draft until @danni-popova has some time to push the latest changes

PR #3258, opened 2025-02-19
feat: add custom detection of bots

e-moran:
This PR adds the ability for user's to add custom bot detection to their `detectBot` rule.


analyze/index.ts
```diff
@@ -45,13 +46,27 @@ const FREE_EMAIL_PROVIDERS = [
...
-  if (typeof detect !== "function") {
-    detect = noOpDetect;
+function noOpDetectBots(): string[] {
+  return [];
+}
+
+function createCoreImports(imports?: {
+  detectSensitiveInfo?: DetectSensitiveInfoFunction;
+  detectBots?: DetectBotsFunction;
+}): ImportObject {
```
blaine-arcjet, line 60: Any reason to build an object here? Just trying to consider how many extra objects could be created on heavy traffic endpoints.

arcjet/index.ts
```diff
@@ -515,19 +516,43 @@ export type SlidingWindowRateLimitOptions<
   max: number;
 };
 
-export type BotOptionsAllow = {
+// This represents the User-facing implementation where the request is an object.
+type DetectBot<T> = (request: ArcjetRequest<{}>) => Array<T>;
+
+// This represents the Wasm-facing implentation where the request is JSON.
+type DetectBotSerialized<T> = (request: string) => Array<T>;
```
blaine-arcjet, line 523: Why isn't this handled in the Analyze wrapper?

arcjet/index.ts
```diff
@@ -515,19 +516,43 @@ export type SlidingWindowRateLimitOptions<
...
+
+type ValidBotEntities<Detect> = Array<
+  // Via https://www.reddit.com/r/typescript/comments/17up72w/comment/k958cb0/
+  // Conditional types distribute over unions. If you have ((string | undefined)
+  // extends undefined ? 1 : 0) it is evaluated separately for each member of
+  // the union, then union-ed together again. The result is (string extends
+  // undefined ? 1 : 0) | (undefined extends undefined ? 1 : 0) which simplifies
+  // to 0 | 1
+  undefined extends Detect
```
blaine-arcjet, line 532: I don't think I ever got `undefined extends` to work correctly, I'll poke at this a bit

arcjet/index.ts
```diff
@@ -1242,14 +1285,21 @@ export function shield(options: ShieldOptions): Primitive<{}> {
...
+  bots: BotOptions<DetectBot>;
   email: EmailOptions;
 };
 
-export function protectSignup<const Characteristics extends string[] = []>(
-  options: ProtectSignupOptions<Characteristics>,
+export function protectSignup<
+  const Detect extends DetectBot<CustomEntities> | undefined,
+  const CustomEntities extends string,
+  const Characteristics extends string[] = [],
```
blaine-arcjet, line 1300: Changing the order of generics would be a breaking page because the function can be called with explicit types.

I guess adding new generics that aren't defaulted can also be considered breaking.

arcjet/test/arcjet.test.ts
```diff
@@ -627,6 +639,141 @@ describe("Primitive > detectBot", () => {
...
+      detect: () => {
+        return ["CUSTOM_BOT_A" as const];
+      },
+    });
+    expect(rule.type).toEqual("BOT");
+    assertIsLocalRule(rule);
+    const result = await rule.protect(context, details);
+    expect(result).toMatchObject({
+      state: "RUN",
+      conclusion: "DENY",
```
blaine-arcjet, line 723: This test isn't working as described in the test name

blaine-arcjet (changes requested):
You marked it draft right as I finished this review 😅

arcjet/index.ts
e-moran: I had to change the order because you can't have non-defaulted after defaulted. I'm thinking that it might be best to default them for both these functions to avoid a breaking change. Is there any reason not to?

analyze/index.ts
e-moran: What do you think of just removing the optionals altogether here? We end up needing to allocate the no-ops anyway, so we can save on runtime type checking by just requiring them

blaine-arcjet: Can we use defaulted arguments instead?

analyze/index.ts
```diff
@@ -131,13 +138,32 @@ export async function isValidEmail(
...
+        throw new Error("object sent for detection was not a request");
+      }
+      return detect(request);
+    };
+  }
+
+  const coreImports = createCoreImports(
+    noOpDetectSensitiveInfo,
+    convertedDetect || noOpDetectBots,
+  );
```
blaine-arcjet, line 166:
```suggestion
  let convertedDetect = noOpDetectBots;
  if (typeof detect === "function") {
    convertedDetect = (req: string) => {
      let request: AnalyzeRequest;
      try {
        request = JSON.parse(req);
      } catch {
        throw new Error("object sent for detection was not a request");
      }
      return detect(request);
    };
  }

  const coreImports = createCoreImports(
    noOpDetectSensitiveInfo,
    convertedDetect,
  );
```
Will this work with the types?

analyze/index.ts
```diff
@@ -131,13 +138,32 @@ export async function isValidEmail(
   }
 }
 
+type RequestDetectBotsFunction = (request: AnalyzeRequest) => Array<string>;
+
 export async function detectBot(
   context: AnalyzeContext,
   request: AnalyzeRequest,
   options: BotConfig,
+  detect?: RequestDetectBotsFunction,
```
blaine-arcjet, line 147: Why is this a separate argument instead of provided via BotConfig? I forget the shape of BotConfig

analyze/index.ts
```diff
@@ -162,7 +188,7 @@ export async function detectSensitiveInfo(
   detect?: DetectSensitiveInfoFunction,
 ): Promise<SensitiveInfoResult> {
   const { log } = context;
-  const coreImports = createCoreImports(detect);
+  const coreImports = createCoreImports(detect || noOpDetectSensitiveInfo);
```
blaine-arcjet, line 191:
```suggestion
  const coreImports = createCoreImports(detect);
```
Doesn't need this if defaulted, right?

arcjet/index.ts
blaine-arcjet, line 523:
```suggestion
```

arcjet/index.ts
blaine-arcjet: I'm not sure if it would break inference. Worth trying out

arcjet/test/arcjet.test.ts
```diff
@@ -627,6 +639,141 @@ describe("Primitive > detectBot", () => {
...
+      query: "",
+      extra: {
+        "extra-test": "extra-test-value",
+      },
+    };
+
+    const [rule] = detectBot({
+      mode: "LIVE",
+      deny: ["CUSTOM_BOT_A"],
+      detect: () => {
```
blaine-arcjet, line 759: Can you update the tests to assert on the request? I'd like to know things are being deserialized correctly.

analyze/index.ts
e-moran: Its because `BotConfig` is generated by jco. So we'd need another type.

analyze/index.ts
e-moran: Yes this works!

arcjet/test/arcjet.test.ts
e-moran: Yeah I've added these in

PR #3455, opened 2025-03-04
feat: Add `@arcjet/inspect` package with utilities to inspect decisions

blaine-arcjet:
I've created a new package called `@arcjet/inspect` that provides utilities for inspecting decisions and the rule results provided on them.

The initial utilities are `isSpoofedBot`, `isVerifiedBot`, and `isMissingUserAgent` which were chosen because we've seen users encounter problems with them. Previously, we've documented the verbose way to handle these cases, but we'll be able to streamline to show these utilities.

It's worth noting that I've made all these utilities ignore `DRY_RUN` results because users wouldn't want dry run rules to be used in informing more detailed decisions. If users are trying to log information, they should iterate the results and use the type-narrowing functions attached to each result.


davidmytton (approved):
Looks good. I've added in doc comments with some examples. I think the `isVerified` example could be better though. Perhaps @e-moran or @blaine-arcjet have some ideas?

inspect/index.ts
```diff
@@ -0,0 +1,213 @@
...
+ *
+ * @link https://docs.arcjet.com/bot-protection/reference#bot-verification
+ */
+export function isVerifiedBot(result: ArcjetRuleResult): boolean {
+  // Use `unknown` argument helpers to guard around the wrong data being passed
+  if (isActive(result) && isBotReason(result.reason)) {
+    return result.reason.isVerified();
+  }
+
+  return false;
```
e-moran, line 158: Returning `false` when a non-bot rule is passed in feels a little bit off to me, but I'm not sure what the best solution is and I've been thinking on it for a few minutes now.

The potential footgun that I see here is that someone could have the following config:
```ts
const aj = arcjet({
   key: process.env.ARCJET_KEY,
   rules: [
     shield({ mode: "LIVE" }),
     detectBot({
       mode: "LIVE",
       allow: ["GOOGLE_CRAWLER"],
     }),
   ],
  });
```
and then remove their bot rule (or else they could update their rules remotely when we support that). Their users would then get the 403 error.

I wonder if its more appropriate to return `undefined` when a non-bot rule result is passed in. Returning `false` feels like we're proactively stating that it is unverified. Like in Rust I'd probably represent this with an `Option<bool>`.

inspect/index.ts
```diff
@@ -0,0 +1,213 @@
...
+ * ```ts
+ * import arcjet, { detectBot } from "@arcjet/next";
+ * import { isSpoofedBot } from "@arcjet/inspect";
+ *
+ * const aj = arcjet({
+ *  key: process.env.ARCJET_KEY,
+ *  rules: [
+ *    detectBot({
+ *      mode: "LIVE",
+ *      allow: [],
```
e-moran, line 70:
```suggestion
 *      allow: ["GOOGLE_CRAWLER", "BING_CRAWLER"],
```

It might be best to include some bots in the allow list, because we only run verification on bots that we are explicitly allowing.

inspect/index.ts
blaine-arcjet: Returning undefined will not work with typescript types for `Array#find`, `Array#some` or any others that need a boolean returned.

blaine-arcjet: I just checked the typescript stdlib that matches our version and it looks like the definition of `Array#some` allows the predicate to return `unknown`. The definition of `Array#find` is an assertion so it may allow undefined, I'll have to check.

blaine-arcjet: `Array#find` seems to work with undefined, so I'm making this change across the utilities. I'm also adding tests to verify that `Array#some` and `Array#find` work on lists of results.

inspect/index.ts
```diff
@@ -0,0 +1,213 @@
...
+    typeof result === "object" &&
+    result !== null &&
+    "state" in result &&
+    typeof result.state === "string" &&
+    result.state !== "DRY_RUN"
+  );
+}
+
+/**
+ * Determines if a `LIVE` bot rule detected a spoofed request. If `true`, the
```
blaine-arcjet, line 47: This is wrong. Rule results are not `LIVE`, they are either `RUN`, `DRY_RUN`, `CACHED`, or `NOT_RUN` and we consider `DRY_RUN` as the only non-active result.

We could restrict that to `RUN` and `CACHED` since `NOT_RUN` is unclear if it is active.

blaine-arcjet: Note that we don't currently use `CACHED` since we don't cache individual rule results. They'd be marked as `NOT_RUN` if a top-level `DENY` decision has been cached.

blaine-arcjet: That's verbose but fine. Using "active" like I had is less committal in case we switch the implementation to exclude `NOT_RUN` or something.

blaine-arcjet: If we explicitly mention "non-DRY_RUN" then we can probably remove the "DRY_RUN rules are ignored"

blaine-arcjet: I also think we should loosen the "you should block it" to "you might want to block it" because we flag things as spoofed when they come through a proxy, which is the whole point of not blocking them by default.

inspect/index.ts
```diff
@@ -0,0 +1,213 @@
...
+ * });
+ *
+ * export async function GET(request: Request) {
+ *  const decision = await aj.protect(request);
+ *
+ *  if (decision.isDenied()) {
+ *    return res.status(403).json({ error: "Forbidden" });
+ *  }
+ *
+ *  if (!decision.results.some(isVerifiedBot)) {
```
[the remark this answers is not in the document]
blaine-arcjet: This doesn't work because every request is going to have a rule result for that includes the bot rule being run and normal clients are not bots.

inspect/index.ts
davidmytton: Sounds good. I wanted to be explicit in exactly how it works because a user reading it without context might not have fully understood, and an AI definitely wouldn't.

inspect/index.ts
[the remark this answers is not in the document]
blaine-arcjet: A better example would be to allow any verified bots through the system (probably even before checking the `decision.isDenied()`.

I think that's when you'd want to use explicit verification checks.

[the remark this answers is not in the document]
blaine-arcjet: e.g. allow all verified search engines and ignore any rate limit signals to avoid giving them a 403

inspect/index.ts
blaine-arcjet: I've changed the implementation so I'm re-requesting review

inspect/index.ts
blaine-arcjet: I made all these `"CATEGORY:SEARCH_ENGINE"` because that's what we have in most docs.

inspect/index.ts
blaine-arcjet: I rewrote the docblocks significantly so I'm re-requesting review

inspect/index.ts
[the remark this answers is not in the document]
blaine-arcjet: I rewrote with the above as the goal and am re-requesting review.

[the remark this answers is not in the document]
davidmytton: I like how this turned out with almost like a bypass for bots we know we want.

inspect/index.ts
davidmytton: All looks good. It's a bit complex, but the examples make it easy to see how it works.

inspect/index.ts
```diff
@@ -149,36 +159,43 @@ export function isSpoofedBot(result: ArcjetRuleResult): boolean {
  *
  * @link https://docs.arcjet.com/bot-protection/reference#bot-verification
  */
-export function isVerifiedBot(result: ArcjetRuleResult): boolean {
+export function isVerifiedBot(result: ArcjetRuleResult): boolean | undefined {
   // Use `unknown` argument helpers to guard around the wrong data being passed
   if (isActive(result) && isBotReason(result.reason)) {
     return result.reason.isVerified();
```
e-moran, line 165: I know this is already merged, but I was just thinking that maybe we should return undefined if the bot reason is errored as well. What do you think @blaine-arcjet?

PR #3465, opened 2025-03-05
chore: Change doc comments key examples

davidmytton:
Makes the examples consistent with our others by using an environment variable and ensures anyone copying the example isn't encouraged to hard code their key.


davidmytton:
That was my original reasoning, but on second thought I don't want someone (human or AI) to just copy/paste and put the secret in code. `process.env` is common enough that if it's not exactly right for the developer, they will know how to change it.

Issue #539, opened 2024-04-09
Upgrade to eslint 9


blaine-arcjet:
Eslint v9 made breaking changes to the way things are configured. This is a fairly big change to propagate across the ecosystem and into our workspace and examples.

We're going to exclude eslint 9+ from our dependabot updates until we resolve the above.

blaine-arcjet:
And upgrade typescript-eslint packages to v8

blaine-arcjet:
This should be completed except in the few examples that seem soft-deprecated or are using an older version.

PR #3554, opened 2025-03-12
deps(example): Upgrade NestJS examples to eslint 9

blaine-arcjet:
This upgrades our NestJS examples to eslint v9. It doesn't seem like the NestJS CLI provides the ability to upgrade templates but I copied their `eslint.config.mjs` file from their newest template.

Towards #539


examples/nestjs-launchdarkly/src/config/arcjet.ts
```diff
@@ -26,29 +26,31 @@ export class ArcjetConfig {
...
 
     // Get the latest configuration from LaunchDarkly
-    const shieldMode = await client.variation(
+    const shieldMode = (await client.variation(
       'shieldMode',
       context,
       ArcjetConfig.defaultConfig.shieldMode,
-    );
-    const rateLimitMode = await client.variation(
+    )) as 'LIVE' | 'DRY_RUN';
```
e-moran, line 33: Is there any way we can avoid this cast, or does LD need to improve this?

e-moran (approved):
Looks good to me, just have one non-blocking question in the ld example

Issue #3706, opened 2025-03-27
SvelteKit adapter doesn't work with Vercel edge runtime


davidmytton:
Using Arcjet with the SvelteKit adapter in the default auto mode works ok, however it fails if you switch the config to use the edge runtime.

To reproduce:

`npm i -D @sveltejs/adapter-vercel`

```js
// svelte.config.js
import adapter from '@sveltejs/adapter-vercel';
import { vitePreprocess } from '@sveltejs/vite-plugin-svelte';

/** @type {import('@sveltejs/kit').Config} */
const config = {
	// Consult https://svelte.dev/docs/kit/integrations
	// for more information about preprocessors
	preprocess: vitePreprocess(),

	kit: {
		// adapter-auto only supports some environments, see https://svelte.dev/docs/kit/adapter-auto for a list.
		// If your environment is not supported, or you settled on a specific environment, switch out the adapter.
		// See https://svelte.dev/docs/kit/adapters for more information about adapters.
		adapter: adapter({ runtime: 'edge' })
	}
};

export default config;
```

```ts
// lib/server/arcjet/index.ts

import { ARCJET_KEY } from '$env/static/private';
import arcjet, { detectBot, shield, slidingWindow, validateEmail } from '@arcjet/sveltekit';

export const middleware_aj = arcjet({
    key: ARCJET_KEY,
    rules: [
        shield({
            mode: 'LIVE'
        }),
        detectBot({
            mode: 'LIVE',
            allow: ['CATEGORY:SEARCH_ENGINE']
        }),
        slidingWindow({
            mode: 'LIVE',
            interval: 15,
            max: 25
        })
    ]
});

export const email_aj = arcjet({
    key: ARCJET_KEY,
    rules: [
        validateEmail({
            mode: "LIVE",
            deny: ["DISPOSABLE", "INVALID", "NO_MX_RECORDS", "FREE"],
        }),
    ],
});
```

```ts
// src/hooks.server.ts
import { redirect, type Handle } from '@sveltejs/kit';
import { middleware_aj } from '$lib/server/arcjet';
import { sequence } from '@sveltejs/kit/hooks';

const middleware: Handle = async ({ event, resolve }) => {

    const path = event.url.pathname;

    const decision = await middleware_aj.protect(event);

    if (decision.isDenied()) {
        if (decision.reason.isRateLimit()) {
            return redirect(303, '/rate-limited?redirect=' + path);
        }
        //,,,
    }

[...]

PR #3715, opened 2025-03-27
trying to understand vercel adapter

blaine-arcjet:
Draft of explorations toward #3706


blaine-arcjet:
Closing this as we don't actually have a solution.

PR #3873, opened 2025-04-11
feat: add plaintext param to custom redact

e-moran:


redact/index.ts
```diff
@@ -29,7 +29,10 @@ export type RedactOptions<Detect> = {
   entities?: ValidEntities<Detect>;
   contextWindowSize?: number;
   detect?: Detect;
-  replace?: (entity: ValidEntities<Detect>[number]) => string | undefined;
+  replace?: (
+    entity: ValidEntities<Detect>[number],
+    plaintext: string,
```
blaine-arcjet, line 34: I wonder if we can be clever with types to make sure old implementations satisfy the type signature. Perhaps a union of the old function signature and the new one?

blaine-arcjet: I missed that you wrote the comment:
> // This is because we introduced `plaintext` in a non-breaking way.
> // This test is to ensure that we don't make a change which breaks it in the future.

This is exactly what I wanted, which indicates this will behave non-breaking when users upgrade 🎉 

I think we're good to go then!

blaine-arcjet (approved):
This needed regen but it is good now.

Issue #3884, opened 2025-04-14
Add code formatting rules


luigimorel:
Working on an issue and I have to literally disable some features in VS Code that enhance my code. Would be nice to have rules that can be picked up by the editor that lint and format the code before push.

blaine-arcjet:
Please don't try to add commit hooks to the project. I've adjusted the issue to only mention formatting rules.

luigimorel:
Ok. Could you push the editor rules that you have/currently use?

blaine-arcjet:
Please note that using Devcontainers is the recommended workflow. As you can see at https://github.com/arcjet/arcjet-js/blob/main/.devcontainer/devcontainer.json, this installs trunk which makes `trunk check` and `trunk fmt` available. It'll also install the trunk extension which lints and formats on save.

luigimorel:
Thank you... 
Maybe, it's time to add clear contributor docs.

PR #3885, opened 2025-04-14
refactor: duplicate error message

luigimorel:
- De-duplicates the errorMessage function across all the packages. 
- I need to work on linting before this can get merged.

### DIFF
--- a/error-utils/index.ts
+++ b/error-utils/index.ts
@@ -1,0 +1,17 @@
+export function errorMessage(err: unknown): string {
+  if (err) {
+    if (typeof err === "string") {
+      return err;
+    }
+
+    if (
+      typeof err === "object" &&
+      "message" in err &&
+      typeof err.message === "string"
+    ) {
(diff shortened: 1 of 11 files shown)


error-utils/index.ts
blaine-arcjet, line 1: Since all our packages live in the `@arcjet/` scope, we are very careful with creating new packages. We probably want to come up with another solution.

blaine-arcjet (commented):
Thanks @luigimorel! I don't think we want to add another package to achieve this, which makes it much more difficult to solve.

I also noticed you opened #3884 but in the meantime, we need PRs to only have the subset of changes being changed (no formatting changes). Perhaps you want to stage your hunks with `git add -p`.

luigimorel:
Closing this. Will just redo it

blaine-arcjet:
Okay. I created #3921 to track my initial thoughts and other difficulties that would come from the proposed solution.

PR #3997, opened 2025-04-29
chore: Enable declaration maps for TypeScript definition files

qw-in:
Was having a poke around and noticed [go to definition](https://code.visualstudio.com/docs/editing/editingevolved#_go-to-definition) was targeting declaration files. Here is a quick patch that enables [declaration maps](https://www.typescriptlang.org/tsconfig/#declarationMap).

Only logic is in the rollup config (see below comment). The rest is `.gitignore` patching for the additional generated map files (I tried to follow the existing pattern)

Feel free to close if you have them intentionally disabled. Had a quick search of the repo and couldn't find anything :)

### DIFF
--- a/rollup-config/index.js
+++ b/rollup-config/index.js
@@ -185,6 +185,7 @@ export function createConfig(root, { plugins = [] } = {}) {
         // generate definition files for our tests
         exclude: ["node_modules", "test/*.ts"],
         declaration: true,
+        declarationMap: true,
         declarationDir: ".",
         noEmitOnError: true,
       }),
--- a/analyze-wasm/.gitignore
+++ b/analyze-wasm/.gitignore
@@ -132,9 +132,12 @@ dist
 # Generated files
 edge-light.js
 edge-light.d.ts
+edge-light.d.ts.map
 index.js
 index.d.ts
+index.d.ts.map
 workerd.js
 workerd.d.ts
+workerd.d.ts.map
 test/*.js
 _virtual/*.js
--- a/analyze/.gitignore
+++ b/analyze/.gitignore
@@ -132,9 +132,12 @@ dist
 # Generated files
 edge-light.js
 edge-light.d.ts
+edge-light.d.ts.map
 index.js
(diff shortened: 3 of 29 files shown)


rollup-config/index.js
```diff
@@ -185,6 +185,7 @@ export function createConfig(root, { plugins = [] } = {}) {
         // generate definition files for our tests
         exclude: ["node_modules", "test/*.ts"],
         declaration: true,
+        declarationMap: true,
```
qw-in, line 188: Relevant change

blaine-arcjet:
Thanks @qw-in! How does this work with the published packages? We've stopped shipping the `.ts` files in published packages because some tools redirect `.js` imports to `.ts` which caused some users to have problems when using different TypeScript versions. If these maps point to non-existent files, I'm worry they will fail to "go to" anywhere.

qw-in:
Ah, good question @blaine-arcjet!

Sorry I had intended to include a note that due to the existing "files" configuration in the `package.json`s it is my understanding that this wouldn't change the published packages at all.

https://github.com/arcjet/arcjet-js/blob/5bd5f5278ccd88a8f484fe84256125a24b92eec0/arcjet/package.json#L27-L33

The cases I can think of to validate would be:
1. Emitting the map files changes the `.d.ts` files in some way
2. Certain packages are missing a "files" key

I'll sanity check those now

qw-in:
Ah, annoyingly it does append a `//# sourceMappingURL=index.d.ts.map` to the declaration file. I'll have a look into that

qw-in:
Apologies, thought this was an easy change 😅

Looks like the options are:
1. Close this and pretend it never happened 😀
2. Merge and ship the declaration files with a source-map comment appended
3. More invasive build system changes (whole can of worms)

I unfortunately can't definitively say 2 won't break things as different editors etc. may implement missing source-map behaviour differently. I validated the fallback behaviour in vscode but you never know

I'd say option 1 for now, cheers

PR #4027, opened 2025-05-01
feat(protocol): Support identifier and version on rules

blaine-arcjet:
This lays the groundwork of support for Rule IDs in the SDK. It updates the protobuf definitions and threads `version` and `rule_id` through the SDK.

It also implements the logic for stable, deterministic, and unique identifiers for local rules. These are mostly unused right now but are useful for double-checking the local and remote Rule IDs are exactly the same. 

These locally generated IDs will be needed for looking up cached decisions in the new cache implementation. Rules that don't currently generate an ID will get them added once they have a local implementation that reads from cache.


arcjet/index.ts
```diff
@@ -50,6 +50,140 @@
...
+    data.writeString(value.toFixed(0));
+  };
+}
+
+function string(key: string, value: string): fieldHasher {
+  return (data: StringWriter) => {
+    data.writeString(key);
+    data.writeString(fieldSeparator);
+    data.writeString(`"`);
+    data.writeString(value.replaceAll(`"`, `\"`));
```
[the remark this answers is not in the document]
blaine-arcjet: I'm pretty sure CodeQL just has no concept of the backtick character in JS because making this change results in the exact same hash. I'm fine changing it but it just seems unnecessary.

arcjet/index.ts
```diff
@@ -50,6 +50,140 @@
...
+}
+
+function stringSliceOrdered(key: string, values: string[]): fieldHasher {
+  return (data: StringWriter) => {
+    data.writeString(key);
+    data.writeString(fieldSeparator);
+    data.writeString("[");
+    for (const value of Array.from(values).sort()) {
+      data.writeString(`"`);
+      data.writeString(value.replaceAll(`"`, `\"`));
```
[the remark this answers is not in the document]
blaine-arcjet: I'm pretty sure CodeQL just has no concept of the backtick character in JS because making this change results in the exact same hash. I'm fine changing it but it just seems unnecessary.

stable-field-hash/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@arcjet/stable-field-hash",
```
blaine-arcjet, line 2: I was grasping for descriptive names and wanted something longer for uniqueness (and avoid any potential collisions in our scope). Open to other suggestions.

davidmytton: Since you're describing it as `Stable hashing utility for Arcjet rule identifiers` then it could be `rule-hash`? Just to make it shorter, but that's a nit and I think this is fine.

davidmytton (approved):

stable-field-hash/package.json
blaine-arcjet: Maybe I should change the description because it is a stable hash mechanism for any key/value pairs. We just use it for rule IDs currently 🤔

blaine-arcjet: Renamed to `@arcjet/stable-hash` and simplified description.

PR #4278, opened 2025-06-02
fix(astro): add actionable error message for fully static sites

### LINKED ISSUE #3094: Astro: Redirects cause adapter to fail
It seems that redirects defined in `astro.config.mjs` cause our adapter to fail. This is due to the `context.isPrerendered` flag being set to `false` yet `context.clientAddress` throws when accessed. This generally makes sense because a redirect doesn't produce a static page itself.

Astro doesn't seem to have anything like `isRedirect` on the context but you can discover redirects during the build process. I experimented with using codegen to make an `isDynamic` function that consumed the discovered redirects but I'm not sure I'm happy with users needed to import the utility explictly.

Instead, we could use this generated logic to warn and produce an `Allow` decision on static pages. Alternatively, we could produce an `Error` decision like we do when IP isn't defined. I'm not sure which is better.

qw-in:
Closes #3094 

In my testing, we only run into #3094 when there are no dynamic routes. In that case `protect` is never being called anyway.

If the error in #3094 is encountered should now always be proceeded by this error message (I think the screenshot is up to date with the message but you get the idea):
Screenshot 2025-06-02 at 2 04 11 PM

cc @arcjet/design-team for unicode char use


blaine-arcjet:
I'd have to dig back into Astro but I thought you could "opt-in" to dynamic pages even when you use the "static" output as the default build target.

If that's the case, then this doesn't work.

qw-in:
@blaine-arcjet `buildOutput` is the effective output taking hybrid rendering bits into account

blaine-arcjet:
So Astro still errors if you use redirects to static files, `static` output configuration, and opt-in to a dynamic route?

qw-in:
@blaine-arcjet not in my testing. As soon as you opt in to dynamic rendering `buildOutput` is a server (astro forces you to add an adapter). The redirects are then fine (which I think is what I was hitting in that commit with the test cases)

qw-in:
Certainly worth a sanity check though, I've just been modifying the `examples/astro-integration` for quick tests

qw-in:
Yeah, double checked:

configured redirect, `static`, & no adapter:

Screenshot 2025-06-02 at 2 35 09 PM

configured redirect, `static`, no adapter, added `test.astro` with `export const prerender = false;`:

Screenshot 2025-06-02 at 2 35 57 PM

Same as above but with adapter added:

Screenshot 2025-06-02 at 2 36 10 PM

blaine-arcjet:
Thanks for the screenshots. I'm pretty sure I played around with adapters in arcjet-docs to test this with the redirects over there and hit the same problem. It's interesting that your "with adapter" doesn't show any redirects being generated (which would seem to indicate they make them all dynamic 🙈)

qw-in:
@blaine-arcjet @wooorm-arcjet Do we want to keep this in some form? I can think of two possible reasons:

1. Support previous Astro versions
2. Avoid user confusion where Arcjet can't protect things given the fully static output

davidmytton:
> Avoid user confusion where Arcjet can't protect things given the fully static output

For this reason yes. If it can be a warning rather than an error then that's better, but either way we should keep it.

blaine-arcjet:
> For this reason yes. If it can be a warning rather than an error then that's better, but either way we should keep it.

This can be an interim solution. During my enumeration of other integrations, I found they will build a "report" showing what was done by the plugin. The Arcjet integration should build a report that shows which routes are protected. We will need a follow up issue.

arcjet-astro/index.ts
```diff
@@ -452,7 +452,16 @@ export default function arcjet<Characteristics extends readonly string[]>(
           order: "pre",
         });
       },
-      "astro:config:done": async ({ injectTypes }) => {
+      "astro:config:done": async ({ buildOutput, injectTypes, logger }) => {
+        if (buildOutput === "static") {
+          logger.error(
```
blaine-arcjet:
```suggestion
          logger.warn(
```

.github/workflows/reusable-examples.yml
```diff
@@ -1038,3 +1038,46 @@ jobs:
...
+      # Environment security
+      - name: Harden Runner
+        uses: step-security/harden-runner@0634a2670c59f64b4a01f0f96f84700a4088b9f0 # v2.12.0
+        with:
+          disable-sudo-and-containers: true
+          egress-policy: block
+          allowed-endpoints: >
+            github.com:443
+            registry.npmjs.org:443
+            decide.arcjet.com:443
```
blaine-arcjet:
```suggestion
```

blaine-arcjet:
#4431 created to replace this temporary warning.

qw-in:
Thanks @blaine-arcjet!

I can't check the Trunk merge box. Before I try the trunk comment wanted to make sure I _should_ merge this? Sorry if I missed docs somewhere (read `CONTRIBUTING.md`)

blaine-arcjet:
> I can't check the Trunk merge box.

I assume you don't have `Maintain` permisssions on this repository which allows editing comments. Probably check with @davidmytton about that since I'm not sure if the comment will work either. I can merge this for now (as we generally do for outside contributions).

qw-in:
Sure feel free to merge! I don't think I need maintain perms right now

Issue #4410, opened 2025-06-12
Add Node 22 and 24 to test matrix


blaine-arcjet:
In #4406, the build process was fixed on Node 24. We didn't see this because I forgot to add Node 22 and 24 to our CI matrix. We should add them to ensure everything works.

PR #4414, opened 2025-06-13
ci: Add Node.js 22, 24

wooorm-arcjet:
Closes GH-4410.

### DIFF
--- a/.github/workflows/reusable-test.yml
+++ b/.github/workflows/reusable-test.yml
@@ -28,6 +28,7 @@ jobs:
           disable-sudo-and-containers: true
           egress-policy: block
           allowed-endpoints: >
+            api.github.com:443
             github.com:443
             nodejs.org:443
             registry.npmjs.org:443
--- a/.github/workflows/reusable-test.yml
+++ b/.github/workflows/reusable-test.yml
@@ -28,11 +28,11 @@ jobs:
           disable-sudo-and-containers: true
           egress-policy: block
           allowed-endpoints: >
-            api.github.com:443
+            *.githubusercontent.com:443
+            *.github.com:443
             github.com:443
             nodejs.org:443
             registry.npmjs.org:443
-            release-assets.githubusercontent.com:443
 
       # Checkout
       # Most toolchains require checkout first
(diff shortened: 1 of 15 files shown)


wooorm-arcjet:
Q: `.trunk` L62?

https://github.com/arcjet/arcjet-js/blob/main/.trunk/trunk.yaml#L62

Does that need changing?

davidmytton:
> Q: `.trunk` L62?
> 
> https://github.com/arcjet/arcjet-js/blob/main/.trunk/trunk.yaml#L62
> 
> Does that need changing?

Yes - we'll need to add in the new versions so that the status checks include them.

wooorm-arcjet:
@davidmytton alright so some security mechanism is blocking https://github.com/arcjet/arcjet-js/actions/runs/15676502192/job/44157939791?pr=4414, I am not sure how to allow it?

davidmytton:
Harden Runner is blocking the download ([see here](https://github.com/arcjet/arcjet-js/runs/44033667447)) because the URL isn't on the allowlist:

https://github.com/arcjet/arcjet-js/blob/1853b1ac9f6da42b7ac9852273426ea53df67017/.github/workflows/reusable-test.yml#L30-L32

You can add the domain to the list and then it should work.

.github/workflows/reusable-test.yml
```diff
@@ -28,7 +28,10 @@ jobs:
           disable-sudo-and-containers: true
           egress-policy: block
           allowed-endpoints: >
+            *.githubusercontent.com:443
+            *.github.com:443
```
blaine-arcjet:
```suggestion
            api.github.com:443
```
I'm not sure there should be any requests to `github.com` that isn't `api.github.com` or `github.com`. Did you see another?

wooorm-arcjet: ~~The replacement would be *both* `github.com` and `api.github.com`: the glob matches zero-or-one(/more?)~~. (nope I was wrong, I checked that, it isn’t)
And then also `release-assets.githubusercontent.com` and then also `objects.githubusercontent.com`.
So, the list was just becoming long. Which is why I looked for globs and found they are possible.

wooorm-arcjet: I’ll just go with https://github.com/arcjet/gravity/pull/72/files, explicit, no globs

PR #4415, opened 2025-06-13
chore: remove `expect`, references to `jest`

wooorm-arcjet:
* I saw references to jest, and decided to remove them.
* I have not removed all use of `expect` too

***!!! please open the test files which GH truncates !!!***

/cc @blaine-arcjet


arcjet/test/arcjet.test.ts
```diff
@@ -869,11 +807,15 @@ describe("Primitive > tokenBucket", () => {
...
     const rules = tokenBucket(options);
-    expect(rules).toHaveLength(1);
-    expect(rules[0].type).toEqual("RATE_LIMIT");
-    expect(rules[0]).toHaveProperty("refillRate", 60);
-    expect(rules[0]).toHaveProperty("interval", 60);
-    expect(rules[0]).toHaveProperty("capacity", 120);
+    assert.equal(rules.length, 1);
+    const rule = rules[0];
+    assert.equal(rule.type, "RATE_LIMIT");
+    // @ts-expect-error: to do: fix types to allow access of properties.
```
blaine-arcjet: Does the type of `assert.equal` apply type narrowing? I've been let down by every JS assertion library I've tried in that they don't narrow type checking on future calls.

blaine-arcjet: We typically follow the pattern of creating an issue and then adding the comment:
```
// TODO(#<issue number>): ...
```

This allows us to have common pattern to grep when resolving an issue.

arcjet/test/arcjet.test.ts
```diff
@@ -3882,22 +3714,22 @@ describe("SDK", () => {
...
-        path: request.path,
-        headers: new Headers(Object.entries(request.headers)),
-        extra: {
-          "extra-test": "extra-test-value",
-        },
-      }),
-      [],
-    ]);
+    assert.equal(client.decide.mock.callCount(), 1);
+    const argument: unknown = [...client.decide.mock.calls[0].arguments].at(1);
```
blaine-arcjet: Is `.arguments` the actual `arguments` object instead of an array? Interesting choice by the `node:test` folks.

arcjet/test/arcjet.test.ts
```diff
@@ -3998,25 +3828,25 @@ describe("SDK", () => {
...
-          "extra-false": "false",
-          "extra-true": "true",
-          "extra-unsupported": "<unsupported value>",
-        },
-      }),
-      [],
-    ]);
+    assert.equal(client.decide.mock.callCount(), 1);
+    const argument: unknown = [...client.decide.mock.calls[0].arguments].at(1);
+    assert.deepEqual(requestAsJson(argument), {
```
blaine-arcjet: I like the pattern you switched to below where the `.at()` was used here instead of the line above. Did you change the pattern for a specific reason?

blaine-arcjet (approved):
This is really awesome. I had some questions for my own education and I gave a tip about how we generally handle TODOs in code.

Once those are resolved, feel free to merge.

arcjet/test/arcjet.test.ts
wooorm-arcjet: It should, if TS can figure out that the `type` field is a discriminating union.
This works in my ASTs: our type for `rule` here would be a union of all possible type interfaces, `.type` would be the unique differentiator between them, and then TS can figure out that this is a “RateLimitRule”.

So, this points to an improvement that I would like to work towards, which I believe possible!

wooorm-arcjet: Right!

Thought I’d write down some reasoning for how I do things is:

- `ts-expect-error` ultimately also means a `todo` to me
- I use `To do` because tools like linters pick up `TODO`, but not `To do`
- I go through `To do`s in code / `expect-error`s before cutting a release
- I also put `To do` comments above URLs that point to a specific version or date or so: `To do: once in a while, update URL`
- Or, `To do: next major: change to $someNewApi`
- There is some balance between what is a real issue and needs to be on the board, and what can remain in the code

arcjet/test/arcjet.test.ts
wooorm-arcjet: it is not 👀 I assumed that because `arguments`.

arcjet/test/arcjet.test.ts
wooorm-arcjet: I think the reason is just that for the previous tests, *all* values were really tested.
Such as that `context` (argument 0) is the same as the given context.
But most tests are almost the same. So a lot of unneeded detail was tested.
I guess I noticed that somewhere and started focussing on what I think is the important part.

I think I applied what you prefer. Is there still a noticeable divide somewhere?

arcjet/test/arcjet.test.ts
wooorm-arcjet: This is a short intro, using our types, perhaps that helps: https://unifiedjs.com/learn/recipe/narrow-node-typescript/.
Note the difference between `Node` (an abstract interface that everything extends) and `Nodes` (a union of concrete things with different values for `type`)

wooorm-arcjet:
My main open question is how to update all the package locks: `"expect"` can be removed from all `package.json`s now. And then? What do I do?

arcjet/test/arcjet.test.ts
wooorm-arcjet: also, ehh, what’s being tested here blaine?! https://github.com/arcjet/arcjet-js/blob/ef3a123d372d5c2006ff120d0a02287e0d28af41/decorate/test/decorate.test.ts#L1703

blaine-arcjet: > also, ehh, what’s being tested here blaine?! 

This seems to be testing that `setRateLimitHeaders` is a noop if provided with an object that looks like a Node `OutgoingResponse` but missing the `hasHeader` function missing for some reason. The expectation could probably be better 🙈

blaine-arcjet:
> My main open question is how to update all the package locks: "expect" can be removed from all package.jsons now. And then? What do I do?

Once `expect` is removed in every package, I think we can remove via `npm i --package-lock-only` to update the lockfile. If the command alone doesn't work, you can manually chop the `expect` entry and then run the command again.

body/test/body.test.ts
```diff
@@ -44,9 +44,9 @@ describe("reads the body from the readable stream", () => {
     const server = http.createServer(async (req, res) => {
       try {
         await readBody(req, { limit: 4 });
-        throw new Error("this should not return successfully");
+        assert.fail("this should not return successfully");
       } catch (err) {
-        expect(err).toEqual(new Error("request entity too large"));
+        assert.equal(String(err), "Error: request entity too large");
```
blaine-arcjet, line 49: Do we want to assert `err` is an instance of `Error` before casting to a string to compare the message?

body/test/body.test.ts
```diff
@@ -44,9 +44,9 @@ describe("reads the body from the readable stream", () => {
     const server = http.createServer(async (req, res) => {
       try {
         await readBody(req, { limit: 4 });
-        throw new Error("this should not return successfully");
+        assert.fail("this should not return successfully");
```
blaine-arcjet, line 47: Nice!

arcjet/test/arcjet.test.ts
```diff
@@ -3715,8 +3715,9 @@ describe("SDK", () => {
 
     const decision = await aj.protect(context, request);
     assert.equal(client.decide.mock.callCount(), 1);
-    const argument: unknown = [...client.decide.mock.calls[0].arguments].at(1);
-    assert.deepEqual(requestAsJson(argument), {
+    const d = client.decide.mock.calls[0].arguments;
```
blaine-arcjet:
```suggestion
```
Leftover I think

arcjet/test/arcjet.test.ts
blaine-arcjet: Yeah, I love narrowing. I was mostly wondering if the `@types/node` defined all the functions in the assert stdlib with `asserts x ...` so they actually narrow.

blaine-arcjet: We should document this somewhere if that's the pattern you plan to follow for the SDK since it differs from the way we do things in other parts of Arcjet. Perhaps the contributor guidelines?

> ts-expect-error ultimately also means a todo to me

This is not always the case because we write our SDK in TypeScript but have to assume it is used by JavaScripters that don't follow the types. Lots of the code and tests exist solely for the fact that we can't trust the TS types we've defined.

> I use To do because tools like linters pick up TODO, but not To do

I don't think we have those linters, but I might be wrong.

> I go through To dos in code / expect-errors before cutting a release

By release, do you mean merging a PR (such as this) to main, or the publishing of a versioned release? Generally we want to track additional work via issues so @arcjet-rei can keep track of everything that still needs to get done to provide accurate timelines.

> There is some balance between what is a real issue and needs to be on the board, and what can remain in the code

Definitely agree! As outlined above, we should make sure to balance visibility for folks like @arcjet-rei and your personal workflow. This is mostly solved by adding some documentation somewhere to clarify expectations.

wooorm-arcjet: > We should document this somewhere if that's the pattern you plan to follow for the SDK since it differs from the way we do things in other parts of Arcjet. Perhaps the contributor guidelines?

:+1: I was not nessecarily pushing my opinion through. I think I can go with the existing style. Whether it needs to change is for later.
But, wanted to talk about the ideas.

> This is not always the case because we write our SDK in TypeScript but have to assume it is used by JavaScripters that don't follow the types. Lots of the code and tests exist solely for the fact that we can't trust the TS types we've defined.

Especially in the tests that is the case, indeed.

For other cases, could you not also *accept* a strictly typed value as a parameter, then cast it to things?
Typically that type would be `unknown`, and then you’d use the regular type system to narrow things again.

First example I found was `directivesToSvelteKitConfig`, from:

```ts
    const key = CONTENT_SECURITY_POLICY_DIRECTIVES.get(
      // @ts-expect-error because we're validating this option key
      optionKey,
    );
    if (!key) {
      throw new NoseconeValidationError(
        `${optionKey} is not a Content-Security-Policy directive`,
      );
    }
```

Suggested:

```ts
    type Cspds = typeof CONTENT_SECURITY_POLICY_DIRECTIVES extends Map<infer I, any> ? I : never
    // Cast because user may provide a value that is not allowed by the types.
    const presumablyKey = optionKey as Cspds
    const key = CONTENT_SECURITY_POLICY_DIRECTIVES.get(presumablyKey);
    if (!key) {
      throw new NoseconeValidationError(
        `${optionKey} is not a Content-Security-Policy directive`,
      );
    }
```

> By release

...i mostly work on low-release-frequency stuff. Where you check out a code base occasionally to land a PR and twice a year do a release.

I hear the stuff about visibility

wooorm-arcjet: Used the `TODO(#1234)` style

body/test/body.test.ts
wooorm-arcjet: I think this is fine!

At some point I’d go with custom errors. Perhaps classes/instances, but at least error codes. And then a check for `RateLimitError: Something` could be enough.

I also think you *must* enforce that things that arcjet throws are instances of `Error`. And you can assume that all reasonable open source code only throws errors too. Now, perhaps, it is good to handle users throwing arbitrary things.

I also do not like `instanceof` much, because of a) different realms, but practically b) `npm` may put different copies of a class in different locations in `node_modules`, they would likely be equivalent enough, but not `instanceof`

arcjet/test/arcjet.test.ts
wooorm-arcjet: Right! To a point: the TS type system doesn’t cover as much as Node’s assertions. So, I try and keep assertions simple.

arcjet/test/arcjet.test.ts
wooorm-arcjet: alright, this can be improved at some point, in another PR

wooorm-arcjet:
> Once expect is removed in every package, I think we can remove via `npm i --package-lock-only` to update the lockfile. If the command alone doesn't work, you can manually chop the expect entry and then run the command again.

`niplo` doesn’t do anything for me; from the docs, I think that just treats the lock as the source of truth to install, instead of updating the lock.

I manually chopped things, and `niplo` still doesn’t do things for me.

I can do `npm update --package-lock-only`, but then it installs updates too.


I manually cut `express` from package.json, package lock, bun lock for now. There will probably be subdependencies cut when you update things.

blaine-arcjet:
I checked out this branch and went into the `package-lock.json` file to remove this entry:
```json
"node_modules/expect": {
      "version": "29.7.0",
      "resolved": "https://registry.npmjs.org/expect/-/expect-29.7.0.tgz",
      "integrity": "sha512-2Zks0hf1VLFYI1kbh0I5jP3KHHyCHpkfyHBzsSXRFgl/Bg9mWYfMW8oD+PdMPlEwy5HNsR9JutYy6pMeOh61nw==",
      "dev": true,
      "license": "MIT",
      "dependencies": {
        "@jest/expect-utils": "^29.7.0",
        "jest-get-type": "^29.6.3",
        "jest-matcher-utils": "^29.7.0",
        "jest-message-util": "^29.7.0",
        "jest-util": "^29.7.0"
      },
      "engines": {
        "node": "^14.15.0 || ^16.10.0 || >=18.0.0"
      }
    },
```

I then ran `npm i --package-lock-only` and it regenerated the lockfile. I will push that here for you.

arcjet/test/arcjet.test.ts
blaine-arcjet: > I was not nessecarily pushing my opinion through. I think I can go with the existing style. Whether it needs to change is for later. But, wanted to talk about the ideas.

Understandable. I'm happy to adjust things, we just need to document them.

> For other cases, could you not also accept a strictly typed value as a parameter, then cast it to things?
> Typically that type would be unknown, and then you’d use the regular type system to narrow things again.

We want to use concrete types for the TypeScript users and various IDE tools. However, there are some cases in the codebase where the value is treated as `unknown` type once inside the function. Such as the `validateX` functions in the SDK.

blaine-arcjet (approved):
Looks great and some great conversations!

wooorm-arcjet:
Note to self that Trunk does not receive events for “resolving” conversations. When those are first open, then resolved, Trunk will not know.

wooorm-arcjet:
Note to Blaine, the flakey test resulted in a failure. As it was flakey, I clicked “re-run failed jobs”. That also somehow failed because GH says files are missing? https://github.com/arcjet/arcjet-js/actions/runs/15753451524?pr=4415

wooorm-arcjet:
Finally!

Issue #4339, opened 2025-06-09
Packages don't link to their subdirectory


blaine-arcjet:
On the npm website, our packages don't link to their individual directory (such as [`@arcjet/inspect`](https://www.npmjs.com/package/@arcjet/inspect)). I thought I had this configured correctly in the `package.json` file but it seems not.

wooorm-arcjet:
I think that (`directory`) just doesn’t work (anymore?)

See the `package.json`s of e.g.
* https://www.npmjs.com/package/@types/mdast?activeTab=code
* https://www.npmjs.com/package/@babel/core?activeTab=code

Both use `homepage` to link to the package in question.

blaine-arcjet:
I definitely didn't realize that packages moved their directory links into the `homepage` field 🙈 Yeah, this seems broken in the npm UI (somebody probably forgot to re-add it in their tailwind rewrite).

Do you think we should link to the source code via homepage or should we keep linking to the arcjet docs?

wooorm-arcjet:
There is something to say for “repository” going to the repository, and not necessarily to a deeper folder. 🤷‍♂️

I think it is fine this way: repo and homepage do what they say they do.

An alternative is to add stuff to the readme, sort of like this: https://github.com/wooorm/markdown-rs#links. A link to “source”?

blaine-arcjet:
A links section with a link to the package source in the readme sounds pretty good 👍

PR #4429, opened 2025-06-16
docs: Add relevant links to each package readme

wooorm-arcjet:
Closes GH-4339.

/cc @blaine-arcjet this is an example in one package, if this format is ✅ then I can apply it to everything

### DIFF
--- a/analyze-wasm/README.md
+++ b/analyze-wasm/README.md
@@ -22,6 +22,9 @@ against common attacks.
 
 This package provides WebAssembly bindings to [Arcjet's][arcjet] local analysis engine.
 
+- [npm package (`@arcjet/analyze-wasm`)](https://www.npmjs.com/package/@arcjet/analyze-wasm)
+- [GitHub source code (`analyze-wasm/` in `arcjet/arcjet-js`)](https://github.com/arcjet/arcjet-js/tree/main/analyze-wasm)
+
 ## Installation
 
 ```shell
--- a/arcjet-astro/README.md
+++ b/arcjet-astro/README.md
@@ -22,6 +22,9 @@ against common attacks.
 
 This is the [Arcjet][arcjet] SDK integration for [Astro][astro].
 
+- [npm package (`@arcjet/astro`)](https://www.npmjs.com/package/@arcjet/astro)
+- [GitHub source code (`arcjet-astro/` in `arcjet/arcjet-js`)](https://github.com/arcjet/arcjet-js/tree/main/arcjet-astro)
+
 ## Getting started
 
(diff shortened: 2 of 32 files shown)


analyze/README.md
```diff
@@ -22,6 +22,11 @@ against common attacks.
 
 This is the [Arcjet][arcjet] local analysis engine.
 
+## Links
+
+- [npm package (`@arcjet/analyze`)](https://www.npmjs.com/package/@arcjet/analyze)
+- [GitHub source code (`analyze/` in `arcjet/arcjet-js`)](https://github.com/arcjet/arcjet-js/tree/main/analyze)
```
blaine-arcjet, line 26: I think it is fine to point at `main` here since we don't have some sort of preprocessor on the markdown, but a future improvement might be to preprocess so we can link to the the tag instead of `main`.

analyze/README.md
```diff
@@ -22,6 +22,11 @@ against common attacks.
 
 This is the [Arcjet][arcjet] local analysis engine.
 
+## Links
```
blaine-arcjet: I like them near the top, but using `h2` pulls my eyes instead of the installation docs. Do we want to keep an `h2` or reduce to `h3`? If keeping an `h2`, we may want to move them under the `## Installation` section 🤔

blaine-arcjet (commented):
Looks good but I had a couple thoughts.

analyze/README.md
wooorm-arcjet: I do think links make more sense *above* `## Install`.
I can also remove the heading: paragraph explaining arcjet, then the paragraph explaining this specific package, list with links, `## Install`.
I’ll do just that.

blaine-arcjet:
This is still a draft so it is unclear if it needs follow up from me.

wooorm-arcjet:
Yes, it does still need a review:
- In the OP I wrote “if this format is ✅ then I can apply it to everything”.
- You had some comments
- I commented on those comments
When that’s :+1: to you I apply it to every readme

analyze/README.md
```diff
@@ -22,6 +22,9 @@ against common attacks.
 
 This is the [Arcjet][arcjet] local analysis engine.
 
+- [npm package (`@arcjet/analyze`)](https://www.npmjs.com/package/@arcjet/analyze)
+- [GitHub source code (`analyze/` in `arcjet/arcjet-js`)](https://github.com/arcjet/arcjet-js/tree/main/analyze)
```
blaine-arcjet, line 26: I like this without the heading. The only thing that felt out of place is the bulleted list but I don't have a better way to display it. 👍

Issue #1836, opened 2024-10-02
Consider removing TS files from published packages


blaine-arcjet:
Some bundlers and build tools opt to load `.ts` files instead of the `.js` and `.d.ts` files we publish. This causes problems when TypeScript is configured differently than we configure it.

We should enumerate the advantages and disadvantages of including the `.ts` files and decide if we want to stop publishing them.

PR #4441, opened 2025-06-17
refactor: Clean `files` fields in `package.json`s

wooorm-arcjet:
* files such as `licese`, `package.json`, and `readme.md` are always included in files
* values were copied over from each other, `_virtual` and `wasm` folders that did not exist were included
* sort files
* less globs if not needed


wooorm-arcjet:
For @blaine-arcjet:

Should [`astro-env.d.ts`](https://github.com/arcjet/arcjet-js/blob/main/arcjet-astro/astro-env.d.ts) be published? It was. To me it seems like not?
Should [`wasm.d.ts`](https://github.com/arcjet/arcjet-js/blob/main/redact-wasm/wasm.d.ts) be published? It is from the WASM packages. I would imagine that such globals should be left to the final end user? They will start to conflict? I have now removed them.

Depends on #4415 for the removal of `jest.config.js`

blaine-arcjet:
> files such as licese, package.json, and readme.md are always included in files

While this has always been the case, I always worry that npm will decide to remove the defaults in some major version to simplify their code (or some other reason). I'd rather be explicit that sorry. Thoughts?

> Should [astro-env.d.ts](https://github.com/arcjet/arcjet-js/blob/main/arcjet-astro/astro-env.d.ts) be published? It was. To me it seems like not?

It looks like this is used by the `internal.ts` file which is read as the [compiled `internal.js` file](https://github.com/arcjet/arcjet-js/blob/main/arcjet-astro/index.ts#L418) so the TypeScript types should only be necessary for our development experience. This probably doesn't need to be published.

> Should [wasm.d.ts](https://github.com/arcjet/arcjet-js/blob/main/redact-wasm/wasm.d.ts) be published? It is from the WASM packages. I would imagine that such globals should be left to the final end user? They will start to conflict? I have now removed them.

I think removing them is correct. I think this was needed when we shipped `.ts` files but now everything should be compiled to JS + `.d.ts` files so we shouldn't publish it.

analyze-wasm/package.json
```diff
@@ -30,13 +30,10 @@
...
   },
   "files": [
-    "LICENSE",
-    "README.md",
     "_virtual/",
     "wasm/",
-    "*.js",
     "*.d.ts",
-    "!*.config.js"
+    "*.js"
```
blaine-arcjet: This change will publish the rollup config, right?

analyze/package.json
```diff
@@ -25,13 +25,8 @@
...
   "files": [
-    "LICENSE",
-    "README.md",
-    "_virtual/",
-    "wasm/",
-    "*.js",
-    "*.d.ts",
-    "!*.config.js"
+    "index.d.ts",
+    "index.js"
```
blaine-arcjet, line 37: I was using `*.js` to ensure proper publishes when a new file was added. I almost always forget to update the `files` entry when adding a new file.

How should we solve this error if we are explicitly listing all files?

wooorm-arcjet:
> While this has always been the case, I always worry that npm will decide to remove the defaults in some major version to simplify their code (or some other reason). I'd rather be explicit that sorry. Thoughts?

It’s an understandable point but I don’t think it’s realistic. Like, I also advocate for `npm init` to add `type: module` but they think that would be breaking too many things too 🤷
I can’t see npm stopping with license info or readmes, to me there seem to be too many tools using that, it seems too useful.
And then the argument that it would break a million existing packages.

> this is used by the `internal.ts`

Right, `astro:env/server`, which is registered in `astro-env.d.ts`, is.
But, my assumption here was that in a normal Astro project, Astro itself would define such Astro things.
That we have it here because we’re not in a normal Astro project.
So, I am not 100%, but I would imagine the same things as `wasm.d.ts` apply here.

> so we shouldn't publish it.

:+1:

analyze-wasm/package.json
wooorm-arcjet: good catch! done

analyze/package.json
wooorm-arcjet: Right! I think this is a) different today than it was before, and b) I am planning on improving that in the next PR.

For a), that was different before export maps. When everything-was-public-api-by-default. Now everything-is-private-by-default. So if you’d add some new file and publish, you‘d immediately find that it doesn’t work.

(With export maps you could again open things up with globs, but I am against that.)

And, with export maps, it doesn’t matter where things are on the file system! A file could be in the root, or in a folder.

And that’s b), I’d like to move to a `lib/` folder, which is published. Point to that from export maps. That simplifies things: normally, everything in that folder is published, everything outside (config, tests, `wasm.d.ts`) isn’t

(Some more adjacent background reasoning: before export maps things were more about files, e.g., if you have an `index.js`, then that could be accessed from `analyze-wasm`, `analyze-wasm/index`, `analyze-wasm/index.js`; Now things are more like `node:assert` and `node:assert/strict` and it doesn’t matter where that code lives on the file system)

blaine-arcjet: > I’d like to move to a lib/ folder, which is published. Point to that from export maps. That simplifies things: normally, everything in that folder is published, everything outside (config, tests, wasm.d.ts) isn’t

I'm fine with this as long as we are still generate the `.js` + `.d.ts` files next to the `.ts` files; however, we cannot publish the `.ts` files, as [we've found that some types of tooling](https://github.com/arcjet/arcjet-js/issues/1836) tries to load/compile the `.ts` file if it exists. I think this would mean you still need to use globs against the `lib` directory.

blaine-arcjet:
> It’s an understandable point but I don’t think it’s realistic. Like, I also advocate for npm init to add type: module but they think that would be breaking too many things too 🤷
I can’t see npm stopping with license info or readmes, to me there seem to be too many tools using that, it seems too useful.
And then the argument that it would break a million existing packages.

Fair enough!

analyze/package.json
blaine-arcjet: > So if you’d add some new file and publish, you‘d immediately find that it doesn’t work.

I'd like to catch this *before* publish. Maybe we can `npm pack` in CI and then try to use the package as packed. That should help us catch any issues.

blaine-arcjet: #4454

blaine-arcjet (approved):
I didn't do a full audit of the files (a CI process should help there) but I trust you matched up files correctly 👍

wooorm-arcjet:
> I trust you matched up files correctly 

I went through `npm pack` pretty thoroughly. You did find one mistake in `analyze-wasm` so I can’t say 100%. But, I believe it’s good.

wooorm-arcjet:
yay

Issue #3921, opened 2025-04-17
Deduplicate `errorMessage` utility across packages


blaine-arcjet:
We currently duplicate the `errorMessage` function across various packages. This is just an internal utility so we don't want it to live as an external package nor be exported from a package.

I think we could solve this by using a rollup "virtual package" (see: Astro or SvelteKit virtual packages) that provides the utilities. However, I think that TypeScript and especially the LSP hard to wrangle because the types won't be directly available.

wooorm-arcjet:
~~Can an import map be used?~~

https://github.com/wooorm/dead-or-alive/blob/057704b01dc422f584eed2bb8294bae8cab40a09/package.json#L27C4-L27C11

https://github.com/wooorm/dead-or-alive/blob/057704b01dc422f584eed2bb8294bae8cab40a09/lib/index.js#L51

(edit: ohh no, because different packages 😢)

blaine-arcjet:
That's an awesome trick that I didn't know was possible! As you stated, these are in different packages and we don't really want to introduce another package to the tree just to pluck/format error messages (see the closed, linked PR).

wooorm-arcjet:
With virtual, you mean https://github.com/rollup/plugins/tree/master/packages/virtual#readme right, right?

Why are rollup config, eslint config, sprintf utilities published, but not this?

What about a `arcjet/utilities`, with say an `/internal-use-only-error-message.js`?

blaine-arcjet:
> With virtual, you mean https://github.com/rollup/plugins/tree/master/packages/virtual#readme right, right?

Nope! Rollup has a "conventional" way to specify virtual modules, which is just creating a string prefixed with `\0`.

From https://rollupjs.org/plugin-development/#conventions:
> If your plugin uses 'virtual modules' (e.g. for helper functions), prefix the module ID with \0. This prevents other plugins from trying to process it.

We use this for our Wasm plugin at https://github.com/arcjet/arcjet-js/blob/ef3a123d372d5c2006ff120d0a02287e0d28af41/rollup-config/index.js#L63-L65

> Why are rollup config, eslint config, sprintf utilities published, but not this?

Sprintf is used by the SDK (via logger) and we want to provide it for users that want to implement their own logger to plug into the Arcjet SDK. Rollup & eslint config are used by our other projects that aren't in this repo.

I also think anything that is in devDeps needs to exist on npm otherwise the lockfile generation will fail.

[...]

PR #4453, opened 2025-06-18
refactor: externalize repeated helper, bundle

wooorm-arcjet:
Related-to: GH-3921.


wooorm-arcjet:
* this is not ready, it’s here to show one approach
* the word `causeToString` is up for debate: to me `cause`, from `error.cause`, which does not need to be an actual `Error`, seems to be a good word for this value which may not be an actual `Error`. `causeToString` seems to closely match what this function does, while allowing future improvements that may not necessarily be related to just the “message”
* Rollup now just copies over (and processes) these files over, so `/inline-helpers/index.js` becomes `some-package/inline-helpers/index.js`. Those need to go in `files` of `package.json` at least. But preferably we could imperove those names that Rollup comes up with somewhat? Or, it could *actually* bundle them *inside*?
* Having this new `index.js` somewhere else, outside of a “package”, may be better. Now turborepo wants to handle it

wooorm-arcjet:
Closing in favor of #4505.

Issue #4343, opened 2025-06-09
Cache tests are flakey


blaine-arcjet:
It seems that the tests for the cache package can be flakey, specifically in CI, because the tests might take a little too long which changes the TTL returned (such as `9` seconds instead of `10` seconds).

wooorm-arcjet:
This also reminds me that it would be great to have some source map stuff to map:

```
        TestContext.<anonymous> (file:///home/runner/work/arcjet-js/arcjet-js/cache/test/memory.test.js:85:59)
```

Into the source where that stuff is

blaine-arcjet:
Fair. The issue is related to timers, so we probably need to adopt the `node:test` mock timer stuff.

wooorm-arcjet:
this may have been solved in #4472, we’ll see.

blaine-arcjet:
Let's close it as resolved and re-open if they fail again.

PR #4472, opened 2025-06-19
test: clean cache tests

wooorm-arcjet:
* Add wrapping test cases around bare assertions
* Use separate caches per test case
* Simplify test values as there are separate cases now
* ~~Only run TypeScript test suites, not JavaScript test cases: previously both ran, meaning the tests ran twice~~ (edit: no longer! That’s a Node 24 thing, before that TS doesn’t run)

This *may* solve the timing issue from #4343. If not, it will make it more obvious.

This *refactors* but does not really *change* the tests: there’s still 100% test coverage.


wooorm-arcjet (commented):
My own code looks great to me, but I cannot approve. Just checking whether this is what trunk wants

blaine-arcjet (approved):
Makes sense! Hopefully this solves the flakey tests

Issue #32, opened 2023-11-06
Figure out how to test RPC timeout


blaine-arcjet:
The `createRouterTransport` utility provided by `@connectrpc/connect` doesn't seem to timeout correctly, so we can't test timeout logic ourselves. We should report it and spend some time to figure out why it doesn't work (possibly even submitting a fix).

Alternatively, getting to zero dependencies might cause us to remove their library.

wooorm-arcjet:
I think this was done in #4479.

PR #4479, opened 2025-06-20
test: clean protocol tests

wooorm-arcjet:
In GH-4415 and GH-4472 I noticed that in particular the protocol tests could use some cleaning up.
I had some extra time to work on a big PR due to holiday in US.

Test cases were often testing equality of big objects as a sort of “fixture” testing. This made it difficult to see *what* was different between test cases. And, if a property would be added to such objects, all tests would need changing.

After cleaning (the first 2 commits), I went through the uncovered lines and added tests cases for them.

Importantly, there are barely any changes to `protocol/*` and `index.ts`. Only because their behavior was not tested, and not allowed by the types, I removed the dead code.

***Please open the files that GH collapses, and tip: go through commits one by one***

### DIFF
--- a/protocol/client.ts
+++ b/protocol/client.ts
@@ -103,6 +103,7 @@ export function createClient(options: ClientOptions): Client {
           cookies: details.cookies,
           query: details.query,
           // TODO(#208): Re-add body
+          // TODO(@wooorm-arcjet): `#208` refers to a dependabot PR: figure out why the above TODO exists.
           // body: details.body,
           extra: details.extra,
           email: details.email,
@@ -161,6 +162,7 @@ export function createClient(options: ClientOptions): Client {
           cookies: details.cookies,
           query: details.query,
           // TODO(#208): Re-add body
+          // TODO(@wooorm-arcjet): `#208` refers to a dependabot PR: figure out why the above TODO exists.
           // body: details.body,
           extra: details.extra,
           email: details.email,
--- a/protocol/client.ts
+++ b/protocol/client.ts
@@ -105,8 +105,7 @@ export function createClient(options: ClientOptions): Client {
           // TODO(#208): Re-add body
           // body: details.body,
           extra: details.extra,
-          // TODO(@wooorm-arcjet): uncovered.
-          email: typeof details.email === "string" ? details.email : undefined,
+          email: details.email,
         },
         rules: protoRules,
       });
@@ -164,7 +163,7 @@ export function createClient(options: ClientOptions): Client {
           // TODO(#208): Re-add body
           // body: details.body,
           extra: details.extra,
-          email: typeof details.email === "string" ? details.email : undefined,
+          email: details.email,
         },
         decision: ArcjetDecisionToProtocol(decision),
         rules: rules.map(ArcjetRuleToProtocol),
--- a/protocol/client.ts
+++ b/protocol/client.ts
@@ -78,7 +78,6 @@ export function createClient(options: ClientOptions): Client {
     ): Promise<ArcjetDecision> {
       const { log } = context;
 
-      // TODO(@wooorm-arcjet): uncovered.
       let hasValidateEmail = false;
       const protoRules: Rule[] = [];
       for (const rule of rules) {
@@ -118,7 +117,6 @@ export function createClient(options: ClientOptions): Client {
         headers: { Authorization: `Bearer ${context.key}` },
         // If an email rule is configured, we double the timeout.
         // See https://github.com/arcjet/arcjet-js/issues/1697
-        // TODO(@wooorm-arcjet): uncovered.
         timeoutMs: hasValidateEmail ? timeout * 2 : timeout,
       });
 
--- a/protocol/client.ts
+++ b/protocol/client.ts
@@ -78,6 +78,7 @@ export function createClient(options: ClientOptions): Client {
     ): Promise<ArcjetDecision> {
       const { log } = context;
 
+      // TODO(@wooorm-arcjet): uncovered.
       let hasValidateEmail = false;
       const protoRules: Rule[] = [];
       for (const rule of rules) {
@@ -105,6 +106,7 @@ export function createClient(options: ClientOptions): Client {
           // TODO(#208): Re-add body
           // body: details.body,
           extra: details.extra,
+          // TODO(@wooorm-arcjet): uncovered.
           email: typeof details.email === "string" ? details.email : undefined,
         },
         rules: protoRules,
@@ -116,6 +118,7 @@ export function createClient(options: ClientOptions): Client {
         headers: { Authorization: `Bearer ${context.key}` },
         // If an email rule is configured, we double the timeout.
         // See https://github.com/arcjet/arcjet-js/issues/1697
+        // TODO(@wooorm-arcjet): uncovered.
         timeoutMs: hasValidateEmail ? timeout * 2 : timeout,
       });
 
--- a/protocol/index.ts
+++ b/protocol/index.ts
@@ -288,10 +288,12 @@ export class ArcjetBotReason extends ArcjetReason {
     this.spoofed = init.spoofed;
   }
 
+  // TODO(@wooorm-arcjet): uncovered.
   isVerified(): boolean {
     return this.verified;
   }
 
+  // TODO(@wooorm-arcjet): uncovered.
   isSpoofed(): boolean {
     return this.spoofed;
   }
@@ -309,6 +311,7 @@ export class ArcjetShieldReason extends ArcjetReason {
   constructor(init: { shieldTriggered?: boolean }) {
     super();
 
+    // TODO(@wooorm-arcjet): uncovered.
     this.shieldTriggered = init.shieldTriggered ?? false;
   }
 }
@@ -321,8 +324,10 @@ export class ArcjetEmailReason extends ArcjetReason {
   constructor(init: { emailTypes?: ArcjetEmailType[] }) {
     super();
     if (typeof init === "undefined") {
+      // TODO(@wooorm-arcjet): uncovered.
       this.emailTypes = [];
     } else {
+      // TODO(@wooorm-arcjet): uncovered.
       this.emailTypes = init.emailTypes ?? [];
     }
   }
@@ -348,6 +353,7 @@ export class ArcjetErrorReason extends ArcjetReason {
 
     // TODO: Get rid of instanceof check
     if (error instanceof Error) {
+      // TODO(@wooorm-arcjet): uncovered.
       this.message = error.message;
       return;
     }
@@ -398,6 +404,7 @@ export class ArcjetRuleResult {
   }
 
   isDenied() {
+    // TODO(@wooorm-arcjet): uncovered.
     return this.conclusion === "DENY";
   }
 }
@@ -558,10 +565,12 @@ export class ArcjetIpDetails {
   }
 
   hasLatitude(): this is RequiredProps<this, "latitude" | "accuracyRadius"> {
+    // TODO(@wooorm-arcjet): uncovered.
     return typeof this.latitude !== "undefined";
   }
 
   hasLongitude(): this is RequiredProps<this, "longitude" | "accuracyRadius"> {
+    // TODO(@wooorm-arcjet): uncovered.
     return typeof this.longitude !== "undefined";
   }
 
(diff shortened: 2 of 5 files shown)


wooorm-arcjet:
Q: The generated `BotRule`, `BotRule_Patterns` in the protocol are not called/covered. Is that because those are deprecated v1 things? If they still exist, would it make sense to somehow cover them anyway? How?

blaine-arcjet:
> Q: The generated BotRule, BotRule_Patterns in the protocol are not called/covered. Is that because those are deprecated v1 things? If they still exist, would it make sense to somehow cover them anyway? How?

Yeah, these are in the protobuf files, which is all codegen'd. Those are still in the protobuf because old SDK versions can still submit to our RPC server. I don't believe there is a way to filter them at codegen, but that would be the only place to remove them.

I don't really want them to have tests because that indicates "support" to me.

protocol/client.ts
```diff
@@ -103,9 +103,10 @@ export function createClient(options: ClientOptions): Client {
           cookies: details.cookies,
           query: details.query,
           // TODO(#208): Re-add body
+          // TODO(@wooorm-arcjet): `#208` refers to a dependabot PR: figure out why the above TODO exists.
           // body: details.body,
```
blaine-arcjet:
```suggestion
```
This is an outdated comment. We've decided that the body will never be sent to our Cloud API since it could contain things we don't want access to.

protocol/client.ts
```diff
@@ -103,9 +103,10 @@ export function createClient(options: ClientOptions): Client {
...
           extra: details.extra,
-          email: typeof details.email === "string" ? details.email : undefined,
+          email: details.email,
```
blaine-arcjet: I think we want defensive programming around this still.

protocol/client.ts
```diff
@@ -161,9 +162,10 @@ export function createClient(options: ClientOptions): Client {
           cookies: details.cookies,
           query: details.query,
           // TODO(#208): Re-add body
+          // TODO(@wooorm-arcjet): `#208` refers to a dependabot PR: figure out why the above TODO exists.
           // body: details.body,
```
blaine-arcjet:
```suggestion
```
See above

protocol/client.ts
```diff
@@ -161,9 +162,10 @@ export function createClient(options: ClientOptions): Client {
...
           extra: details.extra,
-          email: typeof details.email === "string" ? details.email : undefined,
+          email: details.email,
```
blaine-arcjet: see above

protocol/index.ts
```diff
@@ -803,6 +799,9 @@ export interface ArcjetRequestDetails {
 }
 
 export type ArcjetRule<Props extends {} = {}> = {
+  // TODO(@wooorm-arcjet):
+  // if it is intentional that people can extend rules,
+  // then we need to allow that in the types.
```
blaine-arcjet, line 804: This is allowed (see https://docs.arcjet.com/blueprints/defining-custom-rules) but is pretty advanced functionality. The types represent it so *we* have to consider in code that consumes rules.

protocol/test/convert.test.ts
```diff
@@ -57,758 +58,1288 @@ import {
...
-    );
-    assert.ok(
-      ArcjetReasonFromProtocol(
+      );
+
+      assert.ok(reason instanceof ArcjetRateLimitReason);
+      assert.equal(reason.type, "RATE_LIMIT");
+    });
+
+    await t.test("should create another rate limit reason", () => {
```
blaine-arcjet: This is testing deprecated resetTime

protocol/test/convert.test.ts
```diff
@@ -57,758 +58,1288 @@ import {
...
+        );
+      }, /Invalid Conclusion/);
+    });
   });
 
-  test("ArcjetReasonFromProtocol", () => {
-    assert.ok(ArcjetReasonFromProtocol() instanceof ArcjetReason);
-    assert.ok(
-      ArcjetReasonFromProtocol(
+  await t.test("ArcjetReasonFromProtocol", async (t) => {
```
blaine-arcjet, line 353: I'm worried that these tests are only testing instanceof. Did we remove the assertions against the output in your PR that removed `expect`? I think we need to be asserting that all the properties are copied over from the protobuf. We've missed some in the past and that leads to problems.

protocol/test/convert.test.ts
```diff
@@ -57,758 +58,1288 @@ import {
...
+      const reason = ArcjetReasonToProtocol(new ArcjetReason());
+
+      assert.ok(reason instanceof Reason);
+      assert.equal(reason.reason.case, undefined);
+    });
+
+    await t.test(
+      "should create a protocol reason from an arcjet rate limit reason",
+      () => {
+        assert.deepEqual(
```
blaine-arcjet, line 591: See comment about. These are testing the full translation, so it seems weird that we aren't testing the inverse.

protocol/test/convert.test.ts
```diff
@@ -57,758 +58,1288 @@ import {
...
         reason: new ArcjetReason(),
-      }),
-    );
+        ruleId: "rule-id",
+        state: "RUN",
+        ttl: 0,
+      });
+
+      assert.equal(result.isDenied(), false);
+    });
```
blaine-arcjet: This seems to be tested here but should exist in a separate protocol test file. This file is about ensuring the translations between protobuf<>protocol types.

protocol/test/convert.test.ts
```diff
@@ -57,758 +58,1288 @@ import {
...
+            case: "sensitiveInfo",
+            value: { allow: [], deny: [], mode: Mode.DRY_RUN },
+          },
+        }),
+      );
+      assert.equal(rule.rule.case, "sensitiveInfo");
+    });
+  });
+
+  await t.test("ArcjetIpDetails", async (t) => {
```
blaine-arcjet: This should be in a protocol test file. This file is just for converting between protobuf<>protocol

protocol/test/client.test.ts
```diff
@@ -61,1331 +50,805 @@ class TestCache implements Cache {
...
-      },
-      email: "abc@example.com",
-    };
-
-    const router = {
-      decide: mock.fn((args) => {
-        return new DecideResponse({
-          decision: {
-            conclusion: Conclusion.ALLOW,
+            // TODO(@wooorm-arcjet): is it intentional that this is an *empty* rule, not like `rule` at all?
```
blaine-arcjet: Welcome to the world of protobuf, where everything is optional and you hate your life 🙃

protocol/test/client.test.ts
```diff
@@ -61,1331 +50,805 @@ class TestCache implements Cache {
...
+      const client = createClient({
+        ...exampleClientOptions,
+        transport: createRouterTransport(({ service }) => {
+          service(DecideService, {
+            report(reportRequest) {
+              try {
+                assert.equal(calls, 0);
+                calls++;
+
+                // TODO(@wooorm-arcjet): is it intentional that this is an *empty* rule, not like `rule` at all?
```
blaine-arcjet: See above

protocol/test/client.test.ts
blaine-arcjet, line 1: GitHub UI makes this changeset miserable to review, so I've mostly just skimmed it. Comment on a couple of TODO questions

blaine-arcjet (changes requested):
Looks pretty good, but I had some feedback

blaine-arcjet:
Does this also resolve https://github.com/arcjet/arcjet-js/issues/32 ?

protocol/client.ts
wooorm-arcjet: *All* the other properties pass things through. Why is `email` different from `extra`, `query`, etc?
Should those then not also get such checks?

protocol/client.ts
wooorm-arcjet: removed the commented out `body`, and the todos

protocol/client.ts
wooorm-arcjet: removed the commented out `body`, and the todos

protocol/index.ts
wooorm-arcjet: Cool!
It is possible to let users *add* types. I think that could be used here.
Similar to how TS *adds* things in more recent JavaScript versions.

As an example: MDX expressions are an extension to markdown. In the AST, they are added with the `declare module` stuff at the bottom of `index.d.ts` here: https://www.npmjs.com/package/mdast-util-mdx-expression?activeTab=code. It adds “into” the types at `index.d.ts` of https://www.npmjs.com/package/@types/mdast?activeTab=code

protocol/test/convert.test.ts
wooorm-arcjet: added that to the test case name, I think you meant that?

protocol/test/convert.test.ts
wooorm-arcjet: No, it’s been basic `instanceof` before: https://github.com/arcjet/arcjet-js/blame/62fc077ba8417f2016cc65220dc9bba6051cf041/protocol/test/convert.test.ts#L210.
I have added `reason.type` tests in that `expect` PR to check a little more that `instanceof`.
But I agree.

protocol/test/convert.test.ts
wooorm-arcjet: moved to `protocol.test.ts`

protocol/test/convert.test.ts
wooorm-arcjet: moved to `protocol.test.ts`

protocol/test/client.test.ts
wooorm-arcjet: Still, on L266, there are all these properties? Shouldn’t some of those fields go into this rule?

wooorm-arcjet:
> Does this also resolve https://github.com/arcjet/arcjet-js/issues/32?

I think so! I think you had a comment similar to that issue in the code, and I solved it, by sort of sidestepping it by inspecting `timeoutMs()`, which returns the remaining time

protocol/client.ts
blaine-arcjet: All properties besides `email` are provided via an adapter. However, `email` has to be passed to `protect()` because it is dynamic content, such as:

```js
const decision = await aj.protect(req, { email: "foobarbaz" })
```
Since type magic is done for the 2nd argument of protect (required values depending on configured rules), we'd need to do a lot of work to guard around it in adapters. It is just easier to guard this on the string type here and send `undefined` to the server, where validation will fail if they pass a number or something.

protocol/index.ts
blaine-arcjet: I'm not against the idea, but I'd need to see it. I'd rather have simple types than exact types since many users just use JS with no type hinting and saying this can be any string helps code towards that.

protocol/test/client.test.ts
blaine-arcjet: An `ArcjetRule` and the `Rule` in the Protobuf are different. An `ArcjetRule` contains both the local and remote definitions, we use some of the properties to build the remote definition (when convert functions to protobuf). Since all fields in the Protobuf are optional, everything has a "zero" value and the constructor zeros every property—there is no reason to specify zero properties.

protocol/test/convert.test.ts
blaine-arcjet: Do you want to do that in a follow up PR since we don't do it now?

protocol/client.ts
wooorm-arcjet: I think this can be improved in the types then.
It does feel a bit smelly that there is but one such field.
Done.

protocol/client.ts
wooorm-arcjet: done

protocol/index.ts
wooorm-arcjet: I am hopeful we can have both

protocol/test/convert.test.ts
wooorm-arcjet: yeah, I made a note

protocol/test/convert.test.ts
wooorm-arcjet: made a note

protocol/test/client.test.ts
wooorm-arcjet: okay, removed the note

protocol/test/client.test.ts
wooorm-arcjet: removed the note

wooorm-arcjet:
I think that’s all the comments!

protocol/client.ts
blaine-arcjet: `email` is more of a legacy field because it existed outside of the `extra` Map in the protobuf. When I implemented the token bucket rate limit, I added the `requested` field to the `extra` Map instead of adding a new property.

It becomes a little more verbose to consume values in a protobuf Map since you need to see if the key is even available.

Issue #1504, opened 2024-09-04
Consider using the examples as integration tests


e-moran:
All of the examples are currently used to ensure that specific functionality works, and we also test that they build correctly to ensure that we haven't broken types. It shouldn't be too much work to also run and curl them in CI to automate this further.

blaine-arcjet:
We have patterns for integration tests. See:
* https://docs.arcjet.com/testing
* https://github.com/arcjet/arcjet-js/tree/main/examples/express-newman

PR #4480, opened 2025-06-20
build: add separate core, coverage tests

wooorm-arcjet:
In my recent work on tests, I noticed that several packages had no tests even though Node was spawned to test them. Slowing down the build.
I also noticed that coverage was always collected, even though it was often not used. I would like to improve coverage, by
a) failing the build if it goes below threshold,
b) making it more easily visible w/ HTML reports,
c) by actually adding more tests

Before doing that, I thought I’d take some time to clean the `scripts` fields.
This PR adds a uniform `test` script, available in every package, that makes sure that package is perfect. Whether it has linting, building, or (coverage) testing: everything goes in `test`. The one command to run to make sure something is good.
It *also* splits different tasks into different scripts. If available, `test-api` for just the tests. If available, `test-coverage` for coverage.

### DIFF
--- a/.github/workflows/reusable-test.yml
+++ b/.github/workflows/reusable-test.yml
@@ -51,8 +51,5 @@ jobs:
       - name: Install dependencies
         run: npm ci
 
-      - name: Lint
-        run: npm run lint
-
       - name: Run tests
         run: npm test


wooorm-arcjet:
Undid the changes to `examples/`.

The examples being different, sounds like a reason to move them outside of this monorepo.

I can imagine that staying close to what third parties provides, comes into conflict with a) staying secure, b) maintenance cost over the years.

> I have opinions but not enough to push back on this.

This is a good moment to voice them!

> The examples will likely change in the future

In what way?

blaine-arcjet:
> This is a good moment to voice them!

* Using `&&` is not cross platform, it breaks on Windows systems. `pre*` and `post*` work cross platform because they are npm hooks. Separating with `;` is cross platform as far as I understand. This isn't as big a deal since our contributing guidelines require/recommend devcontainers which is always Linux.

* You sorted by alphabetical but I always want to see `prepublishOnly` at the top because publishing is one of my biggest concerns. Publishing correctly takes very high priority in my brain since I have messed it up so many times.

* Linting has nothing to do with tests. I pretty much don't care about linting at all and am happy to leave it exclusively to CI. It takes up precious time I could be spent iterating locally.

* I don't see a reason to separate coverage and no coverage tests. I always want to see coverage.

> In what way?

#72 and #1504

wooorm-arcjet:
Interesting to hear your experience!

> Windows systems […] This isn't as big a deal since our contributing guidelines require/recommend devcontainers which is always Linux.

Right. And also, WSL.

> prepublishOnly […] Publishing correctly takes very high priority in my brain since I have messed it up so many times.

First, I think the better script name would be `prepack` instead. I think that’s the intention. To make sure that generated files are in tarballs.

Second, I think that an `npm test` which makes sure everything is 100% amazing helps solve that mental problem: if `npm test` works you can publish. That’s how I do it.

Third, I believe that that is indeed why people sort things in particular orders. From more important to less. The problem with that, though, is that there are differences between people and differences in time. With a sorted list, if it is that important to you right now, then you know exactly where to find it.

> Linting has nothing to do with tests. I pretty much don't care about linting at all and am happy to leave it exclusively to CI. It takes up precious time I could be spent iterating locally.

I see it another way: everything that checks code, that builds code, that’s all tests. All these tools will find problems and I don’t problems.
Very important for me is that CI runs somewhere else and I want to know if CI will fail locally.

> I don't see a reason to separate coverage and no coverage tests. I always want to see coverage.

This may become a thing when we start failing on low coverage.
You mentioned iterating locally above: that’s often when I don’t want to fail on low coverage. But before push or publish, I do.
You also mentioned precious time: collecting coverage takes time that is sometimes not needed.
And last, the instrumentation for coverage has in the past made it hard for me to debug stack traces.

blaine-arcjet (approved):
Like I previously said, I don't hold my opinions around this strongly, so we can merge this.

blaine-arcjet:
Oh, you may want to remove the explicit lint step in the CI workflow since you've bundled it into the `npm test` command now.

PR #4500, opened 2025-06-24
fix: clean core types

wooorm-arcjet:
Last week I noticed, while working on the tests, that the types could be better.
I looked at them, yesterday, and today. And am finding *a lot* of room for improvement.
There are many type parameters which are either not used, or in some cases incorrect.
There are also many “wrapper” types, which can be inlined, resulting in less code and shorter “stack” traces.

I am sharing this PR before it becomes bigger. It is already bigger than would be nice. But, there is still lot more to do. And it’s all tiny commits. So separate PRs would take a while to merge. I can do that though.

### DIFF
--- a/arcjet-next/index.ts
+++ b/arcjet-next/index.ts
@@ -146,7 +146,7 @@ export interface ArcjetNextRequest {
           [string, { name: string; value: string }]
         >;
       }
-    | Record<string, string>;
+    | Record<string, string | undefined>;
 
   clone?: () => Request;
   body?: unknown;


blaine-arcjet:
As you've now noticed by the failing tests, these generics are all used for type inference on the SDK. They provide the necessary context to augment the `protect()` function's signature.

blaine-arcjet:
As far as I'm aware, TypeScript doesn't have something like Rust's `PhantomData` type (probably because it doesn't need it).

wooorm-arcjet:
Ah, I only tested the main project, I need to look at those then too.
In many cases they can be replaced by `infer`.
Though, this is exactly why I want to do it now: I believe we can do without them.

wooorm-arcjet:
also want to add that sometimes typescript failing is a good thing: I’ll have to look into this better though

Issue #1855, opened 2024-10-03
Consider building structured errors


e-moran:
Structured errors have come up as something that could be useful to our users. 

One example of where this would be useful is if a user wants to return a 400 in case a header is missing that's needed to evaluate a rule, rather than returning a 500 which would indicate a server failure.

PR #4505, opened 2025-06-25
chore: remove unneeded use of errorMessage

wooorm-arcjet:
Closes GH-3921.

### DIFF
--- a/protocol/index.ts
+++ b/protocol/index.ts
@@ -353,6 +353,7 @@ export class ArcjetErrorReason extends ArcjetReason {
     // TODO: Get rid of instanceof check
     if (error instanceof Error) {
       this.message = error.message;
+      // @ts-ignore: this line fails if compiling with TS `lib` before `es2022`, ignore it then.
       this.cause = error.cause;
       return;
     }


wooorm-arcjet:
I saw your comment but didn’t think the idea was good. So I posted an idea for your consideration that I think is better.

As you pointed out, any value can be rejected or thrown. Not just strings and objects with a `message` property which is a string. People could reject `null` and `boolean` and `number`. Weird objects with custom `toString` or `valueOf` functions.
It is not good to use `"Unknown problem"` for all those values.

Then, proper errors and similar objects have more interesting info than a message. This `errorMessage` function drops that, but `String` doesn’t.

blaine-arcjet:
> I saw your comment but didn’t think the idea was good. So I posted an idea for your consideration that I think is better.

We cannot just cast anything to a string because this could leak sensitive info into Arcjet users' logs. This could still happen with `Error` but it is generally expected that you won't put sensitive info in the message itself and would instead attach as extra fields. In the case of a random object that could leak sensitive information, it is much better to log "Unknown Error" or something generic.

All that being said, I'm trending towards not putting the error message in any logs, but instead attaching it as a property on the structured logger. We'd need to check how those are logged in Pino and Winston to make sure things aren't leaked, but I think generally the logging handlers would deal with `error` in some user-specified way.

wooorm-arcjet:
Yes, I tried to touch on that with https://github.com/arcjet/arcjet-js/issues/3921#issuecomment-3005243329.

These log messages that include an error are just more errors with `cause`s.

The difference between an error and a log message is whether something is thrown or reported.

***

Sensitive info: I don’t see that the current printing of most strings is more secure than casting things to `String`. It sounds like something that arcjet could be good at and provide. To strip out tokens and the like.

blaine-arcjet:
@arcjet-rei said we should close anything that we don't want to land. I've clearly stated above my objections to `String()` so I'm going to close this. We should instead do what I said:
> not putting the error message in any logs, but instead attaching it as a property on the structured logger.

wooorm-arcjet:
We should land this as it’s an improvement. Structured errors/logs is what I said and tracked at https://github.com/arcjet/arcjet-js/issues/1855

Issue #2020, opened 2024-10-22
Improve performance of performance metric collection/logging


blaine-arcjet:
In #2019, we ensured the performance metric collection and logging was scoped correctly, but this introduced a closure that captures the label and start time. We should investigate the performance of this—perhaps generating symbols and putting them in a Map is better than closures.

At very least, we should no-op the performance metric functions if loglevel is not `debug`.

PR #4506, opened 2025-06-25
perf: remove latency logging except on `debug`

wooorm-arcjet:
Closes GH-2020.

Pino allows any string as a level. Here we are more strict.

### DIFF
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -176,10 +176,13 @@ class Performance {
     this.log = logger;
   }
 
-  // TODO(#2020): We should no-op this if loglevel is not `debug` to do less work
   measure(label: string) {
+    if (this.log.level !== "debug") {
+      return Function.prototype;
+    }
+
     const start = performance.now();
-    return () => {
+    return (): undefined => {
       const end = performance.now();
       const diff = end - start;
       this.log.debug("LATENCY %s: %sms", label, diff.toFixed(3));
--- a/logger/index.ts
+++ b/logger/index.ts
@@ -64,7 +64,9 @@ function getOutput(obj: unknown, msg: unknown, args: unknown[]) {
 }
 
 export class Logger {
-  #logLevel: number;
+  #logLevel: 0 | 1 | 2 | 3;
+
+  level: "debug" | "error" | "info" | "warn";
 
   constructor(opts: LoggerOptions) {
     if (typeof opts.level !== "string") {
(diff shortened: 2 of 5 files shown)


arcjet/index.ts
```diff
@@ -176,10 +176,13 @@ class Performance {
     this.log = logger;
   }
 
-  // TODO(#2020): We should no-op this if loglevel is not `debug` to do less work
   measure(label: string) {
+    if (this.log.level !== "debug") {
+      return Function.prototype;
```
blaine-arcjet, line 181: Is there a difference between this and a `function noop(){}` defined at the top of the file?

logger/index.ts
```diff
@@ -64,7 +64,9 @@ function getOutput(obj: unknown, msg: unknown, args: unknown[]) {
 }
 
 export class Logger {
-  #logLevel: number;
+  #logLevel: 0 | 1 | 2 | 3;
+
+  level: string;
```
blaine-arcjet: By exposing this string, the level could be changed during execution which would then be out-of-sync with the `#logLevel`. Why not just expose `#logLevel` instead?

blaine-arcjet (commented):
I had a couple comments, but I want to understand if this actually closes the issue linked. The no-op was "at least" but the performance logging, when enabled, could still significantly slow down an application. We want to understand and account for the overhead.

logger/index.ts
wooorm-arcjet: It has to be compatible with Pino types, they expose it. I first had `level` narrower, to just the 4 fields we use, but that failed an example which passes in pino.

arcjet/index.ts
wooorm-arcjet: Is it still defined there? as what? https://github.com/arcjet/arcjet-js/blob/main/arcjet/index.ts 👀 

No, there is no difference. Except that they are not strictly equal, and everything in javascript is mutable, so theoretically someone could mutate it?

logger/index.ts
blaine-arcjet: This makes sense. I think we might want a getter for `level` that returns the string based on the `#logLevel` so they don't fall out of sync.

wooorm-arcjet: i don’t think it can go out of sync because `#logLevel` is private and never set.

Still, sure, made it a getter.

arcjet-rei:
I see one comment on the PR, which seems resolvable, and one unanswered question from @blaine-arcjet above. @wooorm-arcjet, can we get that question answered and the PR updated and merged, or closed?

blaine-arcjet:
I'm going to close this because, as my comment said, this was the "at least" solution and should not resolve the issue without deeper investigation.

PR #4514, opened 2025-06-26
perf: replace runtime detection w/ conditions

wooorm-arcjet:
Platform-specific conditions are exposed when resolving. If that self-reporting is trusted this would do the same, but in less code.

See <https://nodejs.org/api/packages.html#community-conditions-definitions>


wooorm-arcjet:
I thought it was simpler to show this than first open an issue.

Though, for some reason, it seems the Rollup TypeScript plugin doesn’t support this? Hence the failing build. I did some research on what it might be, but couldn‘t find anything.

blaine-arcjet:
This is a neat trick but as you noticed, this requires bundlers or runtimes that support package.json imports, which not all do.

The only reason `edge-light` is even specified in this package is because Next.js has a really poor static analysis pass that flags `process.env` access even if it can't be reached via the other checks.

wooorm-arcjet:
It’s supported in most everything except for old Metro as far as I am aware, I have it in many packages.
I am surprised that Rollup is failing on this, because Rollup can handle my packages fine.
My guess is that the Rollup TS integration is broken.

blaine-arcjet:
> It’s supported in most everything except for old Metro as far as I am aware, I have it in many packages.

How often are your packages that you are mentioned here installed inside of apps? Most of your packages are tools that run in a build process inside a runtime, where Arcjet is an SDK that gets bundled itself.

wooorm-arcjet:
`react-markdown` and MDX are bundled in many things, and that’s just the 2 most visible entry points, another big one is highlighters, which are used on the arcjet website. Much of the markdown stuff is also used in WYSIWYG markdown editors such as milkdown.

wooorm-arcjet:
This is the clearest example of `imports`: https://github.com/vfile/vfile/blob/5e83917c832881569c52f1118c7beea3f12ac12f/package.json#L37 in 2m repos  https://github.com/vfile/vfile/network/dependents.
That occasionally failed, https://redirect.github.com/vfile/vfile/issues/90, https://redirect.github.com/vfile/vfile/issues/81.

The clearest example of conditions is https://github.com/wooorm/decode-named-character-reference/blob/ade22666b2e3dceee6a4dc2b73293c441292904a/package.json#L21-L29 in 800m repos (because it is more recently introduced in the graph) https://github.com/wooorm/decode-named-character-reference/network/dependents.
This example is more fragile. As DOM detection isn’t *exactly* what these conditions are for.
The fragility is mostly noticeable because “edge runtimes” choose browser https://redirect.github.com/wooorm/decode-named-character-reference/pull/6.

So it is super surprising to me that it doesn’t work here. Because I know TypeScript supports this well. And Rollup supports this well. So it must be something in between.

blaine-arcjet:
I don't trust anything to work except *just JavaScript*. We've seen users on NestJS that use some webpack clone written in Rust that only generated CommonJS (and probably didn't support package.json exports).

So unless you can enumerate the stack of each user down to the bundler and verify that everything works for everyone, then I think we should stick to runtime detection; no matter how fragile.

Issue #1076, opened 2024-07-03
Fastify adapter


blaine-arcjet:
We should create a fastify adapter.

One thing I noticed that would be cool is that fastify has an official plugin for generating swagger for your API, which we could convert to a postman collection and test the requests against an API.

davidmytton:
Logging that we've had a user request for NestJS + Fastify.

davidmytton:
We had another request for Fastify. We do support it with the Node.js adapter already, but we can create a more native experience with an official adapter.

Here's a quick example server which works fine using the Node.js adapter today:

```ts
// ESM
import Fastify from "fastify";

const fastify = Fastify({
  logger: true,
});

import arcjet, { detectBot, shield, tokenBucket } from "@arcjet/node";

const aj = arcjet({
  key: process.env.ARCJET_KEY!, // Get your site key from https://app.arcjet.com
  characteristics: ["ip.src"], // Track requests by IP
  rules: [
    // Shield protects your app from common attacks e.g. SQL injection
    shield({ mode: "LIVE" }),
    // Create a bot detection rule
    detectBot({
      mode: "LIVE", // Blocks requests. Use "DRY_RUN" to log only
      // Block all bots except search engine crawlers. See
      // https://arcjet.com/bot-list
      allow: ["CATEGORY:SEARCH_ENGINE"],
    }),
    // Create a token bucket rate limit. Other algorithms are supported.
    tokenBucket({
      mode: "LIVE",
      refillRate: 5, // Refill 5 tokens per interval
      interval: 10, // Refill every 10 seconds
      capacity: 10, // Bucket capacity of 10 tokens
    }),
  ],
});

fastify.get("/", async (request, reply) => {
  const decision = await aj.protect(request, { requested: 5 }); // Deduct 5 tokens from the bucket
  console.log("Arcjet decision", decision);

  if (decision.isDenied()) {
    if (decision.reason.isRateLimit()) {
      reply.code(429).send({ error: "Too many requests" });
    } else if (decision.reason.isBot()) {
      reply.code(403).send({ error: "No bots allowed" });
    } else {
      reply.code(403).send({ error: "Forbidden" });
    }
  }

  return { hello: "world" };
});

/**
 * Run the server!
 */
const start = async () => {
  try {
    await fastify.listen({ port: 3000 });
  } catch (err) {
    fastify.log.error(err);
    process.exit(1);
  }
};
start();

```

blaine-arcjet:
[...]

Issue #1399, opened 2024-08-23
Create NextJS pages router example for `sensitiveInfo` rule


e-moran:
We want to add an example to ensure that this continues to work correctly in to the future. We also need to check that it works correctly if body parsing is disabled, and perhaps add `readBody` if it does not.

PR #4538, opened 2025-07-01
feat: add Fastify integration

wooorm-arcjet:
This commits adds another integration.
This time for Fastify.

Closes GH-1076.

### DIFF
--- a/examples/fastify/package-lock.json
+++ b/examples/fastify/package-lock.json
@@ -1,9 +1,10 @@
 {
-  "name": "fastify-rate-limit",
+  "name": "examples-fastify",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
+      "name": "examples-fastify",
       "dependencies": {
         "@arcjet/fastify": "file:../../arcjet-fastify",
         "fastify": "^5.0.0"
@@ -18,7 +19,6 @@
       "version": "1.0.0-beta.8",
       "license": "Apache-2.0",
       "dependencies": {
-        "@arcjet/body": "1.0.0-beta.8",
         "@arcjet/env": "1.0.0-beta.8",
         "@arcjet/headers": "1.0.0-beta.8",
         "@arcjet/ip": "1.0.0-beta.8",
@@ -34,10 +34,14 @@
         "@rollup/wasm-node": "4.44.1",
         "@types/node": "18.18.0",
         "eslint": "9.29.0",
+        "fastify": "^5.4.0",
         "typescript": "5.8.3"
       },
       "engines": {
         "node": ">=18"
+      },
+      "peerDependencies": {
+        "fastify": ">=5"
       }
     },
     "node_modules/@arcjet/fastify": {
--- a/examples/fastify/package.json
+++ b/examples/fastify/package.json
@@ -7,6 +7,7 @@
     "@types/node": "^20.0.0",
     "typescript": "^5.0.0"
   },
+  "name": "examples-fastify",
   "private": true,
   "scripts": {
     "start": "node --env-file .env.local index.js"


arcjet-fastify/CHANGELOG.md
blaine-arcjet, line 1: We can delete this file. Release Please will create it with the first release.

arcjet-fastify/README.md
```diff
@@ -0,0 +1,172 @@
...
+## Example: app
+
+<!-- TODO(@wooorm-arcjet): there is no fastify example there yet. -->
+
+Try an Arcjet protected app live at [https://example.arcjet.com][arcjet-examples]
+([source code][github-arcjet-examples]).
+
+## Example: Bot detection and rate limiting
+
+This example shows a rather crude block of all automated clients and bots.
```
blaine-arcjet:
```suggestion
This example shows using a rule to block of all automated clients and bots. In real usage, we recommend allowing `CATEGORY:SEARCH_ENGINE` and other useful bots to your application.
```

arcjet-fastify/README.md
```diff
@@ -0,0 +1,172 @@
...
+// Set it as an environment variable instead of hard coding it.
+const arcjetKey = process.env.ARCJET_KEY;
+
+if (!arcjetKey) {
+  throw new Error("Cannot find `ARCJET_KEY` environment variable");
+}
+
+const arcjet = arcjetFastify({
+  // Track requests per ID of authenticated users.
+  characteristics: ["userId"],
```
blaine-arcjet: We actually need to update all the examples to move this `characteristic` to the rate limit rule. Defining characteristics globally was a hack around a global cache but the latest SDK release has a segmented cache per rule.

arcjet-fastify/README.md
```diff
@@ -0,0 +1,172 @@
...
+    .header("Content-Type", "application/json")
+    .send({ message: "Hello world" });
+});
+
+await fastify.listen({ port: 3000 });
+```
+
+## Example: Shield
+
+This example shows a _[Arcjet Shield][arcjet-shield-docs]_ which protects
```
blaine-arcjet:
```suggestion
This example shows _[Arcjet Shield][arcjet-shield-docs]_ which protects
```

arcjet-fastify/README.md
```diff
@@ -0,0 +1,172 @@
...
+Implement rate limiting, bot protection, email verification, and defense
+against common attacks.
+
+This is the [Arcjet][] SDK for [Fastify][].
+
+## Getting started
+
+<!-- TODO(@wooorm-arcjet): this does not exist yet. -->
+
+Visit the [quick start guide][arcjet-quick-start-fastify] to get started.
```
blaine-arcjet: Should we remove and add a follow up issue? The docs website requires a successful TS compile so we need to ship the SDK before it can be landed. Or we could queue them together and land soon after the publish.

arcjet-fastify/README.md
```diff
@@ -0,0 +1,172 @@
...
+```sh
+npm install @arcjet/fastify
+```
+
+## Example: app
+
+<!-- TODO(@wooorm-arcjet): there is no fastify example there yet. -->
+
+Try an Arcjet protected app live at [https://example.arcjet.com][arcjet-examples]
+([source code][github-arcjet-examples]).
```
blaine-arcjet: Should we remove and add a follow up issue? The docs website requires a successful TS compile so we need to ship the SDK before it can be landed. Or we could queue them together and land soon after the publish.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+export function createRemoteClient(
+  options?: RemoteClientOptions | null | undefined,
+) {
+  const settings = options ?? {};
+  const baseUrl = settings.baseUrl ?? baseUrlFromEnvironment(process.env);
+
+  return createClient({
+    baseUrl,
+    // @ts-expect-error: TODO(@wooorm-arcjet): register label in protocol.
+    sdkStack: "FASTIFY",
```
blaine-arcjet: This requires a bunch of cross cutting. It's been a while since I've added another stack so we should do this together and record it.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+/**
+ * Create a remote client.
+ *
+ * @param options
+ *   Configuration (optional).
+ * @returns
+ *   Client.
+ */
+// TODO(@wooorm-arcjet): this looks like an undocumented feature leaking into the public API.
+// Keep it internal.
```
blaine-arcjet: This is actually a feature so we need to document it.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+  });
+}
+
+/**
+ * Type that represents a sort-of instance of our `ArcjetFastify` clients:
+ * plain objects that we build on the fly.
+ *
+ * @typeParam Props
+ *   Accumulated properties that are needed when calling `protect`;
+ *   these come from (repeatedly) calling `withRule` with different rules.
```
blaine-arcjet: They come from the initial rules specified when you instantiate the SDK *and* when you use `withRule`.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+    sdkStack: "FASTIFY",
+    sdkVersion: "__ARCJET_SDK_VERSION__",
+    timeout: settings.timeout ?? (isDevelopment(process.env) ? 1000 : 500),
+    transport: createTransport(baseUrl),
+  });
+}
+
+/**
+ * Type that represents a sort-of instance of our `ArcjetFastify` clients:
+ * plain objects that we build on the fly.
```
blaine-arcjet: This is the SDK itself. The object return is closed over so it has access to private state without leaking that to consumers.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+// TODO(@wooorm-arcjet): this looks like an interface is shared with other clients.
+// Move upwards and extend here.
+export interface ArcjetFastify<Props> {
+  /**
+   * Decide what to do with the given Fastify request using the configured rules.
+   *
+   * @param request
+   *   Request from Fastify.
+   * @param properties
+   *   Properties that are needed for the configured rules.
```
blaine-arcjet, line 142:
```suggestion
   *   Additional properties not provided by `request` that are needed for the configured rules.
```

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+   * @returns
+   *   Promise to a decision about the request.
```
blaine-arcjet:
```suggestion
   *   Promise for a decision about the request.
```
I like to say "Promise for", like "I am promising to give you"

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+   */
+  protect(
+    request: FastifyRequest,
+    ...properties: Props extends WithoutCustomProps ? [] : [Props]
+  ): Promise<ArcjetDecision>;
+
+  /**
+   * Add more rules.
+   *
+   * @param rules
```
blaine-arcjet: `withRule` only accepts 1 rule. A rule is any array of implementations.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+): ArcjetFastify<ExtraProps<Rules> & CharacteristicProps<Characteristics>> {
+  const client = options.client ?? createRemoteClient();
+  const log = options.log
+    ? options.log
+    : new Logger({ level: logLevel(process.env) });
+  const proxies = options.proxies ? options.proxies.map(parseProxy) : undefined;
+
+  // TODO(@wooorm-arcjet): being in development doesn’t seem like something that
+  // should be warned about,
+  // but an `info` log?
```
blaine-arcjet: It's a warning in case you've misconfigured in production. Most people don't log production at info level, so this would show in logs if the environment was set incorrectly (e.g. **NOT** setting a production mode)

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+  // Also, the message is about a public IP but this check doesn’t seem to be?
```
blaine-arcjet: The check to default the value runs on each handler. This was too noisy for customers in their development environment so it runs once on startup.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+  // Why not warn below, and track whether it’s been warned for zero-or-one time?
```
blaine-arcjet: Because customers didn't want this mixed with their other logs.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+        const arcjetRequest = toArcjetRequest(
+          fastifyRequest,
+          log,
+          proxies,
+          properties || {},
+        );
+
+        return aj.protect(
+          { getBody },
+          // @ts-expect-error: TODO(@wooorm-arcjet): fix types for arcjet requests.
```
blaine-arcjet: What is the error here? This works in other adapters.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+              try {
+                expectedLength = parseInt(contentLength, 10);
+              } catch {
+                // Ignore unparsable `content-length`.
+              }
+            }
+
+            // Awaited to throw if it rejects and we'll just return undefined
+            // TODO(@wooorm-arcjet): the try/catch seems to be too big to know definitely that only redirects throw.
+            const body = await readBody(fastifyRequest.raw, {
```
blaine-arcjet: What happens if fastify has already parsed the body? In other frameworks, we needed to look at the request to see if a body was attached before trying to read it ourselves. We also don't want to consume the body and make it unavailable for the handler.

blaine-arcjet: We might want to expand the example to run Sensitive Info detection and also try to use the body in some way.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+              expectedLength,
+              // TODO(@wooorm-arcjet): this looks like a security vulnerability:
```
blaine-arcjet: This is the exact code out of express' body parser module, so I doubt it is a security vuln. They added the limit to avoid DoS attacks.

If you look at `readBody` in rejects if the size exceeds the limit and we'll return an empty body while logging an error. If there is not a body for sensitive info, it returns an error decision.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+ * @param log
+ *   Logger.
+ * @param proxies
+ *   Proxies.
+ * @param properties
+ *   Properties.
+ * @returns
+ *   Arcjet request.
+ */
+function toArcjetRequest<Properties extends PlainObject>(
```
blaine-arcjet, line 285: Why did you decide to pass a bunch of arguments here instead of just closing over the function?

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+  request: FastifyRequest,
+  log: ArcjetLogger,
+  // TODO(@wooorm-arcjet): use `Cidr` type here.
+  proxies: ReadonlyArray<ReturnType<typeof parseProxy>> | undefined,
+  properties: Properties,
+): ArcjetRequest<Properties> {
+  const requestHeaders = request.headers || {};
+  // Extract cookies from original request headers.
+  // TODO(@wooorm-arcjet): why?
```
blaine-arcjet: ArcjetHeaders removes the `cookie` header because it often contains sensitive information. We send cookies to the server as a separate field on the protobuf and handle them differently than other headers due to that potential.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+  // And, why is there dead code for arrays?
```
blaine-arcjet: Not sure what you are talking about.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,382 @@
...
+
+  if (ip === "") {
+    if (isDevelopment(process.env)) {
+      ip = "127.0.0.1";
+    } else {
+      // TODO(@wooorm-arcjet): should this warn on every request?
+      // Is once enough?
+      log.warn(
+        `Client IP address is missing. If this is a dev environment set the ARCJET_ENV env var to "development"`,
+      );
```
blaine-arcjet: In production mode, this is generally a bug and we want to be loud about it.

arcjet-fastify/package.json
```diff
@@ -0,0 +1,72 @@
...
+  "dependencies": {
+    "@arcjet/body": "1.0.0-beta.8",
+    "@arcjet/env": "1.0.0-beta.8",
+    "@arcjet/headers": "1.0.0-beta.8",
+    "@arcjet/ip": "1.0.0-beta.8",
+    "@arcjet/logger": "1.0.0-beta.8",
+    "@arcjet/protocol": "1.0.0-beta.8",
+    "@arcjet/transport": "1.0.0-beta.8",
+    "arcjet": "1.0.0-beta.8",
+    "fastify": "^5.4.0"
```
blaine-arcjet: It looks like this is only for a type, so do we want this to be a peerDep?

arcjet-fastify/CHANGELOG.md
wooorm-arcjet: done

arcjet-fastify/README.md
wooorm-arcjet: I commented out both this and the next section entirely. So that we can add it later.

/cc @qw-in I think examples on the website touch on your work hence the ping

arcjet-fastify/README.md
wooorm-arcjet: (see my above comment)

arcjet-fastify/README.md
wooorm-arcjet: Is there a place on the web where things are explained that can be linked to?
I worry about putting recommendations in such readmes where the space is limited.

arcjet-fastify/README.md
wooorm-arcjet: Good to know. Is there an issue already? Do you want to make one?

There are examples on the website, examples in `arcjet/examples`, examples here in this monorepo, examples in such readmes. It is a lot to keep up to date.

arcjet-fastify/README.md
wooorm-arcjet: thanks, fixed

arcjet-fastify/package.json
wooorm-arcjet: The `FastifyRequest` type is in the types we publish: so it indeed must be a peer or regular dependency.

Normally, I prefer to stay away from peer dependencies: different package managers at different versions treat them differently.
The whole “peer” thing makes more sense when talking about runtime code, especially bundling for the client, when you want to prevent different copies showing. Less of a thing for a type.

I looked at npm and I am kinda surprised that I don’t see people specifying `fastify` in either regular or peer deps, even though they use it https://www.npmjs.com/search?q=fastify&page=1&perPage=20 

I guess https://www.npmjs.com/package/@clerk/fastify?activeTab=code is closest to us. So I went with that.

arcjet-fastify/index.ts
wooorm-arcjet: Maybe? Not sure what you mean?

The SDK is a sort of contract on an object with two fields. That could be a class with instances at the core package. Or a class here in the integration. But here it is an abstract type, and actually a plain frozen object is built. Hence the “sort of instance”.

I worry about the word SDK because there will be different SDKs. And then there are 30 integrations per SDK.

arcjet-fastify/index.ts
wooorm-arcjet: Sure, added that.

I worry about initial rules. And later rules. Two APIs.

As is noticeable in the types, TypeScript isn’t that great at inferring arrays of things. Not to mention a matrix of things. It works with an `as const` but I worry that’s too much to ask from beginners.

An API such as:

```
arcjet(stuff)
  .withRule(rule1)
  .withRule(rule2)
  .withRule(rule3)
```

would be one API and would work better with typescript.

arcjet-fastify/index.ts
wooorm-arcjet: That seems double to me. That there are no properties from `request` in `properties` is implied because the parameter before `properties` *is* `request`.
There is also a conversion between a fastify `request` here and the `request` passed to rules: `request` from fastify as passed by the user will probably have all kinds of properties that are ignored, and passed under different names, to rules?

arcjet-fastify/index.ts
wooorm-arcjet: That seems to focus more on the promise, than on the unboxed value.
I don’t like `for` but sure done

arcjet-fastify/index.ts
wooorm-arcjet: I really don’t like that a singular word is used for an array. As mentioned above also I really don’t like arrays. This will trip up users. But sure I made it a singular.

arcjet-fastify/index.ts
wooorm-arcjet: removed.

arcjet-fastify/index.ts
wooorm-arcjet: removed.

arcjet-fastify/index.ts
wooorm-arcjet: removed.

arcjet-fastify/index.ts
wooorm-arcjet: The other adapters cast to hide the type error with a comment. I documented the error with a `@ts-expect-error` and a TODO because I believe that we need to fix these things and the first step to get that done is to surface them

arcjet-fastify/index.ts
wooorm-arcjet: I tried that and it worked.
We need tests.

arcjet-fastify/index.ts
wooorm-arcjet: OK documented that `limit` throws. There are barely any docs for `@arcjet/body`.

Remains that someone might want to check for information in more or less than 1mb?

arcjet-fastify/index.ts
wooorm-arcjet: I like small functions as high as possible that have access to few things

arcjet-fastify/index.ts
wooorm-arcjet: I removed the dead code

arcjet-fastify/index.ts
wooorm-arcjet: ok pasted your words there.

arcjet-fastify/index.ts
wooorm-arcjet: ok

arcjet-fastify/README.md
qw-in: I can work on the quick-start docs for sure. Full fastify docs will take some time so I'd suggest only syncing release with the quick start

qw-in: https://github.com/arcjet/arcjet-docs/issues/538

wooorm-arcjet: thanks!

arcjet-fastify/README.md
blaine-arcjet: You'd probably want to link to something like https://docs.arcjet.com/bot-protection/identifying-bots#bot-categories

arcjet-fastify/index.ts
blaine-arcjet: Arcjet requires properties of a request to work, such as IP, headers, etc. Those are provided via the "request" a framework provides. Additional, arbitrary properties are passed through the 2nd parameter, as defined by rules.

arcjet-fastify/index.ts
blaine-arcjet: Yeah, adapters need tests.

arcjet-fastify/README.md
blaine-arcjet: #4553 but we shouldn't create more things we need to cleanup here.

arcjet-fastify/README.md
```diff
@@ -0,0 +1,182 @@
...
+
+## License
+
+Licensed under the [Apache License, Version 2.0][apache-license].
+
+[apache-license]: http://www.apache.org/licenses/LICENSE-2.0
+[arcjet]: https://arcjet.com
+[arcjet-shield-docs]: https://docs.arcjet.com/shield/concepts
+[fastify]: https://fastify.dev/
+[github-arcjet-examples]: https://github.com/arcjet/arcjet-js-example
```
blaine-arcjet:
```suggestion
```
I think the linter is mad at this

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,376 @@
...
+import { readBody } from "@arcjet/body";
+import type { Env } from "@arcjet/env";
+import {
+  baseUrl as baseUrlFromEnvironment,
+  isDevelopment,
+  logLevel,
+  platform,
+} from "@arcjet/env";
+import ArcjetHeaders from "@arcjet/headers";
+// TODO(@wooorm-arcjet): Expose `Cidr` from `@arcjet/ip`.
```
blaine-arcjet, line 9: Why?

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,376 @@
...
+import findIp, { parseProxy } from "@arcjet/ip";
+import { Logger } from "@arcjet/logger";
+// TODO(@wooorm-arcjet): we use export maps, this export map leaks files.
+// Export maps should be clean, no file extensions.
```
blaine-arcjet: We currently only use export maps to work around framework/platform/bundler inconsistencies. Did you mean to say "When we use"?

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,376 @@
...
+  Arcjet,
+  CharacteristicProps,
+  Primitive,
+  Product,
+  ExtraProps,
+} from "arcjet";
+import arcjetCore from "arcjet";
+import type { FastifyRequest } from "fastify";
+
+// TODO(@wooorm-arcjet): using `export all` will leak things in the public API,
```
blaine-arcjet, line 29: 100% agree but it is a big lift, especially to make sure these get updated when something in the core changes. I think we'll want an issue opened for this.

arcjet-fastify/index.ts
blaine-arcjet: > That could be a class with instances at the core package. Or a class here in the integration. But here it is an abstract type, and actually a plain frozen object is built. Hence the “sort of instance”.

Classes were explored in the initial design and I chose against them because you can't close over something and stop a user from messing with it. Everything attached to a class can be changed by a user (or more importantly, an attacker) in some way, so we want to close over things like the Cache, Logger, etc.

> I worry about the word SDK because there will be different SDKs. And then there are 30 integrations per SDK.

SDK is just Software Development Kit. The core can be seen as an SDK for the adapters to consume, the adapters are an SDK for our customers to consume.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,376 @@
...
+/**
+ * Type that represents a sort-of instance of our `ArcjetFastify` clients:
+ * plain objects that we build on the fly.
+ *
+ * @typeParam Props
+ *   Accumulated properties that are needed when calling `protect`;
+ *   these come from (repeatedly) calling `withRule` with different rules
+ *   and from the rules passed to the initial `arcjetFastify` call.
+ */
+// TODO(@wooorm-arcjet): this looks like an interface is shared with other clients.
```
blaine-arcjet: This is framework specific. Most frameworks, at minimum, supply `protect()` and `withRule` but that is not strict (e.g. Nest.js prefers decorators). In addition, some runtimes provide other functions (e.g. Bun require the `handler` function to access the socket and get the IP).

arcjet-fastify/index.ts
blaine-arcjet: Makes sense! We should include that issue number from the other adapters.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,376 @@
...
+            if (typeof contentLength === "string") {
+              try {
+                expectedLength = parseInt(contentLength, 10);
+              } catch {
+                // Ignore unparsable `content-length`.
+              }
+            }
+
+            // Awaited to throw if it rejects and we'll just return undefined
+            // TODO(@wooorm-arcjet): the try/catch seems to be too big to know definitely that only redirects throw.
```
blaine-arcjet:
```suggestion
```
Is this still relevant?

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,376 @@
...
+            const body = await readBody(fastifyRequest.raw, {
+              expectedLength,
+              // This will throw `Request entity too large` at `1mb`.
+              // TODO(@wooorm-arcjet): configurable? Otherwise, some constant instead of magic number.
```
blaine-arcjet: I think this requires a long-form discussion so you should probably open an issue.

arcjet-fastify/index.ts
blaine-arcjet: Interesting. I really dislike grab-bag argument lists so I always prefer closing over things if it is not causing a performance degradation.

Doesn't matter to me though, since you're the owner of this adapter.

arcjet-fastify/index.ts
blaine-arcjet: I've just been using the examples as test harnesses that I can check. Can you push the changes you tried in the example so I can test this PR out locally?

blaine-arcjet (commented):
I think we're pretty close here! I want to try out the example locally with the sensitive info rule and body parsing, so if you can push the stuff you tried, that'd be helpful.

arcjet-fastify/index.ts
blaine-arcjet: Oh, I answered this myself when I got down to the `toArcjetRequest` code where you needed to type things that could be inferred if you closed over them.

arcjet-fastify/README.md
wooorm-arcjet: ok, reworded with that link, let me know if you have further feedback!

arcjet-fastify/README.md
wooorm-arcjet: moved it to options passed to the tokenBucket rule here!

arcjet-fastify/index.ts
wooorm-arcjet: right but, I think that’s my point, that SDK is different things to different people. Then I’d prefer to use the terms *core* and *adapters* for the different things.

Maybe it is an idea to have an `Adapter` type in core, which we then use as `satisfies Adapter` in each adapter?

arcjet-fastify/index.ts
wooorm-arcjet: But is that something that an end-user must know? What is and isn’t in `FastifyRequest`?
I feel like for a user `FastifyRequest` is an opaque value.
They don’t need to know what’s in there or not, they just provide it. And they provide some properties to us.

arcjet-fastify/index.ts
wooorm-arcjet: good point, linked #220

arcjet-fastify/README.md
qw-in: Is this something I should be tracking for docs and examples?

Edit: I've opened https://github.com/arcjet/arcjet-docs/issues/542 to avoid polluting your PR 😄

arcjet-fastify/index.ts
wooorm-arcjet: Sadly it didn’t work. I assumed the raw stream would be readable, but it isn’t.
I fixed it. Fastify supports strings and JSON by default. So I just take the string or `JSON.serialize` the value now.
Tested with:

```diff
--- a/examples/fastify-rate-limit/index.js
+++ b/examples/fastify-rate-limit/index.js
@@ -1,4 +1,4 @@
-import arcjetFastify, { fixedWindow, shield } from "@arcjet/fastify";
+import arcjetFastify, { sensitiveInfo } from "@arcjet/fastify";
 import Fastify from "fastify";
 
 // Get your Arcjet key at <https://app.arcjet.com>.
@@ -15,36 +15,20 @@ const arcjet = arcjetFastify({
   key: arcjetKey,
 
   rules: [
-    // Rate limit with a fixed window.
-    // Arcjet also supports other types (sliding window, token bucket).
-    // See <https://docs.arcjet.com/rate-limiting/reference/> for more info.
-    fixedWindow({
-      // You can add a `characteristics` field here to track clients.
-      // The default is `["ip.src"]`.
-      // See <https://docs.arcjet.com/architecture/#fingerprinting> for more info.
-
-      max: 1, // Allow a single request (for demo purposes).
-      mode: "LIVE", // Use `DRY_RUN` instead of `LIVE` to only log.
-      window: "1m", // …reset after this duration.
-    }),
-    // Protect against common attacks.
-    // See <https://docs.arcjet.com/shield/reference> for more info.
-    shield({
-      mode: "LIVE", // Use `DRY_RUN` instead of `LIVE` to only log.
-    }),
+    sensitiveInfo({allow: [], mode: "LIVE"}),
   ],
 });
 
 const fastify = Fastify({ logger: true });
 
-fastify.get("/", async function (request, reply) {
+fastify.post("/", async function (request, reply) {
   const decision = await arcjet.protect(request);
 
   if (decision.isDenied()) {
     return reply
       .status(429)
       .header("Content-Type", "application/json")
-      .send({ message: "Too many requests" });
+      .send({ message: "Deny!" });
   }
 
   return reply
```

And then:

```
$ curl --json '{"it": "example@example.com"}' http://localhost:3000
{"message":"Deny!"}
tilde in fastify-rate-limit on fastify [!]
$ curl --json '{"it": "example.com"}' http://localhost:3000
{"message":"Hello world"}
```

arcjet-fastify/index.ts
wooorm-arcjet: removed `readBody` now so this is out of date

arcjet-fastify/index.ts
wooorm-arcjet: I mostly agree with that, disliking grab-bag goo. But I see the closing-over as also grab bag but with an entire scope. Which can be fine 2 or 3 levels deep. But gets more complex the bigger things get. Here the 2 functions are pretty big and each assign a lot of variables.

What I then like to go with is a `State`, which is basically like `Options` but in `State` everything is defined/initialized/as-expected. `Options` for users, `State` after validation for internal use.

arcjet-fastify/README.md
wooorm-arcjet: which linter where?! `trunk` didn’t catch it locally for me. But indeed, fixed!

arcjet-fastify/index.ts
wooorm-arcjet: I thought it was everywhere, but `"exports"` is in 10 `package.json`s
That explains it then! Changed the note to say something along those lines!

arcjet-fastify/index.ts
wooorm-arcjet: Right yeah, I can see it being useful.

One thing I started doing that may help with responsible `*` use:

https://github.com/wooorm/trigrams/blob/367de4028e16fa0072bdfa524fa5198fcdf7865f/test.js#L6-L11

That way at least it becomes visible when something should/shouldn’t be there.

arcjet-fastify/index.ts
wooorm-arcjet: right, that makes sense. Removed the note for now.

arcjet-fastify/index.ts
wooorm-arcjet: no, due to change in `body` handling I just pushed

arcjet-fastify/index.ts
wooorm-arcjet: https://github.com/arcjet/arcjet-js/issues/4562

wooorm-arcjet:
Thanks for the valuable feedback, I think that’s everything!

arcjet-fastify/index.ts
blaine-arcjet: What is the `Adapter` type in Core? It would just be an empty object because you can't guarantee that every adapter implements `protect` or `withRule` (e.g. NestJS prefers decorators to an explicit API).

arcjet-fastify/index.ts
blaine-arcjet: I love that test. We should have that!

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,352 @@
...
+// TODO(@wooorm-arcjet): remove.
+type PlainObject = {
+  [key: string]: unknown;
+};
+
+/**
+ * Configuration for {@linkcode createRemoteClient}.
+ */
+// TODO(@wooorm-arcjet): this looks like an undocumented feature leaking into the public API.
+// Keep it internal.
```
blaine-arcjet:
```suggestion
// TODO(@wooorm-arcjet): document this undocumented feature.
```

arcjet-fastify/index.ts
blaine-arcjet:
```suggestion
// TODO(@wooorm-arcjet): document this undocumented feature.
```

arcjet-fastify/index.ts
blaine-arcjet: Meeting scheduled for tomorrow so we can get this shipped.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,352 @@
...
+    sdkStack: "FASTIFY",
+    sdkVersion: "__ARCJET_SDK_VERSION__",
+    timeout: settings.timeout ?? (isDevelopment(process.env) ? 1000 : 500),
+    transport: createTransport(baseUrl),
+  });
+}
+
+/**
+ * Type that represents a sort-of instance of our `ArcjetFastify` clients:
+ * plain objects that we build on the fly.
```
blaine-arcjet:
```suggestion
 * Type that represents an instance of the `ArcjetFastify` client.
```
Remove the internal implementation details from the docblock.

arcjet-fastify/index.ts
blaine-arcjet: I think that's a fine way to think about it 👍

arcjet-fastify/index.ts
blaine-arcjet: > would be one API and would work better with typescript.

This is the API. It's just that a "rule" is either a Primitive or Product (which are both aliased as arrays to ease the type logic). You can't specify more than 1 rule per call to `withRule()` and are expected to chain them.

blaine-arcjet (commented):
I had a few more follow up comments/suggestions but I think this is pretty close to being complete. As noted, we need to support the fastify adapter in the protobuf, so I've scheduled that for tomorrow. 

I also want to play around with the adapter in the example so I'll probably push any diff to it into a separate branch and PR into this.

arcjet-fastify/index.ts
blaine-arcjet: Reading https://fastify.dev/docs/latest/Reference/ContentTypeParser/, It looks like users can remove body parsing using:

```ts
fastify.removeAllContentTypeParsers()
```

This would cause the `sensitiveInfo` rule to error. I don't think we need to support this for the first version of this adapter (especially given that the [same problem exists](https://github.com/arcjet/arcjet-js/issues/1399) in our Next.js adapter). Let's open a follow up issue similar to the linked issue as a known limitation.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,352 @@
...
+// Keep it internal.
+export function createRemoteClient(
+  options?: RemoteClientOptions | null | undefined,
+) {
+  const settings = options ?? {};
+  const baseUrl = settings.baseUrl ?? baseUrlFromEnvironment(process.env);
+
+  return createClient({
+    baseUrl,
+    // @ts-expect-error: TODO(@wooorm-arcjet): register label in protocol.
```
blaine-arcjet:
```suggestion
```
We'll remove the expected error after we cross cut.

arcjet-fastify/index.ts
```diff
@@ -0,0 +1,352 @@
...
+          // @ts-expect-error: TODO(#220): fix types for arcjet requests.
+          arcjetRequest,
+        );
+
+        async function getBody() {
+          try {
+            if (typeof fastifyRequest.body === "string") {
+              return fastifyRequest.body;
+            } else {
+              return JSON.stringify(fastifyRequest.body);
```
blaine-arcjet, line 252: With the ability to remove the body parser in mind, we have to check that this isn't null or undefined, right?

blaine-arcjet:
> I also want to play around with the adapter in the example so I'll probably push any diff to it into a separate branch and PR into this.

#4565 contains some tweaks I'd like us to make to the example.

arcjet-fastify/index.ts
wooorm-arcjet: made a note locally to PR that when I have some time

arcjet-fastify/index.ts
wooorm-arcjet: done

arcjet-fastify/index.ts
wooorm-arcjet: sure

arcjet-fastify/index.ts
wooorm-arcjet: ok, handled it

arcjet-fastify/index.ts
blaine-arcjet: #4570 added the stack, this needs to be rebased on main and then the expect-error can be removed.

arcjet-fastify/index.ts
blaine-arcjet: #4571

arcjet-fastify/index.ts
blaine-arcjet: #4572

examples/fastify/index.js
```diff
@@ -0,0 +1,100 @@
...
+
+  rules: [
+    // Rate limit with a fixed window.
+    // Arcjet also supports other types (sliding window, token bucket).
+    // See <https://docs.arcjet.com/rate-limiting/reference/> for more info.
+    fixedWindow({
+      // You can add a `characteristics` field here to track clients.
+      // The default is `["ip.src"]`.
+      // See <https://docs.arcjet.com/architecture/#fingerprinting> for more info.
+
```
blaine-arcjet:
```suggestion
```
Let's remove this comment.

examples/fastify/package-lock.json
```diff
@@ -0,0 +1,652 @@
+{
+  "name": "fastify-rate-limit",
```
blaine-arcjet: This name is pulled from directory if no `name` field is specified so this needs to be regenerated. Since the directory is now named `fastify`, you actually want to add a name to the package.json - I think I used `examples-fastify` in my PR.

examples/fastify/package.json
```diff
@@ -0,0 +1,15 @@
...
+    "@arcjet/fastify": "file:../../arcjet-fastify",
+    "fastify": "^5.0.0"
+  },
+  "devDependencies": {
+    "@types/node": "^20.0.0",
+    "typescript": "^5.0.0"
+  },
+  "private": true,
+  "scripts": {
+    "start": "node --env-file .env.local index.js"
```
blaine-arcjet, line 13: We might want to add a build script that does a tsc compilation to check the types work. We can then add the example to the CI process.

blaine-arcjet (approved):
Besides the `expect-error` comment, I saw a couple more things that we probably want to resolve. However, I'm going to approve pending those items so @wooorm-arcjet can get this merged tomorrow.

qw-in:
Starting to write docs for this and noticing the readme has all of the links in code blocks in `<>`. Is that intentional? Doesn't match the other readmes I checked

arcjet-fastify/index.ts
wooorm-arcjet: done!

examples/fastify/index.js
wooorm-arcjet: done

examples/fastify/package-lock.json
wooorm-arcjet: added name, regenerated lock

examples/fastify/package.json
wooorm-arcjet: good idea! though the examples are currently all a bit different, often not checked, using different tsconfigs. Added a private note for me to figure something out

wooorm-arcjet:
@qw-in yep, intentional. URL detection is messy, tools do it differently. For markdown, it’s an extension to GFM on top of CommonMark. So I like being explicit and using those `<>`s around them. Though, even better would be `[text explaining](where/this/goes)` or `{@link url text}`!

Issue #4477, opened 2025-06-20
Test suites run twice (both the TS and the JS code) on Node 24


wooorm-arcjet:
Looking at https://github.com/arcjet/arcjet-js/actions/runs/15755360466/job/44409479176?pr=4472 for example, run tests -> 24 -> `@arcjet/cache:test`, you can see that the test suite is printed twice. That’s because both the TypeScript, and the JavaScript files, run.

It’s probably good to only run TypeScript. But if we do that, then the tests diverge between older and newer Node versions. So probably not yet.

TypeScript also solves (side-steps) adjacent issues, like that the generated JS, when it fails, prints misleading stacktraces.

And, it leaves the question, does the non-published-code, such as these tests, really need to be compiled? Things are probably faster if not.

Issue #1455, opened 2024-08-30
Improve inference on "detect"-style functions


blaine-arcjet:
In #1446, I improved the type inference on the `detect` function by checking if it extends `undefined`; however, this means that the type inference only works when the `detect` property does not use the object-function shorthand syntax.

For example, this works:

```ts
const [rule] = sensitiveInfo({
        mode: "LIVE",
        allow: ["custom", "CREDIT_CARD_NUMBER"],
// Notice the `:` and arrow function
        detect: (tokens: string[]) => {
          return tokens.map((token) => {
            if (token === "test@example.com") {
              return "custom";
            }
          });
        },
        contextWindowSize: 1,
});
```

But this does not work:
```ts
const [rule] = sensitiveInfo({
        mode: "LIVE",
        allow: ["custom", "CREDIT_CARD_NUMBER"],
// Notice the shorthand syntax
        detect(tokens: string[]) {
          return tokens.map((token) => {
            if (token === "test@example.com") {
              return "custom";
            }
          });
        },
        contextWindowSize: 1,
});
```

I'm not good enough with TypeScript to know how to solve this inference for both cases.

Additionally, it seems that not specifying `tokens: string[]` causes the inference to not work either. I'm not sure why when it infers the arguments correctly for me.

PR #4550, opened 2025-07-03
test: clean core tests

wooorm-arcjet:
This is similar to c67d5179.

It’s big, best to compare side-by-side.

Like that commit, this removes `assert`s that are not really related to what is being tested.
That helps when in the future we add another property somewhere, which could in some cases result in many tests failing.

The main tests were giant. 5k lines.
This commit removes a lot of repetition. Cutting it down to 3.9k lines. A follow-up PR could split things up into different files. Another follow-up PR could improve the coverage by adding more tests.

Some test cases for particular rules were in the general “SDK” group. This commit moves them into the sections for those rules.

Some test cases tested different things.
This commit splits those into different cases.

Another example is a test case that tested that 10+ rules fail by passing 12 rules.
To illustrate, I took that intent and made a test case that asserted `10` rules pass and another that `11` rules fail.

Some names of test cases were a bit confusing compared to what they tested. And labels were a bit different from each other as they were added at various points in time.
This commit reworks labels for clarity and consistency.

I also labelled `@ts-expect-error`s to explain whether they were testing runtime behavior or are up for improvement.

There were already some `TODO`s inline about improving the tests, which I solved.

I found different things that seem like bugs or could otherwise be improved and noted them inline as `TODO`s.

### DIFF
--- a/arcjet/test/arcjet.test.ts
+++ b/arcjet/test/arcjet.test.ts
@@ -1842,7 +1842,7 @@ test("shield", async (t) => {
   });
 
   await t.test("should set `mode: DRY_RUN` w/o `mode`", async () => {
-    // TODO(@wooorm-arcjet): if `{}` is allowed than so should ``? But types do not allow it?
+    // TODO(#4560): if `{}` is allowed then so could/should ``.
     const [rule] = shield({});
     assert.equal(rule.mode, "DRY_RUN");
   });
--- a/arcjet/test/arcjet.test.ts
+++ b/arcjet/test/arcjet.test.ts
@@ -2466,7 +2466,7 @@ test("SDK", async (t) => {
       const testRule: ArcjetRule<{ abc: number }> = {
         version: 0,
         mode: "LIVE",
-        type: "test",
+        type: "example",
         priority: 10000,
         validate() {},
         protect() {
@@ -2533,7 +2533,7 @@ test("SDK", async (t) => {
     const testRule: ArcjetRule<{ abc: number }> = {
       version: 0,
       mode: "LIVE",
-      type: "test",
+      type: "example",
       priority: 10000,
       validate() {},
       protect() {
@@ -2570,70 +2570,7 @@ test("SDK", async (t) => {
     }, /Log is required/);
   });
 
-  await t.test("should support local-only rules", async () => {
-    let calls = 0;
-
-    const aj = arcjet({
-      ...exampleOptions,
-      rules: [
-        [
-          {
-            version: 0,
-            mode: "LIVE",
-            type: "TEST_RULE_LOCAL_ALLOWED",
-            priority: 1,
-            validate() {
-              assert.equal(calls, 0);
-              calls++;
-            },
-            async protect() {
-              assert.equal(calls, 1);
-              calls++;
-              return new ArcjetRuleResult({
-                ruleId: "test-rule-id",
-                fingerprint: "test-fingerprint",
-                ttl: 0,
-                state: "RUN",
-                conclusion: "ALLOW",
-                reason: new ArcjetTestReason(),
-              });
-            },
-          } as const,
-        ],
-        [
-          {
-            version: 0,
-            mode: "LIVE",
-            type: "TEST_RULE_LOCAL_DENIED",
-            priority: 1,
-            validate() {
-              assert.equal(calls, 2);
-              calls++;
-            },
-            async protect() {
-              assert.equal(calls, 3);
-              calls++;
-              return new ArcjetRuleResult({
-                ruleId: "test-rule-id",
-                fingerprint: "test-fingerprint",
-                ttl: 5000,
-                state: "RUN",
-                conclusion: "DENY",
-                reason: new ArcjetTestReason(),
-              });
-            },
-          } as const,
-        ],
-      ],
-    });
-
-    const decision = await aj.protect(exampleContext, exampleDetails);
-    assert.equal(decision.conclusion, "DENY");
-    assert.equal(calls, 4);
-  });
-
-  await t.test("should support only remote rules", async () => {
-    // TODO(@wooorm-arcjet): how is this testing remote rules?
+  await t.test("should support a rule that allows", async () => {
     let calls = 0;
 
     const client = {
@@ -2658,7 +2595,7 @@ test("SDK", async (t) => {
           {
             version: 0,
             mode: "LIVE",
-            type: "TEST_RULE_REMOTE",
+            type: "example-allow",
             priority: 1,
             validate() {
               assert.equal(calls, 0);
@@ -2687,144 +2624,8 @@ test("SDK", async (t) => {
     assert.equal(calls, 3);
   });
 
-  await t.test("should create an SDK with local and remote rules", async () => {
-    // TODO(@wooorm-arcjet): how is this testing remote rules?
-    let calls = 0;
-
-    const aj = arcjet({
-      ...exampleOptions,
-      rules: [
-        [
-          {
-            version: 0,
-            mode: "LIVE",
-            type: "TEST_RULE_LOCAL_ALLOWED",
-            priority: 1,
-            validate() {
-              assert.equal(calls, 0);
-              calls++;
-            },
-            async protect() {
-              assert.equal(calls, 1);
-              calls++;
-              return new ArcjetRuleResult({
-                ruleId: "test-rule-id",
-                fingerprint: "test-fingerprint",
-                ttl: 0,
-                state: "RUN",
-                conclusion: "ALLOW",
-                reason: new ArcjetTestReason(),
-              });
-            },
-          } as const,
-          {
-            version: 0,
-            mode: "LIVE",
-            type: "TEST_RULE_LOCAL_DENIED",
-            priority: 1,
-            validate() {
-              assert.equal(calls, 2);
-              calls++;
-            },
-            async protect() {
-              assert.equal(calls, 3);
-              calls++;
-              return new ArcjetRuleResult({
-                ruleId: "test-rule-id",
-                fingerprint: "test-fingerprint",
-                ttl: 5000,
-                state: "RUN",
-                conclusion: "DENY",
-                reason: new ArcjetTestReason(),
-              });
-            },
-          } as const,
-          {
-            version: 0,
-            mode: "LIVE",
-            type: "TEST_RULE_REMOTE",
-            priority: 1,
-            validate() {
-              assert.fail();
-            },
-            protect() {
-              assert.fail();
-            },
-          } as const,
-        ],
-      ],
-    });
-
-    const decision = await aj.protect(exampleContext, exampleDetails);
-    assert.equal(decision.conclusion, "DENY");
-    assert.equal(calls, 4);
-  });
-
-  await t.test("should call rules until one yields `DENY` (1)", async () => {
-    // TODO(@wooorm-arcjet): the above things were testing that too.
-    let calls = 0;
-
-    const aj = arcjet({
-      ...exampleOptions,
-      rules: [
-        [
-          {
-            version: 0,
-            mode: "LIVE",
-            type: "TEST_RULE_LOCAL_ALLOWED",
-            priority: 1,
-            validate() {
-              assert.equal(calls, 0);
-              calls++;
-            },
-            async protect() {
-              assert.equal(calls, 1);
-              calls++;
-
-              return new ArcjetRuleResult({
-                ruleId: "test-rule-id",
-                fingerprint: "test-fingerprint",
-                ttl: 0,
-                state: "RUN",
-                conclusion: "ALLOW",
-                reason: new ArcjetTestReason(),
-              });
-            },
-          } as const,
-          {
-            version: 0,
-            mode: "LIVE",
-            type: "TEST_RULE_LOCAL_DENIED",
-            priority: 1,
-            validate() {
-              assert.equal(calls, 2);
-              calls++;
-            },
-            async protect() {
-              assert.equal(calls, 3);
-              calls++;
-
-              return new ArcjetRuleResult({
-                ruleId: "test-rule-id",
-                fingerprint: "test-fingerprint",
-                ttl: 5000,
-                state: "RUN",
-                conclusion: "DENY",
-                reason: new ArcjetTestReason(),
-              });
-            },
-          } as const,
-        ],
-      ],
-    });
-
-    const decision = await aj.protect(exampleContext, exampleDetails);
-    assert.equal(decision.conclusion, "DENY");
-    assert.equal(calls, 4);
-  });
-
   await t.test(
-    "should see a rule w/o result from `validate` as an `ALLOW`",
+    "should support a rule that allows and one that denies",
     async () => {
       let calls = 0;
 
@@ -2835,15 +2636,152 @@ test("SDK", async (t) => {
             {
               version: 0,
               mode: "LIVE",
-              type: "TEST_RULE_LOCAL_INCORRECT",
+              type: "example-allow",
               priority: 1,
               validate() {
                 assert.equal(calls, 0);
                 calls++;
               },
-              // @ts-expect-error: test runtime behavior of no return value.
               async protect() {
                 assert.equal(calls, 1);
+                calls++;
+                return new ArcjetRuleResult({
+                  ruleId: "test-rule-id",
+                  fingerprint: "test-fingerprint",
+                  ttl: 0,
+                  state: "RUN",
+                  conclusion: "ALLOW",
+                  reason: new ArcjetTestReason(),
+                });
+              },
+            } as const,
+          ],
+          [
+            {
+              version: 0,
+              mode: "LIVE",
+              type: "example-deny",
+              priority: 1,
+              validate() {
+                assert.equal(calls, 2);
+                calls++;
+              },
+              async protect() {
+                assert.equal(calls, 3);
+                calls++;
+                return new ArcjetRuleResult({
+                  ruleId: "test-rule-id",
+                  fingerprint: "test-fingerprint",
+                  ttl: 5000,
+                  state: "RUN",
+                  conclusion: "DENY",
+                  reason: new ArcjetTestReason(),
+                });
+              },
+            } as const,
+          ],
+        ],
+      });
+
+      const decision = await aj.protect(exampleContext, exampleDetails);
+      assert.equal(decision.conclusion, "DENY");
+      assert.equal(calls, 4);
+    },
+  );
+
+  await t.test(
+    "should support a rule that allows, one that denies, and one that is never called",
+    async () => {
+      let calls = 0;
+
+      const aj = arcjet({
+        ...exampleOptions,
+        rules: [
+          [
+            {
+              version: 0,
+              mode: "LIVE",
+              type: "example-allow",
+              priority: 1,
+              validate() {
+                assert.equal(calls, 0);
+                calls++;
+              },
+              async protect() {
+                assert.equal(calls, 1);
+                calls++;
+                return new ArcjetRuleResult({
+                  ruleId: "test-rule-id",
+                  fingerprint: "test-fingerprint",
+                  ttl: 0,
+                  state: "RUN",
+                  conclusion: "ALLOW",
+                  reason: new ArcjetTestReason(),
+                });
+              },
+            } as const,
+            {
+              version: 0,
+              mode: "LIVE",
+              type: "example-deny",
+              priority: 1,
+              validate() {
+                assert.equal(calls, 2);
+                calls++;
+              },
+              async protect() {
+                assert.equal(calls, 3);
+                calls++;
+                return new ArcjetRuleResult({
+                  ruleId: "test-rule-id",
+                  fingerprint: "test-fingerprint",
+                  ttl: 5000,
+                  state: "RUN",
+                  conclusion: "DENY",
+                  reason: new ArcjetTestReason(),
+                });
+              },
+            } as const,
+            {
+              version: 0,
+              mode: "LIVE",
+              type: "example-never-called",
+              priority: 1,
+              validate() {
+                assert.fail();
+              },
+              protect() {
+                assert.fail();
+              },
+            } as const,
+          ],
+        ],
+      });
+
+      const decision = await aj.protect(exampleContext, exampleDetails);
+      assert.equal(decision.conclusion, "DENY");
+      assert.equal(calls, 4);
+    },
+  );
+
+  await t.test(
+    "should see a rule w/o result from `protect` as an `ALLOW`",
+    async () => {
+      let calls = 0;
+
+      const aj = arcjet({
+        ...exampleOptions,
+        rules: [
+          [
+            {
+              version: 0,
+              mode: "LIVE",
+              type: "example-no-result",
+              priority: 1,
+              validate() {},
+              // @ts-expect-error: test runtime behavior of no return value.
+              async protect() {
+                assert.equal(calls, 0);
                 calls++;
               },
             } as const,
@@ -2852,10 +2790,8 @@ test("SDK", async (t) => {
       });
 
       const decision = await aj.protect(exampleContext, exampleDetails);
-      // TODO(@wooorm-arcjet): I don’t think this comment is correct.
-      // ALLOW because the remote rule was called and it returned ALLOW.
       assert.equal(decision.conclusion, "ALLOW");
-      assert.equal(calls, 2);
+      assert.equal(calls, 1);
     },
   );
 
@@ -2870,7 +2806,7 @@ test("SDK", async (t) => {
           {
             version: 0,
             mode: "LIVE",
-            type: "TEST_RULE_LOCAL_INCORRECT",
+            type: "example-no-validate",
             priority: 1,
             protect() {
               assert.fail();
@@ -2879,7 +2815,7 @@ test("SDK", async (t) => {
           {
             version: 0,
             mode: "LIVE",
-            type: "TEST_RULE_LOCAL_DENIED",
+            type: "example-deny",
             priority: 1,
             validate() {
               assert.equal(calls, 0);
@@ -2926,7 +2862,7 @@ test("SDK", async (t) => {
           {
             version: 0,
             mode: "LIVE",
-            type: "TEST_RULE_LOCAL_INCORRECT",
+            type: "example-no-protect",
             priority: 1,
             validate() {
               assert.equal(calls, 0);
@@ -2936,7 +2872,7 @@ test("SDK", async (t) => {
           {
             version: 0,
             mode: "LIVE",
-            type: "TEST_RULE_LOCAL_DENIED",
+            type: "example-deny",
             priority: 1,
             validate() {
               assert.equal(calls, 1);
@@ -2995,11 +2931,11 @@ test("SDK", async (t) => {
   await t.test("should allow `10` rules", async () => {
     const rules = Array.from(
       { length: 10 },
-      (): Array<ArcjetRule> => [
+      (index): Array<ArcjetRule> => [
         {
           version: 0,
           mode: "LIVE",
-          type: "TEST_RULE_MULTIPLE",
+          type: "example-" + index,
           priority: 1,
           validate() {},
           protect() {
@@ -3019,11 +2955,11 @@ test("SDK", async (t) => {
   await t.test("should conclude error on `11` rules", async () => {
     const rules = Array.from(
       { length: 11 },
-      (): Array<ArcjetRule> => [
+      (index): Array<ArcjetRule> => [
         {
           version: 0,
           mode: "LIVE",
-          type: "TEST_RULE_MULTIPLE",
+          type: "example-" + index,
           priority: 1,
           validate() {},
           protect() {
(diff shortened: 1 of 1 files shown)


wooorm-arcjet:
@blaine-arcjet The thing I don’t understand is that the tests are talking about remote and local rules. It seems like something from the past, or not really tested? Does using a label of `TEST_RULE_REMOTE` make a rule remote?

blaine-arcjet:
> The thing I don’t understand is that the tests are talking about remote and local rules. It seems like something from the past, or not really tested?

This is left over code. The difference was removed in https://github.com/arcjet/arcjet-js/pull/4204.

> Does using a label of TEST_RULE_REMOTE make a rule remote?

Nope. Just left over.

arcjet/package.json
```diff
@@ -43,7 +43,7 @@
     "build": "rollup --config rollup.config.js",
     "lint": "eslint .",
     "pretest": "npm run build",
-    "test": "node --test --experimental-test-coverage"
+    "test": "node --test --experimental-test-coverage -- test/**/*.ts"
```
blaine-arcjet:
```suggestion
    "test": "node --test --experimental-test-coverage"
```

arcjet/test/arcjet.test.ts
```diff
@@ -62,477 +67,443 @@ type Props<P extends Primitive> =
...
+  headers: new Headers(),
+  host: "example.com",
+  ip: "172.100.1.1",
+  method: "GET",
+  path: "/",
+  protocol: "http",
+  query: "",
+};
+
+test("ArcjetDecision", async (t) => {
```
blaine-arcjet:
```suggestion
test("Arcjet*Decision", async (t) => {
```
Is this meant to test all decision types?

arcjet/test/arcjet.test.ts
```diff
@@ -542,273 +513,223 @@ describe("Primitive > detectBot", () => {
...
+
+  await t.test("should use the cache", async () => {
+    let calls = 0;
+    const [rule] = detectBot({ allow: [], mode: "LIVE" });
+    const result = await rule.protect(
+      {
+        ...exampleContext,
+        cache: {
+          async get(namespace, key) {
+            calls++;
```
blaine-arcjet, line 501: Just curious: why do you prefer this pattern to mocking and asserting on the mocks?

arcjet/test/arcjet.test.ts
```diff
@@ -1401,660 +1437,462 @@ describe("Primitive > slidingWindow", () => {
...
-    const [rule] = shield({
-      mode: "LIVE",
-    });
-    assert.equal(rule.type, "SHIELD");
-    assert.equal(rule.mode, "LIVE");
-  });
-
-  test("sets mode as `DRY_RUN` if not specified", async () => {
+  await t.test("should set `mode: DRY_RUN` w/o `mode`", async () => {
+    // TODO(@wooorm-arcjet): if `{}` is allowed than so should ``? But types do not allow it?
```
blaine-arcjet: I don't have a strong opinion on this. We *could* make all rules accept no options and then default to like DRY_RUN + "allow none". Do you want to make a follow up issue so other folks can chime in?

arcjet/test/arcjet.test.ts
```diff
@@ -1401,660 +1437,462 @@ describe("Primitive > slidingWindow", () => {
...
-    };
-
-    const [rule] = shield({
-      mode: "LIVE",
-    });
-    assert.equal(rule.type, "SHIELD");
-    const result = await rule.protect(context, details);
+  await t.test("should not run locally", async () => {
+    const [rule] = shield({ mode: "LIVE" });
+    const result = await rule.protect(exampleContext, exampleDetails);
```
blaine-arcjet: This doesn't really test anything anymore because all rules "run locally" but some don't do anything, some just check the cache, etc.

arcjet/test/arcjet.test.ts
```diff
@@ -2063,740 +1901,460 @@ describe("Primitive > shield", () => {
...
-      cache: new TestCache(),
-      getBody: () => Promise.resolve(undefined),
-    };
-    const details = {
-      email: undefined,
-    };
-
-    const [rule] = sensitiveInfo({ mode: "LIVE", allow: [] });
-    assert.equal(rule.type, "SENSITIVE_INFO");
+  await t.test("should not fail when calling `validate`", () => {
```
blaine-arcjet, line 1938: I think there's an open issue or made a code comment that mentions that we wish we could fail `validate` if we don't have a body, but we defer that to `protect()` to avoid reading + passing the body.

Nothing to do now but though it was worthwhile to share the limitation behind this test.

arcjet/test/arcjet.test.ts
```diff
@@ -2063,740 +1901,460 @@ describe("Primitive > shield", () => {
...
+    const result = await rule.protect(
+      {
+        ...exampleContext,
+        async getBody() {
+          return "127.0.0.1 test@example.com 4242424242424242 +353 87 123 4567";
+        },
+      },
+      exampleDetails,
+    );
     assert.equal(result.conclusion, "DENY");
```
blaine-arcjet, line 1975: This is a bug, DRY_RUN mode should never produce a DENY decision.

arcjet/test/arcjet.test.ts
```diff
@@ -2063,740 +1901,460 @@ describe("Primitive > shield", () => {
...
+      ]);
+      assert.deepEqual(result.reason.denied, [
+        { end: 16, identifiedType: "EMAIL", start: 0 },
+      ]);
+      assert.equal(result.state, "RUN");
+    },
+  );
+
+  await t.test("should support a custom `detect` function", async () => {
+    function detect(tokens: string[]): Array<"CUSTOM" | undefined> {
```
blaine-arcjet: The return type should be inferred on this and augments the types available to `deny` config

arcjet/test/arcjet.test.ts
```diff
@@ -2063,740 +1901,460 @@ describe("Primitive > shield", () => {
...
-      protocol: "http",
-      host: "example.com",
-      path: "/",
-      headers: new Headers(),
-      cookies: "",
-      query: "",
-      extra: {},
-    };
+  await t.test(
+    "should fail id custom `detect` returns non-string",
```
blaine-arcjet:
```suggestion
    "should fail if custom `detect` returns non-string",
```

arcjet/test/arcjet.test.ts
```diff
@@ -2063,740 +1901,460 @@ describe("Primitive > shield", () => {
...
-    assert.equal(rule.type, "SENSITIVE_INFO");
-    const result = await rule.protect(context, details);
-    assert.equal(result.conclusion, "ALLOW");
-    assert.ok(result.reason instanceof ArcjetSensitiveInfoReason);
-    assert.deepEqual(result.reason.allowed, [
+      assert.ok(called);
+    },
+  );
+
+  await t.test("should error if w/o body (`undefined`)", async () => {
```
blaine-arcjet, line 2253: We should add a test for a rejected promise since the readBody can throw if body size is too big.

arcjet/test/arcjet.test.ts
```diff
@@ -2805,414 +2363,154 @@ describe("Primitive > sensitiveInfo", () => {
...
+      const ajOther = aj.withRule(rule);
+      type WithRuleTestOne = Assert<
+        SDKProps<
+          typeof ajOther,
+          { requested: number; userId: string | number | boolean }
+        >
+      >;
+
+      // TODO(@wooorm-arcjet): the types do not relate to reality,
+      // this rule has nothing to do with some field `abc`.
```
blaine-arcjet, line 2410: Additional properties are marshaled into the `extra` map in the protobuf and saved on each request to the backend, so it is fine that a custom rule request additional properties and "do nothing" locally.

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
         { requested: number; userId: string | number | boolean }
       >
     >;
 
-    const testRule = testRuleProps();
-
-    const aj3 = aj.withRule(testRule);
-    type WithRuleTestTwo = Assert<SDKProps<typeof aj3, { abc: number }>>;
+    // TODO(@wooorm-arcjet): the types do not relate to reality,
+    // this rule has nothing to do with some field `abc`.
```
blaine-arcjet, line 2477: See above

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
 
-    const aj = arcjet({
-      key: "test-key",
-      rules: [[testRuleLocalAllowed(), testRuleLocalDenied()]],
-      client,
-      log: mockLogger(),
-    });
-    assert.ok("protect" in aj);
-    assert.equal(typeof aj.protect, "function");
+  // TODO(@wooorm-arcjet): `log` sounds like something we can allow and provide by default.
```
blaine-arcjet, line 2511: Logger depends on log level -> log level comes from environment -> environment is accessed differently in every runtime/framework/platform so we defer this to the adapter to provide.

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
-    assert.equal(typeof aj.protect, "function");
+
+    const decision = await aj.protect(exampleContext, exampleDetails);
+    assert.equal(decision.conclusion, "DENY");
+    assert.equal(calls, 4);
   });
 
-  test("creates a new Arcjet SDK with both local and remote rules", () => {
+  await t.test("should support only remote rules", async () => {
+    // TODO(@wooorm-arcjet): how is this testing remote rules?
```
blaine-arcjet: It doesn't. There is no difference between local & remote rules anymore. If we encounter a rule type that is not supported in the protobuf, it gets sent as an empty rule (defaulted during protocol convert).

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
+    const decision = await aj.protect(exampleContext, exampleDetails);
+    assert.equal(decision.conclusion, "DENY");
+    assert.equal(calls, 4);
+  });
 
-    const decision = await aj.protect(context, request);
-    // ALLOW because the remote rule was called and it returned ALLOW
-    assert.equal(decision.conclusion, "ALLOW");
+  await t.test(
+    "should see a rule w/o result from `validate` as an `ALLOW`",
```
blaine-arcjet: I don't understand this test title.

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
-          ttl: 0,
-          reason: new ArcjetTestReason(),
-          results: [],
-        });
-      }),
-      report: mock.fn(),
-    };
+      const decision = await aj.protect(exampleContext, exampleDetails);
+      // TODO(@wooorm-arcjet): I don’t think this comment is correct.
+      // ALLOW because the remote rule was called and it returned ALLOW.
```
blaine-arcjet: Correct. Rules that don't DENY locally will be sent to the backend on the `Decide` RPC call and use the response for the conclusion.

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
-
-    const decision = await aj.protect(context, request);
-    // DENY because one local rule errored and the other denied
+    const decision = await aj.protect(exampleContext, exampleDetails);
+    assert.equal(calls, 2);
     assert.equal(decision.conclusion, "DENY");
-    const anonymousResult = decision.results.find((d) => d.ruleId === "");
-    assert.ok(anonymousResult);
-    assert.equal(anonymousResult.reason.type, "ERROR");
+    const result = decision.results.find((d) => d.ruleId === "");
```
blaine-arcjet, line 2804: If you remove the rule that produces DENY, you could check the top-level reason's type is `ERROR`

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
-
-    const decision = await aj.protect(context, request);
-    // DENY because one local rule errored and the other denied
+    const decision = await aj.protect(exampleContext, exampleDetails);
+    assert.equal(calls, 3);
     assert.equal(decision.conclusion, "DENY");
-    const anonymousResult = decision.results.find((d) => d.ruleId === "");
-    assert.ok(anonymousResult);
-    assert.equal(anonymousResult.reason.type, "ERROR");
+    const result = decision.results.find((d) => d.ruleId === "");
```
blaine-arcjet, line 2864: See above

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
+      const decision = await aj.protect(exampleContext);
+      assert.equal(decision.conclusion, "ERROR");
+    },
+  );
+
+  await t.test(
+    "should conclude an error if fingerprint cannot be generated (empty request)",
+    async () => {
+      const aj = arcjet(exampleOptions);
+      // TODO(@wooorm-arcjet): if this so clearly throws, then why is an empty object allowed by the types?
```
blaine-arcjet: Fingerprint generation is governed by characteristics so it cannot be prescriptive on what values are actually required in the request.

In the future, fingerprint generation won't even happen at the top-level and each rule will be in control of its own fingerprint.

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
+      const decision = await aj.protect(exampleContext, {});
+      assert.equal(decision.conclusion, "ERROR");
+    },
+  );
+
+  await t.test("should allow `10` rules", async () => {
+    const rules = Array.from(
+      { length: 10 },
```
blaine-arcjet, line 2895: This makes me so sad because I know what it is doing and JavaScript hurts my soul.

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
-        return new ArcjetAllowDecision({
-          ttl: 0,
-          reason: new ArcjetTestReason(),
-          results: [],
-        });
-      }),
-      report: mock.fn(),
-    };
+  await t.test("should call rules until one yields `DENY` (2)", async () => {
+    // TODO(@wooorm-arcjet): seems the same as “should call rules until one yields `DENY`” above.
```
blaine-arcjet: I don't know where this test was migrated from so I can't say what the difference was supposed to be before the refactor.

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
-      ip: request.ip,
-      method: request.method,
-      path: request.path,
-      protocol: request.protocol,
-      query: undefined,
+    const aj = arcjet({ ...exampleOptions, client });
+    await aj.protect(exampleContext, {
+      ...details,
+      // To do: should we drop values of `null`, `undefined`, etc?
+      // Now they result in `<unsupported value>`.
```
blaine-arcjet: Unfortunately protobuf doesn't let us drop those values without losing the key too. We want these values to show in a user's request dashboard so we need some sort of dummy value.

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
-  test("accepts plain object of `raw` headers", async () => {
-    const client = {
-      decide: mock.fn(async () => {
-        return new ArcjetAllowDecision({
+  await t.test("should call `decide` if all rules decide `ALLOW`", async () => {
+    let parameters: unknown;
+
+    const client: Client = {
+      async decide(a, b, ...rest) {
+        parameters = rest;
```
blaine-arcjet, line 2946: Similar to `calls++`, why this over mocking?

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
+        return new ArcjetErrorDecision({
           ttl: 0,
-          reason: new ArcjetTestReason(),
+          reason: new ArcjetErrorReason("reason"),
```
blaine-arcjet, line 2949: Why return an error?

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
-    });
+    assert.deepEqual(parameters, [[rule]]);
   });
 
-  test("converts extra keys with non-string values to string values", async () => {
-    const client = {
-      decide: mock.fn(async () => {
-        return new ArcjetAllowDecision({
-          ttl: 0,
+  await t.test("should not call `decide` if a rule decides `DENY", async () => {
```
blaine-arcjet: maybe something like "calls report instead of deny"?

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
+  });
 
-    const aj = arcjet({
-      key: "test-key",
-      rules: [[allowed]],
-      client,
-      log: mockLogger(),
-    });
+  await t.test(
+    "should not call `report` if all rules decide `ALLOW`",
```
blaine-arcjet: "calls decide instead of report"?

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
-        return new ArcjetErrorDecision({
-          ttl: 0,
-          reason: new ArcjetErrorReason("This decision not under test"),
-          results: [],
-        });
-      }),
-      report: mock.fn(),
-    };
+  await t.test(
+    "should detect `@vercel/request-context` and provide it to `report`",
```
blaine-arcjet: "and provide waitUntil"? It doesn't provide the entire context.

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
-      protocol: "http",
-      host: "example.com",
-      path: "/",
-      headers: new Headers([["User-Agent", "curl/8.1.2"]]),
-      "extra-test": "extra-test-value",
-    };
-    const rule = testRuleLocalDenied();
+    // TODO(@wooorm-arcjet): where would that `Unknown Rule type` be thrown?
+    // What is this really testing: the above object is an `ArcjetRule`,
+    // what would be invalid about it?
```
blaine-arcjet: The protocol only knows how to serialize a subset of types (as defined on the protobuf) but users could provide anything that satisfies the ArcjetRule contract.

arcjet/test/arcjet.test.ts
```diff
@@ -3222,1780 +2520,1368 @@ describe("SDK", () => {
...
+      assert.equal(decision2.isDenied(), false);
+      assert.equal(calls, 2);
+    },
+  );
 
-  test("does not throw if unknown rule type is passed", () => {
-    const client = {
-      decide: mock.fn(async () => {
+  await t.test("should process a remote rule", async () => {
+    // TODO(@wooorm-arcjet): what is this really testing? What is remote about it?
```
blaine-arcjet: Nothing, just leftover

blaine-arcjet (commented):
Nice to see this cleanup 👍 I commented inline; however, I mostly had to judge the tests on the merits of what was written because side-by-side diff didn't handle the massive changeset well.

arcjet/package.json
wooorm-arcjet: Ah yes, I forgot, that’s why the build fails. It speeds stuff up and provides good stack traces on Node 24, which is why I test there: https://github.com/arcjet/arcjet-js/issues/4477.
Removed!

arcjet/test/arcjet.test.ts
wooorm-arcjet: I don’t know, I can think of arguments for either way. No strong feelings, added that `*`!

arcjet/test/arcjet.test.ts
wooorm-arcjet: Boils down to less tools the better and plain.
Especially with tests, I want to really test only *my* code, and not have something else potentially interfering in there.
Mocking here is more used just to capture calls/arguments which isn’t really a problem, but mocking is sometimes also meant as passing some “fake” API where the code *should* work with some real API, say `fs` or `fetch`, and I think that that’s more problematic.

And, obvious, simple javascript over things that put themselves between. That also helps typescript do its thing.

You?

arcjet/test/arcjet.test.ts
wooorm-arcjet: made an issue and referenced it here!

arcjet/test/arcjet.test.ts
wooorm-arcjet: Makes a lot more sense that `local|remote` doesn’t exist anymore; removed those words, removed unneeded tests, and reshuffled!

arcjet/test/arcjet.test.ts
wooorm-arcjet: added a todo for https://github.com/arcjet/arcjet-js/issues/4561

arcjet/test/arcjet.test.ts
wooorm-arcjet: There’s something fishy going on where putting the function in the options doesn’t work:

```ts
    const [rule] = sensitiveInfo({
      contextWindowSize: 1,
      deny: ["CUSTOM"],
          // ^^^^^^^^ Type '"CUSTOM"' is not assignable to type 'ArcjetSensitiveInfoType'. ts(2322)
      detect: function detect(tokens: string[]): Array<"CUSTOM" | undefined> {
          if (token === "bad") {
            return "CUSTOM";
          }
        });
      },
      mode: "LIVE",
    });
```
    
And it importantly doesn’t become solved when I remove the input/output type annotations on the function.
(and it also doesn’t change between `method()`, `regular: function() {`, or `arrow: () => {`)

So it’s kind of a left over from fiddling with all that. I’d prefer to put the function on the options but can’t. Anyway: removed!

arcjet/test/arcjet.test.ts
wooorm-arcjet: fixed!

arcjet/test/arcjet.test.ts
wooorm-arcjet: Good idea! Added this in a private note for me: I hope to add more tests for more coverage in a follow-up. If I don‘t get to that quickly I can put those ideas into an issue.

arcjet/test/arcjet.test.ts
wooorm-arcjet: Really? Weird! So, that means for “remote” rules? Do you have examples of that?
Shouldn’t those rules then locally want to assert those extras?

arcjet/test/arcjet.test.ts
wooorm-arcjet: Good points.
Still, wouldn’t `{debug: console.debug, error: console.error, info: console.info, warn: console.warn}` (or something smarter) be viable?
A small bit of code that defaults to being loud about what’s going on for people trying things out / debugging.
Or do you really think that nobody will use this code directly, it’ll always be through integrations, and it’s important to always be conscious about log levels (in which case I can see that)

arcjet/test/arcjet.test.ts
wooorm-arcjet: reshuffled!

arcjet/test/arcjet.test.ts
wooorm-arcjet: I didn’t really understand the test 😅
Had to do with remote vs local, so this is now removed.

arcjet/test/arcjet.test.ts
wooorm-arcjet: :+1: removed this test as it related to local vs remote

arcjet/test/arcjet.test.ts
wooorm-arcjet: If I remove it, it becomes an “allow” (this test is about *silently ignoring* a missing function).
And, when I print the decision then, I don’t have access to “deeper” errors.

```js
xxx: ArcjetAllowDecision {
  id: 'lreq_01jzanmgvtf6bvc52yhdmctbkr',
  ttl: 0,
  results: [],
  ip: ArcjetIpDetails {
    latitude: undefined,
    longitude: undefined,
    accuracyRadius: undefined,
    timezone: undefined,
    postalCode: undefined,
    city: undefined,
    region: undefined,
    country: undefined,
    countryName: undefined,
    continent: undefined,
    continentName: undefined,
    asn: undefined,
    asnName: undefined,
    asnDomain: undefined,
    asnType: undefined,
    asnCountry: undefined,
    service: undefined
  },
  conclusion: 'ALLOW',
  reason: ArcjetTestReason { type: undefined }
}
```

So: dhould silent “errors” also be on allow decisions? 🤔

arcjet/test/arcjet.test.ts
wooorm-arcjet: thanks for the info, removed the note!

arcjet/test/arcjet.test.ts
wooorm-arcjet: really?!?!

arcjet/test/arcjet.test.ts
wooorm-arcjet: Yeah, had to do with local vs remote difference. It probably used to do something different, but since was basically the same. Removed in the local vs remote cleanup!

arcjet/test/arcjet.test.ts
wooorm-arcjet: Ahhh. Right.
Then, maybe empty string (`''`) for `null`/`undefined`?
Removed this note!

arcjet/test/arcjet.test.ts
wooorm-arcjet: Other than what I wrote above, in this case, there’s something at play:
`decide` throwing means `report` is called with an error. Which puts that error pretty deeply in *its* parameters.
And that makes `decide` a sadly difficult place to put assertions, which throw.

arcjet/test/arcjet.test.ts
wooorm-arcjet: I think the git diff here accidentally makes them appear related, but they aren’t. I think that’s what you reply to?

But: all `ArcjetErrorDecision`s have an `ArcjetErrorReason`.
`ArcjetTestReason` is used on `ArcjetAllowDecision` and `ArcjetDenyDecision`.

So, I think this is correct?

arcjet/test/arcjet.test.ts
wooorm-arcjet: Done. Went with `` should call `report` instead of `decide` if a rule decides `DENY` ``. Good?

arcjet/test/arcjet.test.ts
wooorm-arcjet: done!

arcjet/test/arcjet.test.ts
wooorm-arcjet: good point, done!

arcjet/test/arcjet.test.ts
wooorm-arcjet: Thanks!
This test was removed with the local vs remote commit

wooorm-arcjet:
I think that’s everything! Thanks for all the valuable insights and the thorough review!

arcjet/test/arcjet.test.ts
blaine-arcjet: The problem (and reason why we can't) use empty string is that protobuf uses the empty string as the defaulted value, so there is no distinction between an actual empty string and some other value. It's really messy but a limitation of the serialization framework.

arcjet/test/arcjet.test.ts
blaine-arcjet: Yeah, because the design was likely around `Array.from(array)` but due to DOM values and arguments being "array-like" instead of actually arrays, they had to just look at the `length` property. This means that you are abusing that workaround-in-the-spec to generate the array.

It feels like the hacks Lodash did back in the day and that's why it hurts my soul.

arcjet/test/arcjet.test.ts
blaine-arcjet: >  Still, wouldn’t {debug: console.debug, error: console.error, info: console.info, warn: console.warn} (or something smarter) be viable?

This is not viable because our logger matches Pino, which is object-first, message-second.

Additionally, we are very noisy in debug mode which should not be the default.

>  Or do you really think that nobody will use this code directly, it’ll always be through integrations, and it’s important to always be conscious about log levels (in which case I can see that)

Really, no one should use this. We've only shown it on our Docs website when a user encountered a problem with an unsupported framework. We don't document this package's API anywhere because it is only meant to be consumed by us in building adapters.

arcjet/test/arcjet.test.ts
blaine-arcjet: > Really? Weird! So, that means for “remote” rules? Do you have examples of that?

I don't have examples of it from rules, but you can see `extra` being populated in the [next adapter](https://github.com/arcjet/arcjet-js/blob/main/arcjet-next/index.ts#L506-L520) and then [spread on the request](https://github.com/arcjet/arcjet-js/blob/main/arcjet-next/index.ts#L523). All unknown props are [extracted and added](https://github.com/arcjet/arcjet-js/blob/main/arcjet/index.ts#L2198) to the `extra` field.

> Shouldn’t those rules then locally want to assert those extras?

Local validation only means the local protection won't run. The Request is always set to the server, whether via `Decide` or `Report`.

arcjet/test/arcjet.test.ts
blaine-arcjet: Yeah, I tried to capture the inference limitations in https://github.com/arcjet/arcjet-js/issues/1455

arcjet/test/arcjet.test.ts
blaine-arcjet: > Mocking here is more used just to capture calls/arguments which isn’t really a problem, but mocking is sometimes also meant as passing some “fake” API where the code should work with some real API, 

Yeah. Unfortunately, the Node test runner team decided that mock was enough for actual mocking and "spying" which is how I prefer to use mocks.

> And, obvious, simple javascript over things that put themselves between. That also helps typescript do its thing.

I don't like writing code to test my code, especially for tracking if something did/didn't run. I'd much rather spy and assert the tracked call count of a spy.

blaine-arcjet: A good example of this would be in your test titled `"should support a rule that allows and one that denies"`:

If for some reason a `validate()` function ran twice but a `protect()` function didn't run at all, you'd still end up with the `calls` variable being 4 so the tests would pass. If you instead spied on each of these functions, you'd assert that each was called exactly once.

I feel like writing the code around this `calls` value has inadvertently lead to the possibility of a regression. To implement this without the chance at a regression would require separate variables and get very verbose.

blaine-arcjet: I also don't believe the above is solved by `t.plan()` because some assertions may run and some may not.

arcjet/test/arcjet.test.ts
blaine-arcjet: I need to look into this more to understand what's happening. The rule results is empty which seems wrong.

arcjet/test/arcjet.test.ts
wooorm-arcjet: It is a bit frail, yeah. Though:

- no function should run twice, as they each make sure they are called at the order with an `assert` before incrementing?
- the other tests existing help here

Due to that last point, the call tracking can probably often be omitted. Something else would fail if things were running twice and out of order.
The call tracking exists more because these calls were tested before. But they’re repeated over and over.

Also, I do think there’s still much room for improvement after this PR. Some stuff can be removed, more must be added.

arcjet/test/arcjet.test.ts
wooorm-arcjet: it remains a bit hard to wrap my head around, I’d imagine the Arcjet Decision API to do complex stuff, but if that Decision API *requires* some prop `abc`, I’d imagine it to also be checked for locally in the JS code? And if it is an optional number, I’d expect the local JS code to also fail if it is a string?

arcjet/test/arcjet.test.ts
```diff
@@ -62,477 +67,443 @@ type Props<P extends Primitive> =
...
-    const [rule] = detectBot({ mode: "LIVE", allow: [] });
-    assert.equal(rule.type, "BOT");
     assert.throws(() => {
-      const _ = rule.validate(
-        context,
-        //@ts-expect-error
-        details,
-      );
-    });
+      const _ = rule.validate(exampleContext, {
```
arcjet-rei: Given that `exampleContext` and its attributes are not sealed or otherwise protected from mutation by tests (or the code under test), it feels a little risky to be using the same context across multiple tests. I'd create a function that returns a new blank context for each test run.

arcjet/test/arcjet.test.ts
arcjet-rei: It's been several years since I last thought about the design of `Array.from()`, but I believe Blaine is essentially correct here – `.from()` is designed to work around a bunch of weirdness with natives related to the web platform ("the DOM") and the fact that there is spec language that relies on things defining a length property being treated as valid sources for constructing into an array. It doesn't really bother me, mostly because the alternative is a bunch of dynamic typing nonsense involving branding and symbols, as seen elsewhere in adapting natives into newer versions of ECMAScript in a much more stringent fashion.

arcjet/test/arcjet.test.ts
arcjet-rei: I'm of two minds. In a TDD or BDD context, tests act as the contract for the code under test, and in many if not most situations they should be verifying preconditions, postconditions, and invariants for the functions / APIs being called. From that point of view spies are unnecessary, and mocks are only needed when you're trying to e.g. decouple a library from an SDK or API with external dependencies that are irrelevant to what the test is verifying.

On the other hand, if you are documenting and validating the implementation of an algorithm – especially if there are ordering or data dependencies within the scope of the executed calls that are not directly observable via the outputs of the code – then spies and mocks can be extremely helpful (and relatively terse) in capturing those dependencies. Especially if they let you express those dependencies declaratively – Blaine's example of `calls` being insufficient to say what exactly incremented the counter is a pertinent one, because the mocks and spies might catch race conditions (or failures to be idempotent) that would otherwise be invisible. Also, in some cases, using counters and verifying as you go can lead to subtle logic errors if you happen to reorder tests or reorder validation logic within an individual test case.

It would be one thing if these were heavyweight dependencies, but the bulk of the functionality we need is in the stdlib (whether it _should_ be is a discussion for a different audience), and the rest is not that huge either. I don't see removing devDependencies as having a ton of inherent value.

arcjet/test/arcjet.test.ts
wooorm-arcjet: good idea, done

wooorm-arcjet:
In response to Rei’s review in https://github.com/arcjet/arcjet-js/pull/4550#pullrequestreview-3004112162

> Is there a reason all of the tests are in a single file? I struggle a little bit when I have to work with such huge files.

No, I don’t think there is a good reason. I want to split it up after this!
I wondered about splitting it up before but saw some benefit of keeping it together first.

> calls

There were a lot of test cases, with a lot of lines each in them, and they each asserted many things repeatedly. In this PR I tried to keep the assertions.
My idea for the future though, is that we can have *more* test cases, but less assertions in them. If we check once or twice that `protect` is called before `validate`, we don’t need to check that in other places.

wooorm-arcjet:
Alright, this is stalled. Going to discard it and then maybe at some future time I will try and improve the tests in small bits. Thanks all!

PR #4551, opened 2025-07-03
types: add union of arcjet reasons

wooorm-arcjet:
This is a PR that shows, as an example, how unions can be used, to let TypeScript differentiate based concrete things, such as a `type` field.
But also, a user can now check if a particular reason has an `allowed` field: after that check, TypeScript then can know that the reason is either an `ArcjetSensitiveInfoReason` or an `ArcjetBotReason`.
This prevents the need for `is*` functions, which internally use that `type` field.
While nice, such functions could lead to problems, as they are not really checked:

```
  // This passes, TypeScript accepts functions that yield booleans, and doesn’t check whether the assertion relates to reality.
  isRateLimit(): this is ArcjetRateLimitReason {
    return true;
  }
```

### DIFF
--- a/arcjet-next/README.md
+++ b/arcjet-next/README.md
@@ -117,12 +117,12 @@ export async function GET(req: Request) {
   console.log("Arcjet decision", decision);
 
   if (decision.isDenied()) {
-    if (decision.reason.isRateLimit()) {
+    if (decision.reason.type === "RATE_LIMIT") {
       return NextResponse.json(
         { error: "Too Many Requests", reason: decision.reason },
         { status: 429 },
       );
-    } else if (decision.reason.isBot()) {
+    } else if (decision.reason.type === "BOT") {
       return NextResponse.json(
         { error: "No bots allowed", reason: decision.reason },
         { status: 403 },
@@ -189,7 +189,7 @@ and results.
 ```ts
 for (const result of decision.results) {
   // Check the rate limit rule results
-  if (result.reason.isRateLimit()) {
+  if (result.reason.type === "RATE_LIMIT") {
     if (result.isDenied()) {
       console.log("Rate limit rule returned deny conclusion", result);
     }
@@ -200,7 +200,7 @@ for (const result of decision.results) {
   }
 
   // Log the name of any denied bots
-  if (result.reason.isBot()) {
+  if (result.reason.type === "BOT") {
     if (result.isDenied()) {
       console.log("Detected bot", reason.denied);
     }
--- a/arcjet-next/index.ts
+++ b/arcjet-next/index.ts
@@ -614,7 +614,7 @@ export function createMiddleware(
 
     if (decision.isDenied()) {
       // TODO(#222): Content type negotiation using `Accept` header
-      if (decision.reason.isRateLimit()) {
+      if (decision.reason.type === "RATE_LIMIT") {
         return NextResponse.json(
           { code: 429, message: "Too Many Requests" },
           { status: 429 },
@@ -680,7 +680,7 @@ export function withArcjet<Args extends [ArcjetNextRequest, ...unknown[]], Res>(
     if (decision.isDenied()) {
       if (isNextApiResponse(response)) {
         // TODO(#222): Content type negotiation using `Accept` header
-        if (decision.reason.isRateLimit()) {
+        if (decision.reason.type === "RATE_LIMIT") {
           return response
             .status(429)
             .json({ code: 429, message: "Too Many Requests" });
@@ -689,7 +689,7 @@ export function withArcjet<Args extends [ArcjetNextRequest, ...unknown[]], Res>(
         }
       } else {
         // TODO(#222): Content type negotiation using `Accept` header
-        if (decision.reason.isRateLimit()) {
+        if (decision.reason.type === "RATE_LIMIT") {
           return NextResponse.json(
             { code: 429, message: "Too Many Requests" },
             { status: 429 },
--- a/arcjet-next/index.ts
+++ b/arcjet-next/index.ts
@@ -245,7 +245,7 @@ export interface ArcjetNext<Props extends PlainObject> {
    *   conclusion based on evaluating each of the configured rules. If you wish
    *   to accept Arcjet’s recommended action based on the configured rules then
    *   you can use this property.
-   * - `reason` (`ArcjetReason`) - An object containing more detailed
+   * - `reason` (`ArcjetReasons`) - An object containing more detailed
    *   information about the conclusion.
    * - `results` (`ArcjetRuleResult[]`) - An array of {@link ArcjetRuleResult} objects
    *   containing the results of each rule that was executed.
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -17,12 +17,12 @@ import type {
   ArcjetEmailType,
   ArcjetSensitiveInfoType,
   ArcjetRateLimitRule,
+  ArcjetReasons,
 } from "@arcjet/protocol";
 import {
   ArcjetBotReason,
   ArcjetEmailReason,
   ArcjetErrorReason,
-  ArcjetReason,
   ArcjetRuleResult,
   ArcjetSensitiveInfoReason,
   ArcjetDecision,
@@ -31,6 +31,7 @@ import {
   ArcjetShieldReason,
   ArcjetRateLimitReason,
   ArcjetConclusion,
+  ArcjetUnknownReason,
 } from "@arcjet/protocol";
 import type { Client } from "@arcjet/protocol/client.js";
 import * as analyze from "@arcjet/analyze";
@@ -659,7 +660,7 @@ export type ArcjetRequest<Props extends PlainObject> = Simplify<
 
 type CachedResult = {
   conclusion: ArcjetConclusion;
-  reason: ArcjetReason;
+  reason: ArcjetReasons;
 };
 
 function isRateLimitRule<Props extends PlainObject>(
@@ -2290,7 +2291,7 @@ export default function arcjet<
         ttl: 0,
         state: "NOT_RUN",
         conclusion: "ALLOW",
-        reason: new ArcjetReason(),
+        reason: new ArcjetUnknownReason(),
       });
 
       // Add top-level characteristics to all Rate Limit rules that don't already have
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -792,7 +792,7 @@ export function tokenBucket<
       );
 
       const [cached, ttl] = await context.cache.get(ruleId, fingerprint);
-      if (cached && cached.reason.isRateLimit()) {
+      if (cached && cached.reason.type === "RATE_LIMIT") {
         return new ArcjetRuleResult({
           ruleId,
           fingerprint,
@@ -939,7 +939,7 @@ export function fixedWindow<
       );
 
       const [cached, ttl] = await context.cache.get(ruleId, fingerprint);
-      if (cached && cached.reason.isRateLimit()) {
+      if (cached && cached.reason.type === "RATE_LIMIT") {
         return new ArcjetRuleResult({
           ruleId,
           fingerprint,
@@ -1080,7 +1080,7 @@ export function slidingWindow<
       );
 
       const [cached, ttl] = await context.cache.get(ruleId, fingerprint);
-      if (cached && cached.reason.isRateLimit()) {
+      if (cached && cached.reason.type === "RATE_LIMIT") {
         return new ArcjetRuleResult({
           ruleId,
           fingerprint,
--- a/arcjet/test/arcjet.test.ts
+++ b/arcjet/test/arcjet.test.ts
@@ -10,7 +10,7 @@ import arcjet, {
   ArcjetDenyDecision,
   ArcjetErrorDecision,
   ArcjetChallengeDecision,
-  ArcjetReason,
+  ArcjetUnknownReason,
   ArcjetErrorReason,
   ArcjetRuleResult,
   ArcjetEmailReason,
@@ -76,8 +76,6 @@ function requestAsJson(value: unknown): object {
   return { ...value, headers: Object.fromEntries(value.headers) };
 }
 
-class ArcjetTestReason extends ArcjetReason {}
-
 class TestCache {
   get = mock.fn<() => Promise<[unknown, number]>>(async () => [undefined, 0]);
   set = mock.fn();
@@ -98,7 +96,7 @@ describe("ArcjetDecision", () => {
   test("will default the `id` property if not specified", () => {
     const decision = new ArcjetAllowDecision({
       ttl: 0,
-      reason: new ArcjetTestReason(),
+      reason: new ArcjetUnknownReason(),
       results: [],
     });
     assert.match(decision.id, /^lreq_/);
@@ -108,7 +106,7 @@ describe("ArcjetDecision", () => {
     const decision = new ArcjetAllowDecision({
       id: "abc_123",
       ttl: 0,
-      reason: new ArcjetTestReason(),
+      reason: new ArcjetUnknownReason(),
       results: [],
     });
     assert.equal(decision.id, "abc_123");
@@ -150,7 +148,7 @@ describe("ArcjetDecision", () => {
   test("`isAllowed()` returns true when type is ALLOW", () => {
     const decision = new ArcjetAllowDecision({
       ttl: 0,
-      reason: new ArcjetTestReason(),
+      reason: new ArcjetUnknownReason(),
       results: [],
     });
     assert.equal(decision.isAllowed(), true);
@@ -168,7 +166,7 @@ describe("ArcjetDecision", () => {
   test("`isAllowed()` returns false when type is DENY", () => {
     const decision = new ArcjetDenyDecision({
       ttl: 0,
-      reason: new ArcjetTestReason(),
+      reason: new ArcjetUnknownReason(),
       results: [],
     });
     assert.equal(decision.isAllowed(), false);
@@ -177,7 +175,7 @@ describe("ArcjetDecision", () => {
   test("`isDenied()` returns false when type is ALLOW", () => {
     const decision = new ArcjetAllowDecision({
       ttl: 0,
-      reason: new ArcjetTestReason(),
+      reason: new ArcjetUnknownReason(),
       results: [],
     });
     assert.equal(decision.isDenied(), false);
@@ -195,7 +193,7 @@ describe("ArcjetDecision", () => {
   test("`isDenied()` returns true when type is DENY", () => {
     const decision = new ArcjetDenyDecision({
       ttl: 0,
-      reason: new ArcjetTestReason(),
+      reason: new ArcjetUnknownReason(),
       results: [],
     });
     assert.equal(decision.isDenied(), true);
@@ -204,7 +202,7 @@ describe("ArcjetDecision", () => {
   test("`isChallenged()` returns true when type is CHALLENGE", () => {
     const decision = new ArcjetChallengeDecision({
       ttl: 0,
-      reason: new ArcjetTestReason(),
+      reason: new ArcjetUnknownReason(),
       results: [],
     });
     assert.equal(decision.isChallenged(), true);
@@ -213,7 +211,7 @@ describe("ArcjetDecision", () => {
   test("`isErrored()` returns false when type is ALLOW", () => {
     const decision = new ArcjetAllowDecision({
       ttl: 0,
-      reason: new ArcjetTestReason(),
+      reason: new ArcjetUnknownReason(),
       results: [],
     });
     assert.equal(decision.isErrored(), false);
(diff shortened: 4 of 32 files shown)


examples/express-sensitive-info/index.js
```diff
@@ -27,7 +27,7 @@ app.use(express.text());
 app.post('/', async (req, res) => {
   const decision = await aj.protect(req);
 
-  if (decision.isDenied() && decision.reason.isSensitiveInfo()) {
+  if (decision.isDenied() && decision.reason.type === 'SENSITIVE_INFO') {
```
qw-in:
```suggestion
  if (decision.isDenied() && decision.reason.type === "SENSITIVE_INFO") {
```

protocol/index.ts
```diff
@@ -184,40 +184,81 @@ export const ArcjetSensitiveInfoType = Object.freeze({
...
+  ERROR: ArcjetErrorReason;
+  RATE_LIMIT: ArcjetRateLimitReason;
+  SENSITIVE_INFO: ArcjetSensitiveInfoReason;
+  SHIELD: ArcjetShieldReason;
+}
+
+/**
+ * Union of all concrete Arcjet reasons.
+ */
+export type ArcjetReasons =
```
qw-in, line 208: What about custom rules?

Minor thing, but plural here makes me expect an array:

```suggestion
export type ArcjetReason =
```

blaine-arcjet (changes requested):
We want to support this but it's *in addition* to the type narrowing functions.

Matching on strings is great for advanced programming patterns, such as a switch statement to handle all cases! 💪  

However, the type narrowing functions are the slickest way to get up-and-running, since they provide much better LSP support through autocomplete, docs, etc. They also make for examples that read like sentences.

arcjet-next/index.ts
```diff
@@ -245,7 +245,7 @@ export interface ArcjetNext<Props extends PlainObject> {
    *   conclusion based on evaluating each of the configured rules. If you wish
    *   to accept Arcjet’s recommended action based on the configured rules then
    *   you can use this property.
-   * - `reason` (`ArcjetReason`) - An object containing more detailed
+   * - `reason` (`ArcjetReasons`) - An object containing more detailed
```
arcjet-rei, line 248: What is the rationale for changing the name of the class / type? As @qw-in mentions, this leads me to expect this to be a container / list / group.

arcjet-next/README.md
```diff
@@ -117,12 +117,12 @@ export async function GET(req: Request) {
   console.log("Arcjet decision", decision);
 
   if (decision.isDenied()) {
-    if (decision.reason.isRateLimit()) {
+    if (decision.reason.type === "RATE_LIMIT") {
```
arcjet-rei: Replacing a predicate function with an attribute simplifies the interface of `ArcjetReason`, but going by this example, it mostly seems to exchange one form of polymorphism (predicates on an object of a given type) for another (structural typing based on the type of the `Reason`. These approaches seem equivalent in terms of the level of leverage they offer from a type perspective. Am I missing something?

protocol/test/convert.test.ts
```diff
@@ -352,15 +352,11 @@ test("convert", async (t) => {
   await t.test("ArcjetReasonFromProtocol", async (t) => {
     await t.test("should create an anonymous reason w/o proto", () => {
       const reason = ArcjetReasonFromProtocol();
-
-      assert.ok(reason instanceof ArcjetReason);
```
arcjet-rei, line 363: What is the value of doing an `instanceof` check in the tests? If the purpose of this change is to leverage TypeScript's type narrowing, shouldn't the test be doing that instead and then ensuring that `type` is absent / undefined or present / defined as expected? 

(Tests are allowed to do things that production code should not, but explicit branding tests like `instanceof` are a bit of an antipattern to me and it's always worth asking why `instanceof` is being used.)

arcjet/index.ts
```diff
@@ -31,6 +31,7 @@ import {
   ArcjetShieldReason,
   ArcjetRateLimitReason,
   ArcjetConclusion,
+  ArcjetUnknownReason,
```
arcjet-rei, line 34: This class / type feels misnamed. If we look at the decision-making process for Arcjet as a closed system, then it shouldn't be possible for there to be an "unknown" reason. If the analysis has not yet gotten to a point where it can point to the explicit cause for a grant or denial, the reason is more undefined than unknown. (It's also possible that I'm not really understanding how this class is meant to be used). Is this actually a use of the Nil Object pattern (i.e. signaling that there is reason because  this is a byproduct of an exception or some internal-only way of pointing out that a Reason is inapplicable for e.g. testing)?

arcjet/test/arcjet.test.ts
```diff
@@ -76,8 +76,6 @@ function requestAsJson(value: unknown): object {
   return { ...value, headers: Object.fromEntries(value.headers) };
 }
 
-class ArcjetTestReason extends ArcjetReason {}
```
arcjet-rei, line 79: This is how I would expect the type hierarchy to be used to test things, and I think that substituting the generic `ArcjetUnknownReason` for `ArcjetTestReason` makes the tests less clear.

arcjet-rei (changes requested):
So from looking through the PR, I see three changes:

1. Creating a new type, `ArcjetUnknownReason`, with a blank / missing `type` field. This seems to be exposing an internal piece of the state of the SDK, which is to say when something is in an indeterminate state within the decision process. What is the rationale for exposing this to SDK users?
2. Introducing a concrete enumeration via a union type of all of the different kind of reasons the decision process can produce. This feels useful, particularly if it allows us to create inline developer documentation of each response, but I'm not sure it supplants the `is*` methods, as they seem to solve separate problems of developer ergonomics. (In general, I don't like having more than one way to do things, but since the whole purpose of the SDK is to make Arcjet convenient for our customers to use, I think a good argument could be made for keeping both).
3. Renaming `ArcjetReason` to `ArcjetReasons`, which confuses me a little. The thing about union types and unions than more general is that they allow a single value to inhabit a constrained range of types. Even so, the basic contract of `ArcjetReason` remains the same – in the places where customers are going to use `ArcjetReason`, it's going to be be checking the type / reason of the result of a single call. What's the rationale for pluralizing the name of the type?

examples/express-sensitive-info/index.js
wooorm-arcjet: Done! Kinda weird that there are other `'` strings in this file. And linting didn’t catch it for me. So, that may be an opportunity.

protocol/index.ts
wooorm-arcjet: We can’t use a singular: the singular is *still* in use below. That singular doesn’t change.
It is plural because this is a *union* of types. So, several things.

The singular `ArcjetReason` then means the abstract idea of a thing. That’s what every thing *extends* from.

wooorm-arcjet:
Response to @blaine-arcjet’s [comment](https://github.com/arcjet/arcjet-js/pull/4551#pullrequestreview-2984850998):

> We want to support this but it's in addition to the type narrowing functions.

OK, removed the deprecations.

> Matching on strings is great for advanced programming patterns, such as a switch statement to handle all cases! 💪

> However, the type narrowing functions are the slickest way to get up-and-running, since they provide much better LSP support through autocomplete, docs, etc. They also make for examples that read like sentences.

I believe the opposite. The functions *use* strings internally. But that is hidden from users. Users need to learn an API. But that API uses names from those strings. So people have to kinda learn 2 things but use 1 thing.
I don’t understand what you mean by better autocomplete or docs from `is*` functions, as there is also autocomplete and docs based on the `type` field.

arcjet-next/index.ts
wooorm-arcjet: Because it is! `ArcjetReasons` is an enumeration of all reasons that we use. A union in TypeScript.
`ArcjetReason` also still exists. As the abstract interface that everything extends from.

arcjet-next/README.md
wooorm-arcjet: Right, the core idea for me is less function calls is better.
Though, the important part is that we `is*` functions look at the `type` field internally, and then do a `this is ArcjetErrorReason` type check on the return value.
And that is double to me: users can check the `type` field too. And TypeScript can do the type checks itself.
So also, less code is better, less API to learn/explain is better.

protocol/test/convert.test.ts
wooorm-arcjet: Importantly, there were barely tests for this, in https://github.com/arcjet/arcjet-js/pull/4516 I added more. So I’m sure there’s more to be improved here.

In other places @blaine-arcjet has also mentioned/jotted down that his preference is not using `instanceof`. Sounds like you and I are of similar opinions.

But, it indeed it narrows in the type system. Well, better put is, it just overwrites stuff:

```ts
    const value = {}
    if (value instanceof ArcjetErrorReason) { // Fine
      console.log(value) // Here `value` is `ArcjetReason`
    }
```

```ts
    const value = {}
    if (value.type === 'blob') { // Type error
      console.log(value) // `value` remains `{}`
    }
```

`instanceof` was used in the tests to type narrow. But now, asserting the `type` field narrows too.

arcjet/index.ts
wooorm-arcjet: Yeah it’s a bit awkward.
I think it’s used for both in the code.
With `type: undefined`, so I used that too, but also as a fall through after `bot`, `email`, `shield`, etc. doesn’t match (so the word “unknown”).
Preferably we could do without it. Not sure if that’s possible though!

arcjet/test/arcjet.test.ts
wooorm-arcjet: I don’t think much was being tested here: for `ArcjetErrorDecision`s, `ArcjetErrorReason`s are used.
But for `ArcjetAllowDecision` and `ArcjetDenyDecision`, the value passed is more “empty” to me. A default. Passed to satisfy the types and the code.
So maybe it’s better to start passing reasons that look a bit more like actual reasons.

wooorm-arcjet:
Response to @arcjet-rei’s [comment](https://github.com/arcjet/arcjet-js/pull/4551#pullrequestreview-2984860808):

> What is the rationale for exposing this to SDK users?

The rationale is that it sort of existed already: a reason that was nothing, not bot, edge rule, email, &c., is sometimes used.
I’d rather not have it though.

> but I'm not sure it supplants the is* methods, as they seem to solve separate problems of developer ergonomics. (In general, I don't like having more than one way to do things, but since the whole purpose of the SDK is to make Arcjet convenient for our customers to use, I think a good argument could be made for keeping both).

Right, so, I do believe in that. That the trade offs of 1 way instead of 2 ways, no 10 functions, no 10 docs entries, outweighs.
So, curious at hearing your developer ergonomics/DX ideas.
And, why `if (reason.isBot()) {` is more convenient than `if (reason.type === 'bot') {`.
One more reason: if someone does `console.log(reason)` it will print `{type: 'bot', ...}` so I think a lot of people will check for that.

> Renaming `ArcjetReason` to `ArcjetReasons`, which confuses me a little. The thing about union types and unions more generally is that they allow a single value to inhabit a constrained range of types. Even so, the basic contract of ArcjetReason remains the same – in the places where customers are going to use ArcjetReason, it's going to be be checking the type / reason of the result of a single call. What's the rationale for pluralizing the name of the type?

The rationale is that:
a) `ArcjetReason` still exists (just like `Element` and `HTMLElement` in the DOM) so something else is needed
b) plural for a union makes sense in my head: singular `Blob` for the shared abstract interface (with certain methods and a `type` field), and then plural `Blobs` for a union of all actual interfaces that are used with more different methods and fields
c) The plural is available, because I don’t think there are good reasons to do type aliases for things-in-arrays or things-in-sets or so, people can type `Array<Blob>` and `Set<Blob>` instead of `Blobs`
d) The alternative would be things like `BaseArcjetReason`, `ArcjetReasonAbstract`. Especially with suffixes, it conflicts with `type: 'abstract'`. For prefixes, maybe, but to me they’re not really better?

wooorm-arcjet:
Thanks for all the feedback everyone!

wooorm-arcjet:
I am surprised about the examples failing for NestJS and for Hono. It looks like they are reading `.ts` files? @blaine-arcjet is this related to why you stopped publish `.ts` files because tools read them? Or, do you have any other idea why only Nest and Hono would fail in this case?

examples/express-sensitive-info/index.js
wooorm-arcjet: Looks like some of the examples use different styles than the rest of the project. They stay close to the default examples from their framework, and then add Arcjet. That has some benefits of course, but also makes things like this harder to review. So perhaps something we can revisit and discuss.

blaine-arcjet:
> OK, removed the deprecations.

Thanks! We should remove the changes to the examples and READMEs. We can add more examples and docs for "Exhaustive handling of reasons".

>  I believe the opposite. The functions use strings internally. But that is hidden from users. Users need to learn an API. But that API uses names from those strings. So people have to kinda learn 2 things but use 1 thing.

Users don't read documentation, so we need to meet them inside their tools (LSP, MCP, etc). We should also think about the onboarding -> upskilling progression, where you can get started easily by following the LSP or MCP and then deepen your Arcjet integration as your product grows.

For example, I think about the onboarding as such:

```ts
// Create a client somewhere
const aj = arcjet({
  key: "...",
  rules: [
    shield(),
  ]
});

// Added to their handler or middleware
const decision = await aj.protect(req)

if (decision.isDenied()) {
  return Response("Denied", { status: 403 })
}
```

As their product grows, they decide they need rate limiting, so they update their client rules and handler logic:

```diff
// Client
const aj = arcjet({
  key: "...",
  rules: [
    shield(),
+   rateLimit({ ... }),
  ]
});

// Handler
const decision = await aj.protect(req)

if (decision.isDenied()) {
+  if (decision.reason.isRateLimit()) {
+    return Response("Rate limited", { status:  429 })
+  } else {
    return Response("Denied", { status: 403 })
+  }
}
```

And I see your improvements here as another "rung" on the upskilling ladder as they need to handle more rules and reasons:

```diff
// Client
const aj = arcjet({
  key: "...",
  rules: [
    shield(),
    rateLimit({ ... }),
+   validateEmail({ ... }),
  ]
});

// Handler
const decision = await aj.protect(req)

if (decision.isDenied()) {
-  if (decision.reason.isRateLimit()) {
+  switch (decision.reason) {
+    case "EMAIL":
+      return Response("Invalid email", { status: 400 })
+    case "RATE_LIMIT":
       return Response("Rate limited", { status:  429 })
-  } else {
+    default:
       return Response("Denied", { status: 403 })
  }
}
```

> I don’t understand what you mean by better autocomplete or docs from is* functions, as there is also autocomplete and docs based on the type field.

I think the best way to show this is via screenshots of the VSCode autocomplete.

For the individual functions, the JSDoc information is focused on the specific rule type and we can show things like how to configure the rule, the additional properties that will be available, etc.

Screenshot 2025-07-07 at 10 32 38 AM

In contrast, documenting the `type` field results in information overload since every string needs to be documented. This sample is already too big for initial user onboarding since we are trying to answer "what do they need to know *right now*?"

Screenshot 2025-07-07 at 10 41 28 AM

blaine-arcjet:
> I am surprised about the examples failing for NestJS and for Hono. It looks like they are reading .ts files? @blaine-arcjet is this related to why you stopped publish .ts files because tools read them? Or, do you have any other idea why only Nest and Hono would fail in this case?

Without digging into this deeply, my gut reaction would be that these examples don't use `strict` mode. Do these unions require TypeScript in strict mode?

blaine-arcjet:
Regarding the conversations about naming of `ArcjetReasons`:

While I think I'd prefer the singular term in a clean-slate implementation, I don't think these changes are worthy of a breaking API change. We already use `ArcjetReason` as the base class for reasons, so renaming that is a breaking change. We could use something like `ArcjetReasonType` or some such name instead of `ArcjetReasons` but naming things is one of the hardest problems in computers.

arcjet-next/index.ts
qw-in: > ArcjetReasons is an enumeration of all reasons that we use

I think this is where I'm coming from:

```typescript
// I wouldn't name this variable "reasons", so imo the type shouldn't be "Reasons"
const reason: ArcjetReasons = "SENSITIVE_INFO";
```

To satisfy the type `ArcjetReasons` a value must "pick" _one_ option from the enumerated reasons. Thinking about the types in isolation I agree can feel plural, but it only ever represents _one_ value.

Which probably gets to the crux of it. I feel types should be named for the value(s) they represent not "for themselves"

wooorm-arcjet: It is a union of interfaces, your example here is a union of their string `type` fields. After this PR the code example would be:

```ts
const type: ArcjetReasons['type'] = "SENSITIVE_INFO";
```

I can see how a singular type would in some way be preferable, but there are other reasons against the alternatives, so, tradeoffs, which I mentioned at the bottom of https://github.com/arcjet/arcjet-js/pull/4551#issuecomment-3036452490 (the alphabetic list)

wooorm-arcjet:
In response to @blaine-arcjet in https://github.com/arcjet/arcjet-js/pull/4551#issuecomment-3046046734

> Users don't read documentation, so we need to meet them inside their tools (LSP, MCP, etc). We should also think about the onboarding -> upskilling progression, where you can get started easily by following the LSP or MCP and then deepen your Arcjet integration as your product grows.

These are all points that I subscribe to, but I end up believing that a discriminating field is better at it than `is*` functions. (More examples below to illustrate that below).

> And I see your improvements here as another "rung" on the upskilling ladder as they need to handle more rules and reasons:

I don’t think I see what you mean with this example. It looks like you change from if/else to a switch, but do not prefer switches?
I personally wouldn’t use a switch either I think. I’d use the following style with either API:

```js
  if (decision.reason.isEmail()) {
    return Response("Invalid email", { status: 400 })
  }

  if (decision.reason.isRateLimit()) {
    return Response("Rate limited", { status: 429 })
  }

  return Response("Denied", { status: 403 })
```

```js
  if (decision.reason.type === 'EMAIL') {
    return Response("Invalid email", { status: 400 })
  }

  if (decision.reason.type === 'RATE_LIMIT') {
    return Response("Rate limited", { status: 429 })
  }

  return Response("Denied", { status: 403 })
```

I think the second is better because:

a) it is clear that things are exclusive; strings cannot be different, but for functions it is not know whether `isRateLimit` and `isEmail` could both be true.
b) no function calls
c) uses fields that show the actual difference instead of abstraction that uses a field internally
d) no type predicate functions (`isRateLimit(): this is ArcjetRateLimitReason { return true; }` type checks fine, it is not checked)
e) not needed to have 7 functions in code, tests, and docs
f) no collapsing with `isSpoofed`, `isVerified`, and other methods that check other things than `type`

To illuistrate some user mistakes that can be checked with fields but not `is*` functions, exclusivity:

```ts
const reason = new ArcjetBotReason({allowed: [], denied: [], verified: false, spoofed: false}) as ArcjetReasons;

if (reason.isBot()) {
  console.log("This is a bot reason.");
} else if (reason.isBot()) {
  console.log("This is also bot reason.");
}
```

```ts
const reason = new ArcjetBotReason({allowed: [], denied: [], verified: false, spoofed: false}) as ArcjetReasons;

if (reason.type === 'BOT') {
  console.log("This is a bot reason.");
} else if (reason.type === 'BOT') {
        // ^^^^^^^^^^^^^^^^^^^^^ This comparison appears to be unintentional because the types '"SHIELD" | "EDGE_RULE" | "EMAIL" | "ERROR" | "RATE_LIMIT" | "SENSITIVE_INFO" | undefined' and '"BOT"' have no overlap.
  console.log("This is a bot reason.");
}
```

And inclusivity:

```ts
const reason = new ArcjetBotReason({allowed: [], denied: [], verified: false, spoofed: false}) as ArcjetReasons;

if (reason.isBot()) {
  if (reason.isEdgeRule()) {
    console.log("Is this an edge rule bot reason?");
  }
}
```

```ts
const reason = new ArcjetBotReason({allowed: [], denied: [], verified: false, spoofed: false}) as ArcjetReasons;

if (reason.type === 'BOT') {
  if (reason.type === 'EDGE_RULE') {
   // ^^^^^^^^^^^^^^^^^^^^^^^^^^^ This comparison appears to be unintentional because the types '"BOT"' and '"EDGE_RULE"' have no overlap.
    console.log("Is this an edge rule bot reason?");
  }
}
```

I think this improved type checking will really help people.

> information overload

It’s a problem that I agree with. But it seems to me that not having 7 functions to see/learn to me seems like less information to see/learn. In your first screenshot, without `is*` functions, the user would be guided to the only field, the `type` field:

```
if (reason.|) {
        // - type (property) type: "BOT" | "EDGE_RULE" | "EMAIL" | "ERROR" | "RATE_LIMIT" | "SENSITIVE_INFO" | "SHIELD" | undefined
}
```

And there to the exclusive values:

```
if (reason.type === '|') {
                // - BOT
                // - EDGE_RULE
                // - EMAIL
                // - ERROR
                // - RATE_LIMIT
                // - SENSITIVE_INFO
                // - SHIELD
                // - undefined
}
```

Then, after a check (if no `is*` functions existed):

```
if (reason.type === 'BOT') {
  if (reason.|)
          // - allowed
          // - denied
          // - isSpoofed
          // - isVerified
          // - spoofed
          // - type
          // - verified
}
```

I do think the example you show about verbose docs is indeed a bit much. We have some but not all all the control over what (different) IDEs show. We have to find a balance.

PR #4552, opened 2025-07-03
chore: Ensure all sample code checking .isDenied() returns error response.

qw-in:
It is important that all sample code checking `decision.isDenied()` either returns an error range response or very clearly explains in a comment why it is choosing not to.

### DIFF
--- a/examples/nextjs-clerk-rate-limit/app/api/arcjet/route.ts
+++ b/examples/nextjs-clerk-rate-limit/app/api/arcjet/route.ts
@@ -68,6 +68,15 @@ export async function GET(req: NextRequest) {
         }
       );
     }
+    return NextResponse.json(
+      {
+        error: "Forbidden",
+        reason: decision.reason,
+      },
+      {
+        status: 403,
+      }
+    );
   }
   
   let reset: Date | undefined;


qw-in:
This is a follow-on from our internal discussion. I've checked every instance where code matched this regexp: `/if\s*\(\w+\.isDenied\(\)\)/`

blaine-arcjet:
@qw-in what do you think about the conditions in `examples/express-bots` and `examples/express-sensitive-info`?

qw-in:
> @qw-in what do you think about the conditions in examples/express-bots and examples/express-sensitive-info?

I think I should widen my search 😀 

I'll try to catch those. Generally I'm not sure if we should suggest checking more than just `decision.isDenied()` at the top level.

qw-in:
I've updated both examples. `examples/express-sensitive-info` was simple but `examples/express-bots` I mostly rewrote to use `@arcjet/inspect` based on our docs

wooorm-arcjet (approved):
changes looks good to me, but I am not an expert yet on what the recommended guidelines are so can’t speak to whether this is a smart idea!

qw-in:
Thanks @wooorm-arcjet! I think you or another maintainer will need to merge

wooorm-arcjet:
You couldn’t click the “To merge this pull request, check the box to the left or comment /trunk merge below.” checkbox, even after my ✅ review?
Good to know!

PR #4564, opened 2025-07-07
test: add tests for `analyze`

wooorm-arcjet:
This tests basically everything in `analyze`. Left some comments for the uncovered lines. And other things I ran into.
There’s a lot of tests for sensitive info detection and other things in the core tests too. The interface here seems mostly for internal use. So I didn’t go too far with all the different options and results.

### DIFF
--- a/analyze/test/analyze.test.ts
+++ b/analyze/test/analyze.test.ts
@@ -1,0 +1,257 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+import {
+  detectBot,
+  detectSensitiveInfo,
+  generateFingerprint,
+  isValidEmail,
+} from "../index.js";
+
+const exampleContext = { characteristics: [], log: console };
+const exampleEmailOptions = {
+  allowDomainLiteral: false,
+  allow: [],
+  requireTopLevelDomain: true,
+};
+
+test("detectBot", async function (t) {
+  await t.test("should fail w/o user agent", async function () {
+    await assert.rejects(
+      detectBot(
+        exampleContext,
+        { ip: "127.0.0.1" },
+        {
+          tag: "allowed-bot-config",
+          val: { entities: [], skipCustomDetect: false },
+        },
+      ),
+      /user-agent header is empty/,
+    );
+  });
+
+  await t.test("should allow a non-bot", async function () {
+    const result = await detectBot(
+      exampleContext,
+      { ip: "127.0.0.1", headers: { "user-agent": "Mozilla/5.0" } },
+      {
+        tag: "allowed-bot-config",
+        val: { entities: [], skipCustomDetect: false },
+      },
+    );
+
+    assert.deepEqual(result, {
+      allowed: [],
+      denied: [],
+      spoofed: false,
+      verified: false,
+    });
+  });
+
+  await t.test("should detect a bot (curl)", async function () {
+    const result = await detectBot(
+      exampleContext,
+      { ip: "127.0.0.1", headers: { "user-agent": "curl/8.1.2" } },
+      {
+        tag: "allowed-bot-config",
+        val: { entities: [], skipCustomDetect: false },
+      },
+    );
+
+    assert.deepEqual(result, {
+      allowed: [],
+      denied: ["CURL"],
+      spoofed: false,
+      verified: false,
+    });
+  });
+
+  await t.test("should detect a bot (googlebot)", async function () {
+    const result = await detectBot(
+      exampleContext,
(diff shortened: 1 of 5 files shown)


analyze/test/analyze.test.ts
```diff
@@ -0,0 +1,256 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+import {
+  detectBot,
+  detectSensitiveInfo,
+  generateFingerprint,
+  isValidEmail,
+} from "../index.js";
+
+const exampleContext = { characteristics: [], log: console };
```
blaine-arcjet, line 10: We probably want a noop logger to avoid spamming the console during tests. I find it frustrating when I can't find failures due to all the logging.

analyze/test/analyze.test.ts
```diff
@@ -0,0 +1,256 @@
...
+      allowed: [],
+      denied: ["GOOGLE_CRAWLER"],
+      spoofed: false,
+      verified: false,
+    });
+  });
+});
+
+test("detectSensitiveInfo", async function (t) {
+  await t.test("should detect sensitive info", async function () {
```
blaine-arcjet, line 88: Are you planning to add tests for all types of detections? If so, you might want to tag this as email

analyze/test/analyze.test.ts
```diff
@@ -0,0 +1,256 @@
...
+      1,
+    );
+
+    assert.deepEqual(result, {
+      allowed: [],
+      denied: [{ end: 7, identifiedType: { tag: "email" }, start: 2 }],
+    });
+  });
+
+  await t.test("should not detect regular stuff", async function () {
```
blaine-arcjet:
```suggestion
  await t.test("should not detect non-sensitive info", async function () {
```
As an uninformed reader of these tests, what is "regular stuff"?

analyze/test/analyze.test.ts
```diff
@@ -0,0 +1,256 @@
...
+    assert.deepEqual(result, {
+      allowed: [],
+      denied: [
+        { end: 5, identifiedType: { tag: "custom", val: "c" }, start: 4 },
+      ],
+    });
+  });
+});
+
+test("generateFingerprint", async function (t) {
```
blaine-arcjet, line 144: Needs characteristic tests

analyze/test/analyze.test.ts
```diff
@@ -0,0 +1,256 @@
...
+      "fp::2::30cc6b092efff7b35f658730073f40ceae0a724873e1ff175826fc57e1462149",
+    );
+  });
+
+  await t.test("should fail", async function () {
+    await assert.rejects(
+      generateFingerprint(
+        exampleContext,
+        // Note: not sure why `cookies` existing should fail but it does.
+        { cookies: "test-cookie", ip: "127.0.0.1" },
```
blaine-arcjet: Because that's not a cookie. Cookies are `key=value` values, `key=` (empty value) should also work.

analyze/test/analyze.test.ts
wooorm-arcjet: Maybe, but I also didn’t want important log messages to go ignored, sometimes they are good to see, or to be explicit about ignoring them. So, I’d rather do `log: {...console, // Ignore debug log\ndebug() {}}` or so *if* something can be ignored.

analyze/test/analyze.test.ts
wooorm-arcjet: Ah, AI autocompleted this value.
I do wish the errors would be a bit better

analyze/test/analyze.test.ts
wooorm-arcjet: done

analyze/test/analyze.test.ts
wooorm-arcjet: I wasn’t, focussing solely on the javascript/interface/input/output

analyze/test/analyze.test.ts
wooorm-arcjet: done

analyze/test/analyze.test.ts
wooorm-arcjet: changed the comment

blaine-arcjet (approved):

analyze/test/analyze.test.ts
blaine-arcjet: Yeah, structured errors from the WebAssembly is planned but it is a large task since they are not supported in Gravity.

PR #4565, opened 2025-07-07
chore: Rename fastify example, show body parsing, and link logger

blaine-arcjet:
I updated the example in #4538 to include a route that tests the body parsing. Since it was doing more than rate limiting, I renamed it. I also wanted our logs to look like Fastify logs so I showed how to link the logger between Fastify and the Arcjet client—I don't see a good way to do this automatically for Arcjet users.


wooorm-arcjet:
I don’t think trunk likes merging into non-main branches

wooorm-arcjet (changes requested):
OK I manually landed this.
Good to have this.
I did refactor the code a little bit: the POST handler was also getting rate limited, so I made that explicit. And then I just moved the sensitive info rule up to the other rules.
Also used long `--flags` in the example and showed example output.

(“requesting changes” to make sure trunk doesn’t accidentally try and merge this into something)

blaine-arcjet:
Thanks for pulling this in!

PR #4587, opened 2025-07-10
test: add tests for public interface

wooorm-arcjet:
This adds tests for what’s exposed from packages.
They make it visible when something is accidentally leaked.
Especially because of current `export *` use.
From https://github.com/arcjet/arcjet-js/pull/4538#discussion_r2185727579

In a future PR I’d like to add export maps everywhere. That prevents leaking entire files. Then these test cases can be updated to check the paths that are exposed too.

### DIFF
--- a/ip/test/index.test.ts
+++ b/ip/test/index.test.ts
@@ -1,0 +1,12 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+test("@arcjet/ip", async function (t) {
+  await t.test("should expose the public api", async function () {
+    assert.deepEqual(Object.keys(await import("../index.js")).sort(), [
+      // TODO(@wooorm-arcjet): use named exports.
+      "default",
+      "parseProxy",
+    ]);
+  });
+});
--- a/nosecone-next/test/index.test.ts
+++ b/nosecone-next/test/index.test.ts
@@ -1,0 +1,15 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+test("@nosecone/next", async function (t) {
+  await t.test("should expose the public api", async function () {
+    assert.deepEqual(Object.keys(await import("../index.js")).sort(), [
+      "createMiddleware",
+      // TODO(@wooorm-arcjet): use named exports.
+      "default",
+      // TODO(@wooorm-arcjet): use a clearer name: defaults for what, function to generate them?
+      "defaults",
+      "withVercelToolbar",
+    ]);
+  });
+});
--- a/nosecone/test/nosecone.test.ts
+++ b/nosecone/test/nosecone.test.ts
@@ -1,5 +1,5 @@
 import assert from "node:assert";
-import { describe, it } from "node:test";
+import { describe, it, test } from "node:test";
 
 import nosecone, {
   defaults,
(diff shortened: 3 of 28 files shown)


wooorm-arcjet:
not sure why this PR would hit `release-assets.githubusercontent.com` (caught by StepSecurity), but others don’t?

blaine-arcjet:
> not sure why this PR would hit release-assets.githubusercontent.com (caught by StepSecurity), but others don’t?

This is a flake because GItHub Runners only sometimes try to access this URL. There is a [blog post](https://www.stepsecurity.io/blog/harden-runner-detects-new-traffic-to-release-assets-githubusercontent-com-across-multiple-customers) about it. We need to get around to adding it to our allowlist of endpoints in all the CI flows.

ip/test/index.test.ts
```diff
@@ -0,0 +1,12 @@
+import assert from "node:assert/strict";
+import test from "node:test";
+
+test("@arcjet/ip", async function (t) {
+  await t.test("should expose the public api", async function () {
+    assert.deepEqual(Object.keys(await import("../index.js")).sort(), [
+      // TODO(@wooorm-arcjet): use named exports.
+      "default",
```
blaine-arcjet, line 8: This intentionally does not use named exports because the function is named `findIP` when is it much nicer to import as `ip` but that conflicts with a variable named `ip` in a lot of places. And I really don't like `import { findIP as ip }` nor `import { ip as findIP }`/

nosecone-next/test/index.test.ts
```diff
@@ -0,0 +1,15 @@
...
+import test from "node:test";
+
+test("@nosecone/next", async function (t) {
+  await t.test("should expose the public api", async function () {
+    assert.deepEqual(Object.keys(await import("../index.js")).sort(), [
+      "createMiddleware",
+      // TODO(@wooorm-arcjet): use named exports.
+      "default",
+      // TODO(@wooorm-arcjet): use a clearer name: defaults for what, function to generate them?
+      "defaults",
```
blaine-arcjet, line 11: `defaults` are used in spreads to override.

nosecone/test/nosecone.test.ts
```diff
@@ -20,6 +20,41 @@ import nosecone, {
   withVercelToolbar,
 } from "../index.js";
 
+test("nosecone", async function (t) {
+  await t.test("should expose the public api", async function () {
+    assert.deepEqual(Object.keys(await import("../index.js")).sort(), [
+      // TODO(@wooorm-arcjet): that’s a ton, perhaps make it smaller.
```
blaine-arcjet, line 26: This will actually get larger. There are a lot of security headers.

blaine-arcjet (approved):

wooorm-arcjet:
done in https://github.com/arcjet/arcjet-js/pull/4598

ip/test/index.test.ts
wooorm-arcjet: Hmm, I don’t see the problem that much, I think `findIp` sounds fine? Next to `parseProxy` and other future IP related things? 🤔
I get using `@arcjet/ip` as a package name for different IP related things. But I don’t see it as important to have one “blessed” export from it?

nosecone-next/test/index.test.ts
wooorm-arcjet: Right! What I wanted to point out:
a) with different exposed functions, it may get unclear for which function these are the defaults
b) adjacently, maybe nice to have `createDefaults()` to prevent mutation

nosecone/test/nosecone.test.ts
wooorm-arcjet: Would it be possible to have them in `nosecone/headers`? Or, `@nosecone/headers`?

Issue #4338, opened 2025-06-09
Improve readmes for individual packages


blaine-arcjet:
The quality of a README file varies dramatically for each package. We should standardize/improve the READMEs with minimal documentation.

blaine-arcjet:
I see @davidmytton assigned @qw-in but I'm going to assign @wooorm-arcjet because he's going to need to get a lay of the land and that's a perfect time to improve these.

davidmytton:
See https://github.com/arcjet/arcjet-js/blob/main/arcjet-next/README.md as an example of a full README: description of Arcjet, feature highlights, quick start, common code snippets and FAQs. We could add more common examples as well.

As a third party example, I have found https://github.com/workos/authkit-nextjs/blob/main/README.md to be useful recently. It covers everything I might want to do.

davidmytton:
They've all been standardized now.

PR #4633, opened 2025-07-15
docs: add uniform install section to readmes

wooorm-arcjet:
This PR adds a uniform install section to everything.

Some changes with reasons:

* `Install` instead of `Installation` — I just don’t see the need for the extra characters
* `sh` instead of `shell` — same, people often use shorter flags here rather than entire language names (they are interchangeable on GH and in all tools that I am aware of)
* no `-S` flag, which is the current behavior, and for people that choose to install with particular flags, this allows that.
* mention `npm` and specifically also `Node.js` and here
* explain that things are ESM only

Related-to: GH-4338.

### DIFF
--- a/analyze-wasm/README.md
+++ b/analyze-wasm/README.md
@@ -25,10 +25,13 @@ This package provides WebAssembly bindings to [Arcjet's][arcjet] local analysis 
 - [npm package (`@arcjet/analyze-wasm`)](https://www.npmjs.com/package/@arcjet/analyze-wasm)
 - [GitHub source code (`analyze-wasm/` in `arcjet/arcjet-js`)](https://github.com/arcjet/arcjet-js/tree/main/analyze-wasm)
 
-## Installation
+## Install
 
-```shell
-npm install -S @arcjet/analyze-wasm
+This package is ESM only.
+Install with npm in Node.js:
+
+```sh
+npm install @arcjet/analyze-wasm
 ```
 
 ## Use
--- a/arcjet-deno/README.md
+++ b/arcjet-deno/README.md
@@ -25,18 +25,14 @@ This is the [Arcjet][arcjet] SDK for [Deno][deno].
 - [npm package (`@arcjet/deno`)](https://www.npmjs.com/package/@arcjet/deno)
 - [GitHub source code (`arcjet-deno/` in `arcjet/arcjet-js`)](https://github.com/arcjet/arcjet-js/tree/main/arcjet-deno)
 
-<!--
-
 ## Install
 
 This package is ESM only.
-Install with XXX:
+Install with Deno:
 
 ```sh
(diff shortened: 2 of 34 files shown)


arcjet-deno/README.md
```diff
@@ -25,6 +25,19 @@ This is the [Arcjet][arcjet] SDK for [Deno][deno].
...
 
+<!--
+
+## Install
+
+This package is ESM only.
+Install with XXX:
+
+```sh
+TODO: what should we recommend?
```
qw-in: I would expect this to be

```

deno install npm:@arcjet/deno

```

Unless we are publishing to jsr? I have not had a chance to try deno yet, I can test it if you'd like

qw-in (approved):

arcjet-deno/README.md
wooorm-arcjet: I did that, that’s probably good!

analyze-wasm/README.md
```diff
@@ -25,10 +25,13 @@ This package provides WebAssembly bindings to [Arcjet's][arcjet] local analysis
 - [npm package (`@arcjet/analyze-wasm`)](https://www.npmjs.com/package/@arcjet/analyze-wasm)
 - [GitHub source code (`analyze-wasm/` in `arcjet/arcjet-js`)](https://github.com/arcjet/arcjet-js/tree/main/analyze-wasm)
 
-## Installation
+## Install
 
-```shell
-npm install -S @arcjet/analyze-wasm
+This package is ESM only.
```
qw-in, line 30: This is maybe a wee bit terse - is there a generic explainer we can link to? I think I've seen one before 🤔

analyze-wasm/README.md
```diff
@@ -25,10 +25,13 @@ This package provides WebAssembly bindings to [Arcjet's][arcjet] local analysis
...
+Install with npm in Node.js:
```
qw-in, line 31:
```suggestion
Install with npm for Node.js:
```

I think this makes slightly more sense to me, but very on the fence. Trying out combinations in my head I think I prefer it?`Install with npm [in|for] [nodejs|bun|deno]`

qw-in: llm thinks `in` is better so probably ignore me

analyze-wasm/README.md
wooorm-arcjet: I used to always link the the gist by Sindre: https://redirect.github.com/wooorm/dead-or-alive/tree/057704b01dc422f584eed2bb8294bae8cab40a09#install. But we’re getting to a time where people know ESM is a thing. So here I decided to drop it.

davidmytton (approved):

davidmytton (approved):

wooorm-arcjet:
Thanks for the review!

PR #4640, opened 2025-07-16
chore!: use `export` maps everywhere

wooorm-arcjet:
* add `export` fields for packages that didn’t have them; which makes our APIs private-by-default, preventing people from using internals, and makes it non-semver-breaking to move files around
* remove unneeded `main` fields
* remove potentially wrong `types` field; this field is not needed, because TS loads `.d.ts` next to a file, and this field is easy to get wrong with export maps, because the types are *different* for different files (to illustrate, some ESM file loaded with an `import` condition has different types than a CJS file at a `require` condition)

There is one case where different files (without conditions) are used from our other SDK packages: `client.js` in `@arcjet/protocol`. The change here is that that switches to `@arcjet/protocol/client`. We could consider exposing `createClient` from `@arcjet/protocol` itself?

An open question is, how should proto internals,
`@arcjet/protocol/proto/decide/v1alpha1/decide_pb.js` be exposed? `@arcjet/protocol/decide-v1-alpha.1` perhaps?

Related-to: GH-4587.

### DIFF
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -1343,10 +1343,12 @@ export function sensitiveInfo<
       if (typeof options.detect !== "undefined") {
         const detect = options.detect;
         convertedDetect = (tokens: string[]) => {
-          return detect(tokens)
-            .filter((e) => typeof e !== "undefined")
-            // @ts-ignore: TODO(@wooorm-arcjet): this fails when building on Bun, investigate.
-            .map(protocolSensitiveInfoEntitiesToAnalyze);
+          return (
+            detect(tokens)
+              .filter((e) => typeof e !== "undefined")
+              // @ts-ignore: TODO(@wooorm-arcjet): this fails when building on Bun, investigate.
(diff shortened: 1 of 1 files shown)


wooorm-arcjet:
Not sure why this affects the example type checking.
They do often have old `tsconfig`s; and the type errors are things we can and should improve in core.

wooorm-arcjet:
By using export maps, we can encapsulate our APIs. We can use conditions. We can shuffle files around. There are some benefits.
But we can’t really have those benefits if we attempt to *also* “support” `main` and `types`.

I have a good experience using just `exports` in my open source work

blaine-arcjet (changes requested):
Export maps have the same fundamental problem as Node's `internal` package: users cannot get at what they need to work around their own problems.

wooorm-arcjet:
We should land this, end users should not use private APIs.
End users can *still* use private APIs by importing file URLs.

PR #4641, opened 2025-07-16
refactor: replace export/import star with names

wooorm-arcjet:
Related-to: GH-4587.

### DIFF
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -48,7 +48,7 @@ import type {
 import { parse as parseDuration } from "@arcjet/duration";
 import ArcjetHeaders from "@arcjet/headers";
 import { runtime } from "@arcjet/runtime";
-// TODO(@wooorm-arcjet): discuss different names and remove `import *`.
+// TODO(@wooorm-arcjet): discuss exposing `hasher` as a named export.
 import * as hasher from "@arcjet/stable-hash";
 import { MemoryCache } from "@arcjet/cache";
 
--- a/stable-hash/test/index.test.ts
+++ b/stable-hash/test/index.test.ts
@@ -4,10 +4,7 @@ import test from "node:test";
 test("@arcjet/stable-hash", async function (t) {
   await t.test("should expose the public api", async function () {
     assert.deepEqual(Object.keys(await import("../index.js")).sort(), [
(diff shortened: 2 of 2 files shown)


arcjet/index.ts
```diff
@@ -33,16 +33,22 @@ import {
...
   SensitiveInfoEntity,
   BotConfig,
   EmailValidationConfig,
 } from "@arcjet/analyze";
-import * as duration from "@arcjet/duration";
+import { parse as parseDuration } from "@arcjet/duration";
 import ArcjetHeaders from "@arcjet/headers";
 import { runtime } from "@arcjet/runtime";
+// TODO(@wooorm-arcjet): discuss different names and remove `import *`.
 import * as hasher from "@arcjet/stable-hash";
```
blaine-arcjet: This package mirrors code in the Cloud Service which requires uniformity between them so please don't change it unless the Go code matches.

stable-hash/test/index.test.ts
```diff
@@ -4,6 +4,10 @@ import test from "node:test";
 test("@arcjet/stable-hash", async function (t) {
   await t.test("should expose the public api", async function () {
     assert.deepEqual(Object.keys(await import("../index.js")).sort(), [
+      // TODO(@wooorm-arcjet): discuss different names,
+      // `hashBool`, `hashString`, `hashUint32`.
+      // For `stringSliceOrdered`, the ordering seems to be implied for stable hashing,
+      // `slice` seems to mean array`, perhaps `arrayString` or `stringArray`?
```
blaine-arcjet: This mirrors Go code so please don't change it.

arcjet/index.ts
wooorm-arcjet: ok, though, if that is the intention, then I think it makes more sense to have a named export `hasher` which is an object that has stuff like `stringSliceOrdered`

wooorm-arcjet: amended the comments

stable-hash/test/index.test.ts
wooorm-arcjet: amended the comments

wooorm-arcjet:
Okay, it seems the decision is to keep using export/import stars.

PR #4654, opened 2025-07-17
docs: add info on derivative works to license sections

wooorm-arcjet:
Several readmes have `## Implementation` sections. Most but not all of those sections relate to which derivative works the implementation was inspired by and how it differs.

While this information often doesn’t have to be top of mind, it is important for people who wonder about license stuff. So I think it is best to put it *inside* the license section. Using a similar format. Also mentioning the copyright holders.

### DIFF
--- a/body/README.md
+++ b/body/README.md
@@ -59,6 +59,11 @@ console.log(body);
 
 [Apache License, Version 2.0][apache-license] © [Arcjet Labs, Inc.][arcjet]
 
+Derivative work based on [`raw-body`][node-raw-body] licensed under
+[MIT][node-raw-body-license] © Jonathan Ong and Douglas Christopher Wilson.
+Our work removes features that we do not use: no buffers, no sync interface.
+
 [arcjet]: https://arcjet.com
-[node-raw-body]: https://github.com/stream-utils/raw-body/blob/191e4b6506dcf77198eed01c8feb4b6817008342/test/index.js
+[node-raw-body-license]: https://github.com/stream-utils/raw-body/blob/191e4b6/LICENSE
+[node-raw-body]: https://github.com/stream-utils/raw-body/blob/191e4b6/test/index.js
 [apache-license]: http://www.apache.org/licenses/LICENSE-2.0
--- a/duration/README.md
+++ b/duration/README.md
@@ -57,7 +57,12 @@ console.log(seconds); // prints 3600
 
 [Apache License, Version 2.0][apache-license] © [Arcjet Labs, Inc.][arcjet]
 
+Derivative work based on [`time.ParseDuration`][go-parse-duration] from
+`golang/go` licensed under [BSD-3.0][go-parse-duration-license] © The Go Authors.
+Our work ports to TypeScript.
+
 [arcjet]: https://arcjet.com
-[go-parser]: https://github.com/golang/go/blob/c18ddc84e1ec6406b26f7e9d0e1ee3d1908d7c27/src/time/format.go#L1589-L1686
+[go-parse-duration-license]: https://github.com/golang/go/blob/c18ddc84e/LICENSE
+[go-parse-duration]: https://github.com/golang/go/blob/c18ddc84e/src/time/format.go#L1589-L1686
 [apache-license]: http://www.apache.org/licenses/LICENSE-2.0
 [github-vercel-ms]: https://github.com/vercel/ms
--- a/ip/README.md
+++ b/ip/README.md
@@ -74,7 +74,26 @@ are detected, an empty string is returned.
 
 [Apache License, Version 2.0][apache-license] © [Arcjet Labs, Inc.][arcjet]
 
+Derivative work based on [`Parser` in `std::net`][rust-parser],
+[`is_global` on `Ipv4Addr`][rust-ipv4-is-global], and
+[`is_global` on `Ipv6Addr`][rust-ipv6-is-global]
+from `rust-lang/rust`,
+dual licensed under [MIT][rust-license-mit] and
+[Apache-2.0][rust-license-apache] © contributors.
+Our work ports to TypeScript so that we have the same functionality in both
+languages.
+
+Derivative work based on [`getClientIp` from `request-ip`][request-ip-client-ip]
+licensed under [MIT][request-ip-license] © Petar Bojinov.
+Our work cherry picks only what we need.
+
 [arcjet]: https://arcjet.com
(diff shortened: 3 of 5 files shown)


qw-in (commented):
I think it all makes sense. I'm not confident I have the knowledge around licensing to know if saying these are "derivative work"s has any implications for Arcjet so I don't feel comfortable approving that specific bit

body/README.md
```diff
@@ -40,21 +40,15 @@ const body = await readBody(request, { limit: 1024 });
...
-## Implementation
-
-The implementation of this library is based on the [raw-body][node-raw-body]
-package. Licensed MIT with licenses included in our source code.
-
-We've chosen to re-implement the logic to read the body from the stream to keep
-the dependency tree for our packages as light as possible. Our implementation only
-provides the functionality that we need, specifically it excludes the functionality
-to return the stream as a buffer and always parses it as a utf-8 string. The interface
-was also changed to only support promises rather than the sync implementation provided by `raw-body`.
```
arcjet-rei: I think this is valuable context that should be retained, if possible. AIUI, we are under no particular restriction given the terms of the MIT license, so while it makes sense to move the context of our modifications into the `## License` sectsion, I would still keep the "Derivative work" sentence and the original explanation of the why of our modifications intact, as separate paragraphs.

duration/README.md
```diff
@@ -36,22 +36,15 @@ const seconds = parse("1h");
...
-
-The implementation of this library is based on the [ParseDuration][go-parser]
-function in the Go stdlib. Originally licensed BSD-3.0 with the license included
-in our source code.
-
-We've chosen the approach of porting this to TypeScript because our protocol
-operates exclusively on unsigned 32-bit integers representing seconds. However,
-we don't want to require our SDK users to manually calculate seconds. By
-providing this utility, our SDK can accept duration strings and numbers while
-normalizing the values for our protocol.
```
arcjet-rei: Similarly, I think it is valuable context to include the design goals (of working with unsigned 32-bit values capturing durations in seconds while still providing utilities for doing the duration calculations).

ip/README.md
```diff
@@ -56,29 +56,29 @@ platform is supplied in the `options` argument.
...
-[global ipv6 comparisons][rust-global-ipv6] in the Rust stdlib and the [header
-lookups][request-ip-headers] in the [request-ip] package. Both licensed MIT with
-licenses included in our source code.
-
-We've chosen the approach of porting Rust's IP Parser because capturing RegExps
-can be the source of ReDoS attacks, which we need to avoid. We also wanted to
-keep our implementation as close to Rust as possible because we will be relying
-on the Rust stdlib implementation in the future, with a fallback to this
-implementation. As such, we'll need to track changes in Rust's implementation,
-even though it seems to change infrequently.
```
arcjet-rei: Similarly, it is valuable (especially in a security-related product) to explicitly note when we have made a choice intended to minimize attack surface or mitigate potential vulnerabilities (I have encountered numerous ReDoS attacks in the wild).

sprintf/README.md
```diff
@@ -55,21 +55,17 @@ Object substitution supports any value that is not `undefined`.
...
 - `%%` - Replaced by the literal `%` character.
 
-## Implementation
-
-This implementation of this library is based on [quick-format-unescaped], which
-is licensed MIT with licenses included in our source code.
-
-The goal of this library is to be more restrictive than `quick-format-unescaped`
-while maintaining as much compatibility as possible, since [pino] uses it to
-format strings.
```
arcjet-rei: It seems valuable to note that this functionality is intended to ease working with our primary logging interface.

arcjet-rei (changes requested):
I agree with the decision to move the notes on implementation to the `## License` section, but I don't see the point of dropping so much useful context while doing so. The details removed are not extraneous and are valuable context for our developers, so moving the implementation notes is fine, but dropping them is not.

wooorm-arcjet:
@arcjet-rei Hmm, yes, looking at just this PR, some information is lost.
My main argument is that such information is better moved into separate sections:
`## What is this?` and `## When should I use this?` more higher up and relating to the current state, more for potential users,
and an optional `## Origin story` right above `## License` with more historic info, more for us.
Important also to note that there are several other packages where similar historic decisions were probably made, regarding their “raison d’être”, but that aren’t derivative works, and thus have no such section (yet).

davidmytton:
> My main argument is that such information is better moved into separate sections:

Ok with me - do you want to add those sections so we don't lose the info?

arcjet-rei:
> My main argument is that such information is better moved into separate sections: `## What is this?` and `## When should I use this?` more higher up and relating to the current state, more for potential users, and an optional `## Origin story` right above `## License` with more historic info, more for us.

What you're describing sounds like a refactor of the documentation and documentation comments, and I don't object to that in principle. However, if you're moving the documentation to a separate section, then I would expect the new sections you describe to be part of the same changeset so that details don't get lost as part of the reorganization.

arcjet-rei:
Minor nit: "derivate" is a word, but it's not really the right word for what you're describing here, which is either a "derivative work" or a "derived product."

body/README.md
wooorm-arcjet: Done in https://github.com/arcjet/arcjet-js/pull/4789

duration/README.md
wooorm-arcjet: Done in https://github.com/arcjet/arcjet-js/pull/4789

ip/README.md
wooorm-arcjet: Done in https://github.com/arcjet/arcjet-js/pull/4789

sprintf/README.md
wooorm-arcjet: Done in https://github.com/arcjet/arcjet-js/pull/4789

wooorm-arcjet:
> derivate

No clue how that got in there. I made the same mistake a couple of times too. Weird.

> However, if you're moving the documentation to a separate section, then I would expect the new sections you describe to be part of the same changeset so that details don't get lost as part of the reorganization.

Moved in https://github.com/arcjet/arcjet-js/pull/4789 too.

An understandable wish, but because of the large number of packages and the complete rewrite I am slowly doing, I am splitting by section, or even batching that into a couple of readmes per PR. I hope that’s acceptable.

arcjet-rei (approved):
Removing my block on this because @wooorm-arcjet has assured me that the context will be restored in a subsequent changeset. I still think that separating the removal of information and its addition somewhere else is a bad practice liable to result in things being lost and should be avoided in general. FYI, I will be more strict about this over time.

PR #4655, opened 2025-07-17
docs: add uniform use section to readmes

wooorm-arcjet:
This PR adds a uniform use section to everything.

Many packages had few examples.
Several didn’t work anymore.
Packages that integrate with other tools have many examples on the website.
`@arcjet/next` had a lot of (outdated) information.

This adds one `## Use` section right under `## Install` that shows a *minimal*
example of how a package can work.
My idea is that you can paste it into an `example.ts`,
install things,
and it works.
It’s not a goal to show everything that’s possible with a package.

For that reason,
I opted to only show `shield` with the integrations.
And to have them all do the same things,
so that people can see how Arcjet is (typically) the same and only the other
tools differ.

How to use rules in your client,
and all other ways to use arcjet and integrations,
is then left to the website.

Note that I do want to add `## API` sections too.
I *do* want to cover everything that’s exposed in readmes.
Just not in the top-most simple example.

Related-to: GH-4338.


arcjet-astro/README.md
```diff
@@ -40,204 +40,58 @@ Try an Arcjet protected app live at [https://example.arcjet.com][example-url]
...
-The rule is defined in your `astro.config.mjs` file:
-
-```js
-// @ts-check
-import { defineConfig } from "astro/config";
 import arcjet, { shield } from "@arcjet/astro";
+import { defineConfig } from "astro/config";
 
-// https://astro.build/config
 export default defineConfig({
```
qw-in, line 51: Astro needs a server adapter configured before Arcjet will run

arcjet-astro/README.md
```diff
@@ -40,204 +40,58 @@ Try an Arcjet protected app live at [https://example.arcjet.com][example-url]
...
+        // Use `DRY_RUN` instead of `LIVE` to only log.
+        shield({ mode: "LIVE" }),
       ],
     }),
   ],
 });
 ```
 
-Then Arcjet is called from within this page route:
+…then use Arcjet in dynamic Astro pages (such as `src/pages/api.json.ts`):
```
qw-in:
```suggestion
…then use Arcjet in on-demand routes (such as `src/pages/api.json.ts`):
```

It's been tricky getting this wording to match their docs. I probably need to do another pass on the docs too, but I think this is the most accurate term? It's quite confusing as the exported flag is `prerender = false` as well

https://docs.astro.build/en/guides/routing/#on-demand-dynamic-routes

arcjet-astro/README.md
```diff
@@ -40,204 +40,58 @@ Try an Arcjet protected app live at [https://example.arcjet.com][example-url]
...
-  } else {
-    return Response.json({ message: "Hello world" });
+    return Response.json({ message: "Forbidden" }, { status: 403 });
   }
+
+  return Response.json({ message: "Hello world" });
 };
 ```
 
+For more on how to configure Arcjet with Astro and how to protect Astro,
```
qw-in, line 84: Maybe where applicable we can link the full example app as well?

arcjet-bun/README.md
```diff
@@ -40,94 +40,53 @@ Try an Arcjet protected app live at [https://example.arcjet.com][example-url]
...
-
-## Shield example
+import arcjet, { shield } from "@arcjet/bun";
 
-[Arcjet Shield][shield-concepts-docs] protects your application against common
-attacks, including the OWASP Top 10. You can run Shield on every request with
-negligible performance impact.
+// Get your Arcjet key at <https://app.arcjet.com>.
+// Set it as an environment variable instead of hard coding it.
+const arcjetKey = process.env.ARCJET_KEY;
```
qw-in: Does Bun no longer use Bun.env?

arcjet-deno/README.md
```diff
@@ -25,93 +25,57 @@ This is the [Arcjet][arcjet] SDK for [Deno][deno].
...
-negligible performance impact.
+TODO(@wooorm-arcjet): This is missing?
 
-```ts
-import "jsr:@std/dotenv/load";
+For more on how to configure Arcjet with Deno and how to protect Deno,
+see the [Arcjet Deno SDK reference][arcjet-reference-deno] on our website.
 
-import arcjet, { shield } from "npm:@arcjet/deno";
+[arcjet-reference-deno]: https://docs.arcjet.com/reference/deno
```
qw-in, line 71: Should this be with the other links? Tricky to tell in the diff

arcjet/README.md
```diff
@@ -40,77 +40,70 @@ integration.
 npm install -S arcjet
```
qw-in, line 40: Do we want people to use this package? Maybe worth removing and saying internal package if we don't

qw-in (commented):
Looked through a good portion of the diff, will come back to get the last bit. 

I think beyond specific comments I'm wondering if we can link the full example apps where applicable after the use section?

arcjet-astro/README.md
wooorm-arcjet: I am sorry, I do not understand what you mean

arcjet-astro/README.md
wooorm-arcjet: I... do not know. So, sure!

arcjet-astro/README.md
wooorm-arcjet: Aren’t they linked from there?

They are also linked above already for these integrations.
I plan on looking at the other sections, and that will then cover these example sections, perhaps we can discuss it then?

arcjet-bun/README.md
wooorm-arcjet: Didn’t think of that. That also works. Probably more bun-y to use that 🐰

arcjet-deno/README.md
wooorm-arcjet: When it is back. It’s commented out: as it doesn’t exist yet!

arcjet/README.md
wooorm-arcjet: Not really no, but a) Install sections are GH-4633. And b) there are several packages that we do not want people to use, that should probably go higher up as a big note!

Though, having people make adapters/integrations themselves doesn’t seem too bad to me.

arcjet-astro/README.md
qw-in: I believe a server adapter needs to be configured before [on-demand rendering](https://docs.astro.build/en/guides/on-demand-rendering/) will work (and Arcjet only protects on-demand routes). IIRC this is true in development too

arcjet-astro/README.md
qw-in: Ahh, maybe I missed them - sorry about that

arcjet-deno/README.md
qw-in: Another example + reference for my list then 😀

wooorm-arcjet: indeed ;)

arcjet-astro/README.md
wooorm-arcjet: So are you saying that that should be described before this code block? A link to https://docs.astro.build/en/guides/on-demand-rendering/#server-adapters, and stuff of adding `export const prerender = false` to pages?

Doesn’t the next code block, introduced with: “…then use Arcjet in on-demand routes”, provide an alternative?

qw-in: Astro will prompt them as well in an error message as well so they'll probably find it and can probably omit here as you say

arcjet-rei (approved):
It took me a while, but I looked at all the changed files, and this all looks good to me.

wooorm-arcjet:
Thanks for taking all that time, appreciate it!

PR #4666, opened 2025-07-18
build(examples): add strict type checking

wooorm-arcjet:
This adds a `typecheck` npm script that runs a type check with `strict: true`.
Otherwise, I removed everything that was not needed from `tsconfig.json`s.
And applied some small fixes that were caught by the stricter type requirement.
This should help us catch more problems.

I also noticed that the following are not used in the reusable examples workflow.

* `bun-hono-rate-limit`
* `bun-rate-limit`
* `deno-sensitive-info`
* `express-bots`
* `express-newman`
* `express-sensitive-info`
* `nextjs-react-hook-form`
* `nodejs-express-launchdarkly`
* `nodejs-express-rate-limit`
* `nodejs-express-validate-email`
* `nodejs-rate-limit`

### DIFF
--- a/.github/workflows/reusable-examples.yml
+++ b/.github/workflows/reusable-examples.yml
@@ -54,6 +54,9 @@ jobs:
         working-directory: examples/gatsby-rate-limit
         run: npm run build
 
+      - working-directory: examples/gatsby-rate-limit
+        run: npm run typecheck
+
   nestjs:
     name: NestJS
     runs-on: ubuntu-latest
@@ -96,6 +99,9 @@ jobs:
       - name: Build
         working-directory: examples/nestjs
         run: npm run build
+
+      - working-directory: examples/nestjs
+        run: npm run typecheck
 
   nestjs-fastify:
     name: NestJS + Fastify
@@ -140,6 +146,9 @@ jobs:
         working-directory: examples/nestjs-fastify
         run: npm run build
 
+      - working-directory: examples/nestjs-fastify
+        run: npm run typecheck
+
   nestjs-graphql:
     name: NestJS + GraphQL
     runs-on: ubuntu-latest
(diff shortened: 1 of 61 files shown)


.github/workflows/reusable-examples.yml
```diff
@@ -54,6 +54,9 @@ jobs:
         working-directory: examples/gatsby-rate-limit
         run: npm run build
 
+      - working-directory: examples/gatsby-rate-limit
```
arcjet-rei, line 57: I know that an obsession with DRY is a hobgoblin of little minds, but this file is getting very large. This leads to a question: at what size / what number of repeated actions does it make sense to either split the file down into smaller pieces, or to templatify the action, given that all of the examples seem to have identical build steps and almost no parameterization. I don't object to the verbosity of the file, as I do see explicit / concrete specification of system actions as a plus (and also I'm not ultimately the one responsible for day-to-day maintenance of this file).

examples/sveltekit/package.json
```diff
@@ -10,7 +10,8 @@
 		"check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json",
 		"check:watch": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json --watch",
 		"lint": "prettier --check . && eslint .",
-		"format": "prettier --write ."
+		"format": "prettier --write .",
+    "typecheck": "tsc --noEmit"
```
arcjet-rei: Whitespace?

arcjet-rei (approved):
One nit I leave to you to resolve before merging, and one non-blocking question / discussion point. Otherwise, I like the simplification and that even with that simplification, very few changes were required to the code.

.github/workflows/reusable-examples.yml
wooorm-arcjet: I very much agree. I want to sort the file, add missing examples, and then indeed figure out if all the steps can be the same for each example, and then only use the folder name as a variable.

It’s currently also hard to figure out from a CI failure what folder the problem is in, as the names are a bit different. Using the folder name as the variable will help I think.

Currently, there is *intended* difference between these examples, they sometimes come from what that framework recommends to start with for tooling and only add arcjet. However, all that difference does make it hard to maintain for us.

wooorm-arcjet: see https://github.com/arcjet/arcjet-js/pull/4692!

PR #4692, opened 2025-07-21
build(examples): simplify reusable examples

wooorm-arcjet:
This removes all the repetition per example folder and instead uses a matrix for all the repetition. Cutting down 1.2k lines!

There is a teeny tiny downside that Astro requires an `ARCJET_KEY` and the ±4 Clerk examples require a `NEXT_PUBLIC_CLERK_PUBLISHABLE_KEY`, which are now set for the build of every example. But, these are both fake tokens. So no danger. So not a real problem in my opinion.

I did look for a second at the missing examples. They are missing for various reasons. Which can be solved in a follow up:

build broken:
- `nextjs-react-hook-form`

build missing:
- `express-sensitive-info`
- `express-bots`
- `express-newman`
- `nodejs-express-rate-limit`
- `nodejs-express-launchdarkly`
- `nodejs-express-validate-email`
- `nodejs-rate-limit`

bun:
- `bun-hono-rate-limit`
- `bun-rate-limit`

deno:
- `deno-sensitive-info`

Related-to: GH-4666 😈


wooorm-arcjet:
~~Q for folks: anyone have or know of a super simple workflow to run deno, bun?~~

Found some, see: https://github.com/arcjet/arcjet-js/pull/4693.

arcjet-rei (approved):
I like it when things are cut by 1.2K lines. It feels like there's an opportunity for collaboration between @arcjet/js-sdk-team and @arcjet/dx-team to share effort and / or simplify things here, but for now I'll leave that for @qw-in and @wooorm-arcjet to work out over time.

wooorm-arcjet:
Sweet!!

PR #4693, opened 2025-07-21
build(examples): add bun, deno examples

wooorm-arcjet:
This PR adds installation and type checking of the Bun and Deno examples.
For Deno, Node.js is still used to install and build the entire monorepo.
As Bun is closer to Node.js, it also installs and builds the entire monorepo.

Important to note, that these 3 examples were not used much, and I don’t think Dependabot looks at them.
This does also update their locks. I looked through them and they look normal to me.

Related-to: GH-4692.


wooorm-arcjet:
Thanks for reviewing @arcjet-rei!

PR #4694, opened 2025-07-21
test: add tests for unsupported headers per platform

wooorm-arcjet:
I noticed that the implementation specifically does not support most headers if a particular `platform` is set while working on GH-4632. And, in some cases, when a platform is not set, several platform-specific headers are not supported either. Behavior that makes sense when you think about it.

This commit thus adds checks for when headers are not supported.

Related-to: GH-4632.

### DIFF
--- a/ip/test/ip.test.ts
+++ b/ip/test/ip.test.ts
@@ -305,91 +305,268 @@ test("`findIp`", async (t) => {
     }
   });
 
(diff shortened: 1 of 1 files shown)


arcjet-rei (approved):
LGTM

wooorm-arcjet:
Thanks for looking at my code @arcjet-rei!

PR #4695, opened 2025-07-21
fix(ip): support readonly arrays, null, and exact optional props

wooorm-arcjet:
When working on GH-4538 I noticed that readonly arrays were not allowed in the types but looking at the code it does seem intentional that parameters are not mutated (a good thing to not touch them of course).

Additionally users that configure TypeScript with
`exactOptionalPropertyTypes`, would get a type error if an `undefined` value was explicitly passed, instead of an option not set. This adds `undefined` as an allowed type for options next to being omitted.

Finally, I explicitly added `null` in the allowed types for options and parameters too.
While I do hope that people use `undefined` in almost all cases, I think that it is good for our APIs to *handle* `null` as well.

Related-to: GH-4538.

### DIFF
--- a/arcjet-fastify/index.ts
+++ b/arcjet-fastify/index.ts
@@ -301,8 +301,7 @@ function toArcjetRequest<Properties extends PlainObject>(
 
   let ip = findIp(
     { headers, socket: request.socket },
-    // TODO(@wooorm-arcjet): readonly support in `findIp`.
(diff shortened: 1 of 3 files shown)


arcjet-rei (approved):
LGTM

wooorm-arcjet:
Yay! Thanks @arcjet-rei!

Issue #16, opened 2023-12-11
Remove TypeScript DOM stdlib usage


blaine-arcjet:
According to the node typescript type maintainers in https://github.com/DefinitelyTyped/DefinitelyTyped/pull/66824, people should never combine `@types/node` and the `DOM` types from the TS standard lib. However, the node types don't contain types for the WebAssembly global and they only exist in `DOM`, so we need to use them right now. However, the Headers type in `@types/node` conflicts with the Headers in `DOM`.

We should contribute the WebAssembly types (and any other we are relying on from `DOM`) to `@types/node` so we don't need the TS standard library.

PR #4721, opened 2025-07-24
build(examples): add missing examples

wooorm-arcjet:
This PR adds 9 missing examples to the example build.
It adds strict type checking to them, which caught some potential bugs.

There were 3 `express-*` examples and 3 `node-express-*` examples, I have them the same name (removing `node-` as a prefix).

This does change some package locks, as it installs `@types/express`. I looked through them and verified it looks good.


blaine-arcjet:
Heads up: renaming examples likely requires URL changes in the docs.

wooorm-arcjet:
Not sure these are referenced anywhere: https://github.com/search?q=org%3Aarcjet+-repo%3Aarcjet%2Farcjet-js+nodejs-express&type=code. But good point and /cc @qw-in

examples/express-bots/tsconfig.json
```diff
@@ -1,8 +1,9 @@
 {
-    "compilerOptions": {
-        "allowJs": true,
-        "noEmit": true
-    },
-    "include": ["index.js"],
-    "exclude": ["node_modules"]
+  "compilerOptions": {
+    "checkJs": true,
+    "lib": ["dom.iterable", "dom", "esnext"],
```
qw-in, line 4: Why does an express app need dom?

examples/express-bots/tsconfig.json
```diff
@@ -1,8 +1,9 @@
...
+    "module": "node16",
```
qw-in, line 5: Why not nodenext?

examples/express-sensitive-info/index.js
```diff
@@ -2,12 +2,18 @@ import arcjet, { sensitiveInfo, shield } from "@arcjet/node";
 import express from "express";
 
 const app = express();
-const port = 3000;
+const port = 300
```
qw-in: Seems accidental

examples/express-bots/tsconfig.json
wooorm-arcjet: because we do: iteration over `Headers` is used.

I’d like to get away from needing it. I think that’s also what https://github.com/arcjet/arcjet-js/issues/16 is about

examples/express-bots/tsconfig.json
wooorm-arcjet: Not much thought in it. `node16` just feels like a more locked down version than `nodenext`, the latter seems like a sliding window that would change in the future

examples/express-sensitive-info/index.js
wooorm-arcjet: good catch, done!

wooorm-arcjet:
Yay, thanks all!

PR #4722, opened 2025-07-24
types: add `exactOptionalPropertyTypes`

wooorm-arcjet:
This adds `exactOptionalPropertyTypes` to our packages’ `tsconfig.json`s.
Some people (including us) pass `undefined` as a value, instead of totally omitting a field/parameter.
They are two slightly different things, and this is now supported.

Notably, I did *not* touch the wasm built stuff, so the code around that is slightly awkward, as I am now explicitly omitting fields (as the wasm types want) instead of passing `undefined`.

### DIFF
--- a/analyze-wasm/tsconfig.json
+++ b/analyze-wasm/tsconfig.json
@@ -1,4 +1,7 @@
 {
+  "compilerOptions": {
+    "exactOptionalPropertyTypes": true
+  },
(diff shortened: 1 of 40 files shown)


blaine-arcjet:
I also don't really like this but if you were going to change it, it should have been in the base config and then opt-out in places where it doesn't work.

Issue #4478, opened 2025-06-20
Policy on casing abbreviations in identifiers


wooorm-arcjet:
I see methods such as `hasASN` and `isVpn`.

The more code is added, the more confusing similar cases will be added, a good moment to decide/fix things is now, before semver stable.

My preference is going with abbreviations as “normal” words: `hasAsn`, `isVpn`, `XmlHttpRequest`, `getElementById`, `fileUrlToPath`, `randomUuid`, `readIpv6Address`.
And, in some cases, the problem could also be side-stepped by not using abbreviations: `getElementByIdentifier`

Useful regex: `\w+?[a-z][A-Z]{2,}\w+?(?="|\()`

davidmytton:
Is there any prior art we can review from popular libraries to see how others do this? In the Go ecosystem we try to follow [Google's decision on initialisms](https://google.github.io/styleguide/go/decisions#initialisms):

> Words in names that are initialisms or acronyms (e.g., URL and NATO) should have the same case. URL should appear as URL or url (as in urlPony, or URLPony), never as Url. As a general rule, identifiers (e.g., ID and DB) should also be capitalized similar to their usage in English prose.

But we should do whatever feels most native to the language we're writing in.

wooorm-arcjet:
I like https://w3c.github.io/aria/#idl_attr_disambiguation, though their constraints are `aria-` prefix and that HTML attributes *otherwise* have no dashes (and no casing). The last point there says “(e.g., `ID` becomes `Id`).”

Important to note that in the end it is based on “what words are”. And that becomes vague, especially in a space with jargon: we developers treat “inline” and “fullscreen” as words, but are they really to “normal” people?

Then, there are always inconsistencies. I recall somewhat recently the difference between `document.fullscreenEnabled`, `document.webkitFullscreenEnabled`, `document.mozFullScreenEnabled`.

In the end, I think it’s best to think of how it should be dash-cased, and camelcase from that, thus favoring a simple, more programatic algorithm (like https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#name_conversion, ignoring the `data-` specific handling). I mentioned that in https://github.com/arcjet/arcjet-js/pull/4723#issuecomment-3116962336:

[...]

PR #4723, opened 2025-07-24
refactor(ip): rename identifiers to match other casing

wooorm-arcjet:
When working on the fastify integration, i used `@arcjet/ip`.

I noticed that I needed the `Cidr` type, which is a return value of `parseProxy` from there.
So I expose it here.

I also noticed that there are 2 functions, `parseProxy` as a named export and `findIp` as the default export.
I expose the latter as a named export too, and deprecate the default symbol.

Internally, the casing of abbreviations was more capital heavy, and not always consistent (`parseCIDR` and `isIPv4Cidr`). Especially around symbols such as `isIPv4Cidr` (previous casing), I find it hard to come up with good arguments why that `P` is uppercase but `v` lowercase.
I like using capitals only for word boundaries: `parseCidr`, `isIpv4Cidr`, `Cidr`.
So, before exposing values, I changed their internal names. See GH-4478 for more on that.

### DIFF
--- a/arcjet-astro/internal.ts
+++ b/arcjet-astro/internal.ts
@@ -9,7 +9,7 @@ import type {
   Arcjet,
   CharacteristicProps,
 } from "arcjet";
-import findIP, { parseProxy } from "@arcjet/ip";
+import findIp, { parseProxy } from "@arcjet/ip";
 import { ArcjetHeaders } from "@arcjet/headers";
 import { baseUrl, isDevelopment, logLevel, platform } from "@arcjet/env";
 import { Logger } from "@arcjet/logger";
@@ -227,7 +227,7 @@ export function createArcjetClient<
     const headers = new ArcjetHeaders(request.headers);
 
     const url = new URL(request.url);
-    let ip = findIP(
+    let ip = findIp(
       {
         ip: clientAddress,
         headers,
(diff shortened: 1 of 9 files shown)


blaine-arcjet:
IPv4 and IPv6 are specific terms: https://en.wikipedia.org/wiki/IPv4, https://en.wikipedia.org/wiki/IPv6
CIDR is abbreviated as all caps: https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing

Stylistic changes like this grow the scope of a PR and make it harder to review the functional changes. We should keep PRs focused on the task outlined in the semantic commit. I suspect these crept in due to the desire to name the external API `findIp` instead of `findIP`, the bikeshedding of which is why I exported it as a default when I wrote the library.

wooorm-arcjet:
Internally different cases are already used, see my OP: `parseCIDR`, `isIPv4Cidr`.
`GitHub` and `npm` are also specific terms and in normal usage cased a certain way, but, that casing I believe should not be used in camelcased symbol names. More on that in OP and in GH-4478.

Another way how I think of it: come up with a lower dash case name, and then turn `-[a-z]` into caps. So, `parse-cidr`, `is-ipv4-cidr`, `from-github`, `to-npm` -> `parseCidr`, `isIpv4Cidr`, `fromGithub`, `toNpm`.

As for 2 PRs, if you want that I will do that but I didn’t because that’s what GH-4478 is about.
In the mean time I split the commits and marked them as being about renaming internals or about exposing values

wooorm-arcjet:
I split this work. This PR is now only refactoring. I’ll export the type, and the named export, in follow up PRs!

PR #4724, opened 2025-07-24
docs(fastify): mention quick start example in readme

wooorm-arcjet:
I noticed that the fastify quick start is now live, so, added it.

@qw-in There are still commented out references to `https://docs.arcjet.com/reference/fastify` and `https://example.arcjet.com` for when that is live.

@qw-in I also noticed that there are many references to `https://github.com/arcjet/arcjet-js-example` as a link to the source code of `https://example.arcjet.com`. It looks like that now redirects to `https://github.com/arcjet/example-nextjs`. Is that correct, for the source code of *all* examples? Should there be another URL?

### DIFF
--- a/arcjet-fastify/README.md
+++ b/arcjet-fastify/README.md
@@ -23,17 +23,9 @@ against common attacks.
 
 This is the [Arcjet][] SDK for [Fastify][].
 
-<!--
-
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


davidmytton (approved):

wooorm-arcjet:
Thanks both for the review!

Issue #1029, opened 2024-06-25
Should we include typescript in peerDependencies?


davidmytton:
I had a user report a compile error which turned out to be because they were using TypeScript 4.9.5. Updating to the latest version (5.5.2) solved this. Should we add TypeScript >= to peerDependencies?

```
21:28:15.573 • Next.js 14.2.3
--
21:28:15.574 |  
21:28:15.575 | Linting and checking validity of types
21:28:24.363 | Failed to compile.
21:28:24.363
21:28:24.364 /node_modules/-pnpm/@arcjet+next@1.0.0-alpha.13_@bufbuild+protobuf@1.10.0_@connectrpc+connect@1.4.0_next@14.2.3/node_modules/@arcjet/next/index.d.ts:84:36
21:28:24.364 Type error: Type parameter declaration expected.
21:28:24.364 |  
21:28:24.364 | 82 | * @param options - Arcjet configuration options to apply to all requests.
21:28:24.364 | 83 | * /
21:28:24.364 | > 84 export default function arcjetNext<const Rules extends (Primitive | Product)[]>(options: ArcjetOptions<Rules>): ArcjetNext<Simpify<ExtraProps<Rules>>>;
21:28:24.364 |  
21:28:24.364 | 85 /**
21:28:24.364 | 86 | * Protects your Next.js application using Arcjet middleware. 
21:28:24.364 | 87
21:28:24.399 | ELIFECYCLE Command failed with exit code 1.
21:28:24.420 | Error: Command "pnpm run build" exited with 1
21:28:24.977

```

blaine-arcjet:
I don't believe I've ever seen a project specify typescript as a peerDependency so I don't think we should.

Additionally, peerDependencies wouldn't solve the issue here because the newer TypeScript would be installed in `node_modules/@arcjet/next/node_modules/typescript` and they would still be compiling with typescript 4.9.5 which they installed at the top level.

wooorm-arcjet:
I second Blaine.

TS is more like Node or Bun or another engine. And a proper solution then is, I believe, to have:

a) some policy on how far back to support TS, e.g., ±2 years (sadly there is nothing like [`nodejs/release`](https://github.com/nodejs/release#nodejs-release-working-group) for TS),
b) document that somewhere,
c) somehow test that policy (far future nice to have)
d) only update the supported version in major releases

[...]

PR #4739, opened 2025-07-28
deps: update and unlock `@types/node`

wooorm-arcjet:
`@types/node` was locked at an old version in dependabot config.
This unlocks that.
This also updates it to the most recent version.


wooorm-arcjet:
trunk reports an “issue” that is already in the `package-lock.json`

blaine-arcjet:
We don't want to use the latest version, but instead the version synced with our lowest supported Node version. This allows us to only rely on types for APIs that exist in that version of Node.

I believe that will be Node v20 since we are dropping v18

> trunk reports an “issue” that is already in the package-lock.json

This is also a security issue that you need to resolve: https://github.com/arcjet/arcjet-js/security/dependabot/427

wooorm-arcjet:
> This allows us to only rely on types for APIs that exist in that version of Node.

CI already catches that and better (and it will get better).
We support a range of Node.js, not just the oldest Node. We use APIs available in modern Node. `Headers` is an example.
If it was a peer dependency, sure, `>=18`.

davidmytton (commented):
This was pinned because:

```
# Ignore updates to the @types/node package due to conflict between
# Headers in DOM.
```

Has that been resolved?

wooorm-arcjet:
Yes! This PR working suggests that that is improved, and quickly locally trying that out too! https://github.com/arcjet/arcjet-js/issues/16! I want to remove them next (though with all the examples there may be legitimate use of `dom`, or incorrect use by other tools!)

arcjet-rei:
When @arcjet/engineering-team discussed this in a status meeting last week, there was an issue raised about the maximum version of `@types/node` that we wanted to allow for backwards compatibility reasons. Can @davidmytton or @blaine-arcjet summarize that discussion here so we can figure out how to capture that policy (preferably in an ADR) and figure out what to do with this PR?

wooorm-arcjet:
It was discussed in https://github.com/arcjet/arcjet-js/issues/1029 and https://github.com/arcjet/arcjet-js/pull/4517

Issue #3327, opened 2025-02-25
Improve JSDoc comments


davidmytton:
We have some JSDoc comments which show up in your IDE on hover, but not everything is documented e.g. rules like `shield` and `slidingWindow`. We should improve coverage and add more detailed help/examples. This is helpful for developers and AI alike.

PR #4741, opened 2025-07-28
docs: add JSDocs to exposed API in 10 utilities

wooorm-arcjet:
This commit adds JSDoc documentation to everything exposed from 10 of the about 30 packages.

There’s a lot we are exposing that we don’t need to expose, but also things that are in parameters/return values/elsewhere which we should probably expose, which I noted inline.

As for `@arcjet/inspect`, that’s something that was already documented, but also (more up to date) on the website. I kept some small useful things in the code and the readme here, but mostly deferred with links to the website.

Related-to: GH-3327.

### DIFF
--- a/analyze/index.ts
+++ b/analyze/index.ts
@@ -37,9 +37,13 @@ type AnalyzeRequest = {
 };
 
 export {
+  // TODO(@wooorm-arcjet): add JSDocs to values from WebAssembly.
   type EmailValidationConfig,
+  // TODO(@wooorm-arcjet): add JSDocs to values from WebAssembly.
   type BotConfig,
+  // TODO(@wooorm-arcjet): add JSDocs to values from WebAssembly.
   type SensitiveInfoEntity,
+  // TODO(@wooorm-arcjet): add JSDocs to values from WebAssembly.
   type DetectedSensitiveInfoEntity,
 };
 
--- a/analyze/index.ts
+++ b/analyze/index.ts
@@ -111,7 +111,7 @@ function createCoreImports(detect?: DetectSensitiveInfoFunction): ImportObject {
  * @param request
  *   Request.
  * @returns
- *   Promise to a SHA-256 fingerprint.
+ *   Promise for a SHA-256 fingerprint.
  */
 export async function generateFingerprint(
   context: AnalyzeContext,
@@ -144,7 +144,7 @@ export async function generateFingerprint(
  * @param options
  *   Configuration.
  * @returns
- *   Promise to a result.
+ *   Promise for a result.
  */
 export async function isValidEmail(
   context: AnalyzeContext,
@@ -175,7 +175,7 @@ export async function isValidEmail(
  * @param options
  *   Configuration.
  * @returns
- *   Promise to a result.
+ *   Promise for a result.
  */
 /// TODO(@wooorm-arcjet): expose `BotEntity`, `BotResult`.
 export async function detectBot(
@@ -211,7 +211,7 @@ export async function detectBot(
  * @param detect
  *   Function to detect sensitive info (optional).
  * @returns
- *   Promise to a result.
+ *   Promise for a result.
  */
 // TODO(@wooorm-arcjet): expose `DetectSensitiveInfoFunction`, `SensitiveInfoEntities`, `SensitiveInfoResult`.
 // TODO(@wooorm-arcjet): less parameters, `5` is a lot.
--- a/analyze/index.ts
+++ b/analyze/index.ts
@@ -1,10 +1,14 @@
 import { initializeWasm } from "@arcjet/analyze-wasm";
 import type {
+  // TODO(@wooorm-arcjet): rename to `DetectBotOptions`.
   BotConfig,
+  // TODO(@wooorm-arcjet): expose, rename to `DetectBotResult`.
   BotResult,
   DetectedSensitiveInfoEntity,
   DetectSensitiveInfoFunction,
+  // TODO(@wooorm-arcjet): rename to `ValidateEmailOptions`.
   EmailValidationConfig,
+  // TODO(@wooorm-arcjet): expose, rename to `ValidateEmailResult`.
   EmailValidationResult,
   SensitiveInfoEntities,
   SensitiveInfoEntity,
@@ -13,11 +17,13 @@ import type {
 } from "@arcjet/analyze-wasm";
 import type { ArcjetLogger } from "@arcjet/protocol";
 
+/// TODO(@wooorm-arcjet): expose, as `Context`.
 interface AnalyzeContext {
   log: ArcjetLogger;
   characteristics: string[];
 }
 
+/// TODO(@wooorm-arcjet): expose, as `Request`.
 type AnalyzeRequest = {
   ip?: string;
   method?: string;
@@ -92,13 +98,20 @@ function createCoreImports(detect?: DetectSensitiveInfoFunction): ImportObject {
   };
 }
 
-// TODO(@wooorm-arcjet): document what is used to fingerprint.
 /**
- * Generate a fingerprint for the client. This is used to identify the client
- * across multiple requests.
- * @param context - The Arcjet Analyze context.
- * @param request - The request to fingerprint.
- * @returns A SHA-256 string fingerprint.
+ * Generate a fingerprint.
+ *
+ * Fingerprints can be used to identify the client across multiple requests.
+ *
+ * This considers different things on the `request` based on the passed
+ * `context.characteristics`.
+ *
+ * @param context
+ *   Context.
+ * @param request
+ *   Request.
+ * @returns
+ *   Promise to a SHA-256 fingerprint.
  */
 export async function generateFingerprint(
   context: AnalyzeContext,
@@ -121,10 +134,21 @@ export async function generateFingerprint(
   return "";
 }
 
-// TODO(@wooorm-arcjet): docs.
+/**
+ * Check whether an email is valid.
+ *
+ * @param context
+ *   Context.
+ * @param value
+ *   Value.
+ * @param options
+ *   Configuration.
+ * @returns
+ *   Promise to a result.
+ */
 export async function isValidEmail(
   context: AnalyzeContext,
-  candidate: string,
+  value: string,
   options: EmailValidationConfig,
 ): Promise<EmailValidationResult> {
   const { log } = context;
@@ -132,7 +156,7 @@ export async function isValidEmail(
   const analyze = await initializeWasm(coreImports);
 
   if (typeof analyze !== "undefined") {
-    return analyze.isValidEmail(candidate, options);
+    return analyze.isValidEmail(value, options);
     // Ignore the `else` branch as we test in places that have WebAssembly.
     /* node:coverage ignore next 4 */
   }
@@ -141,7 +165,19 @@ export async function isValidEmail(
   return { blocked: [], validity: "valid" };
 }
 
-// TODO(@wooorm-arcjet): docs.
+/**
+ * Detect whether a request is by a bot.
+ *
+ * @param context
+ *   Context.
+ * @param request
+ *   Request.
+ * @param options
+ *   Configuration.
+ * @returns
+ *   Promise to a result.
+ */
+/// TODO(@wooorm-arcjet): expose `BotEntity`, `BotResult`.
 export async function detectBot(
   context: AnalyzeContext,
   request: AnalyzeRequest,
@@ -161,12 +197,30 @@ export async function detectBot(
   return { allowed: [], denied: [], spoofed: false, verified: false };
 }
 
-// TODO(@wooorm-arcjet): docs.
+/**
+ * Detect sensitive info in a value.
+ *
+ * @param context
+ *   Context.
+ * @param value
+ *   Value.
+ * @param entities
+ *   Entities to detect.
+ * @param contextWindowSize
+ *   Number of tokens to pass to `detect`.
+ * @param detect
+ *   Function to detect sensitive info (optional).
+ * @returns
+ *   Promise to a result.
+ */
+// TODO(@wooorm-arcjet): expose `DetectSensitiveInfoFunction`, `SensitiveInfoEntities`, `SensitiveInfoResult`.
+// TODO(@wooorm-arcjet): less parameters, `5` is a lot.
 export async function detectSensitiveInfo(
   context: AnalyzeContext,
-  candidate: string,
+  value: string,
   entities: SensitiveInfoEntities,
   contextWindowSize: number,
+  // TODO(@wooorm-arcjet): allow `null`, `undefined`.
   detect?: DetectSensitiveInfoFunction,
 ): Promise<SensitiveInfoResult> {
   const { log } = context;
@@ -175,7 +229,7 @@ export async function detectSensitiveInfo(
 
   if (typeof analyze !== "undefined") {
     const skipCustomDetect = typeof detect !== "function";
-    return analyze.detectSensitiveInfo(candidate, {
+    return analyze.detectSensitiveInfo(value, {
       entities,
       contextWindowSize,
       skipCustomDetect,
--- a/analyze/index.ts
+++ b/analyze/index.ts
@@ -100,6 +100,9 @@ function createCoreImports(detect?: DetectSensitiveInfoFunction): ImportObject {
  * This considers different things on the `request` based on the passed
  * `context.characteristics`.
  *
+ * See [*Fingerprints* on
+ * `docs.arcjet.com`](https://docs.arcjet.com/fingerprints/) for more info.
+ *
  * @param context
  *   Context.
  * @param request
--- a/analyze/index.ts
+++ b/analyze/index.ts
@@ -1,14 +1,10 @@
 import { initializeWasm } from "@arcjet/analyze-wasm";
 import type {
-  // TODO(@wooorm-arcjet): rename to `DetectBotOptions`.
   BotConfig,
-  // TODO(@wooorm-arcjet): expose, rename to `DetectBotResult`.
   BotResult,
   DetectedSensitiveInfoEntity,
   DetectSensitiveInfoFunction,
-  // TODO(@wooorm-arcjet): rename to `ValidateEmailOptions`.
   EmailValidationConfig,
-  // TODO(@wooorm-arcjet): expose, rename to `ValidateEmailResult`.
   EmailValidationResult,
   SensitiveInfoEntities,
   SensitiveInfoEntity,
@@ -17,13 +13,11 @@ import type {
 } from "@arcjet/analyze-wasm";
 import type { ArcjetLogger } from "@arcjet/protocol";
 
-/// TODO(@wooorm-arcjet): expose, as `Context`.
 interface AnalyzeContext {
   log: ArcjetLogger;
   characteristics: string[];
 }
 
-/// TODO(@wooorm-arcjet): expose, as `Request`.
 type AnalyzeRequest = {
   ip?: string;
   method?: string;
@@ -37,13 +31,9 @@ type AnalyzeRequest = {
 };
 
 export {
-  // TODO(@wooorm-arcjet): add JSDocs to values from WebAssembly.
   type EmailValidationConfig,
-  // TODO(@wooorm-arcjet): add JSDocs to values from WebAssembly.
   type BotConfig,
-  // TODO(@wooorm-arcjet): add JSDocs to values from WebAssembly.
   type SensitiveInfoEntity,
-  // TODO(@wooorm-arcjet): add JSDocs to values from WebAssembly.
   type DetectedSensitiveInfoEntity,
 };
 
@@ -181,7 +171,6 @@ export async function isValidEmail(
  * @returns
  *   Promise for a result.
  */
-/// TODO(@wooorm-arcjet): expose `BotEntity`, `BotResult`.
 export async function detectBot(
   context: AnalyzeContext,
   request: AnalyzeRequest,
@@ -219,14 +208,11 @@ export async function detectBot(
  * @returns
  *   Promise for a result.
  */
-// TODO(@wooorm-arcjet): expose `DetectSensitiveInfoFunction`, `SensitiveInfoEntities`, `SensitiveInfoResult`.
-// TODO(@wooorm-arcjet): less parameters, `5` is a lot.
 export async function detectSensitiveInfo(
   context: AnalyzeContext,
   value: string,
   entities: SensitiveInfoEntities,
   contextWindowSize: number,
-  // TODO(@wooorm-arcjet): allow `null`, `undefined`.
   detect?: DetectSensitiveInfoFunction,
 ): Promise<SensitiveInfoResult> {
   const { log } = context;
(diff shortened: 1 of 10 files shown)


analyze-wasm/README.md
```diff
@@ -35,6 +35,10 @@ npm install -S @arcjet/analyze-wasm
 
 Use [`@arcjet/analyze`][file-analyze] instead.
 
+## API
+
+Use [`@arcjet/analyze`][file-analyze] instead.
```
davidmytton:
```suggestion
This is an internal package. Use [`@arcjet/analyze`][file-analyze] instead.
```

Just to be specific in case someone comes across it.

analyze/README.md
```diff
@@ -55,6 +58,240 @@ console.log(result);
 // => { blocked: [], validity: "valid" }
 ```
 
+## API
+
+This package exports the identifiers [`detectBot`][api-detect-bot],
```
davidmytton: Add a note to say this is an internal package and shouldn't be used directly.

body/README.md
```diff
@@ -37,6 +39,64 @@ const body = await readBody(fs.createReadStream("example.ts"), { limit: 1024 });
 console.log(body);
 ```
 
+## API
+
+This package exports the identifier
```
davidmytton: Add a note to say this is an internal package and shouldn't be used directly.

I won't add this on every one, but let's do it everywhere these internal docs are exposed. It's unlikely that someone browsing will come across them - more likely they get picked up via search or LLM.

wooorm-arcjet:
@davidmytton I retconned this branch/PR to remove everything relating to readmes (which I have locally and can land after). Now this is only about JSDocs of exposed symbols, and relates to GH-3327

wooorm-arcjet:
@qw-in would it make sense to have you/DX as an “allowed :+1:”er on this repo for such docs related work?

analyze/index.ts
```diff
@@ -92,13 +98,20 @@ function createCoreImports(detect?: DetectSensitiveInfoFunction): ImportObject {
...
- * @returns A SHA-256 string fingerprint.
+ * Generate a fingerprint.
+ *
+ * Fingerprints can be used to identify the client across multiple requests.
+ *
+ * This considers different things on the `request` based on the passed
+ * `context.characteristics`.
+ *
+ * @param context
+ *   Context.
```
davidmytton, line 107: Does this mean that it will link to the new interface name when you rename it per your TODO?

analyze/index.ts
```diff
@@ -121,18 +134,29 @@ export async function generateFingerprint(
...
+ *   Value.
+ * @param options
+ *   Configuration.
+ * @returns
+ *   Promise to a result.
+ */
 export async function isValidEmail(
   context: AnalyzeContext,
-  candidate: string,
+  value: string,
```
davidmytton, line 148: What's the reasoning for renaming this?

duration/index.ts
```diff
@@ -68,87 +68,91 @@ function leadingInt(s: string): [number, string] {
...
+ * If a number is passed it is validated and returned.
  *
- * If a string is provided to the function, it must be in the form of digits
- * followed by a unit. Supported units are `s` (seconds), `m` (minutes), `h`
- * (hours), and `d` (days).
- *
- * @param s The value to parse into seconds.
- * @returns A number representing seconds parsed from the provided duration.
+ * If a string is passed it must be in the form of digits followed by a unit.
+ * Supported units are `s` (seconds),
```
davidmytton, line 77: Why are these wrapped?

headers/index.ts
```diff
@@ -3,21 +3,17 @@ function isIterable(val: any): val is Iterable<any> {
 }
 
 /**
- * This Fetch API interface allows you to perform various actions on HTTP
```
davidmytton, line 6: Has this detail been lost?

inspect/index.ts
```diff
@@ -44,57 +44,30 @@ function isActive(
...
- *
- * Note that spoofed bot detection is not available on free plans.
- *
- * @param {ArcjetRuleResult} result - The rule result to inspect.
- * @returns `true` if the bot rule result was not `"DRY_RUN"` and a spoofed bot
- * was detected, `false` if the bot rule result was not `"DRY_RUN"` and a
- * spoofed bot was not detected, or `undefined` if the rule result was from a
- * `"DRY_RUN"` bot rule or a non-bot rule.
- *
- * @example
```
davidmytton, line 63: Why has this example been removed?

davidmytton (commented):
I've taken a look through. Would like @arcjet-rei to take a more detailed look.

analyze/index.ts
wooorm-arcjet: No.

For:

```
 * @param $1
 *   $2
 ```
 
`$1` has to be the name of a parameter (you don’t have to order these `@params` the same way as below, but it makes sense. So, they are matched together based on `$1`. This adds `$2` and the type defined below to the same symbol.
In JSDoc in JS, you’d need a type: `@param {$type} $1 $2`. But in TS it is not used.

`$2` is just plain text (TS and most places support markdown + also JSDoc inline things such as `@link`.
It’s a bit different between tools if a `-` is supported, so `$1 - $2`. But whitespace is always supported in my experience.
`@param $1` is also often long:

```
 * @param someVeryLongSymbol The
 *   docs are this and that.
```

Which is why I prefer a break:

```
 * @param someVeryLongSymbol
 *   The docs are this and that.
```

Back to the Q: No, renaming the type itself does not affect these.
But renaming parameters does!

File: analyze/index.ts (hunk 5 above)
Comment (wooorm-arcjet) [reply analyze/index.ts:148]: Perhaps you/y’all have more context for `candidate`! For me it the word is a bit out of the blue!

I choose value because this is the value that is being validated. Kinda like `x` in math, or `index` in a loop.

Same with how I applied it to redacting sensitive info, or `duration` below. There’s some idea I think that people have of passing some main thing to work on, and for me `value` captures that “primaryness”

File: duration/index.ts (hunk 6 above)
Comment (wooorm-arcjet) [reply duration/index.ts:77]: I automatically write in https://sembr.org/ I guess! I am not extremely strict about it tho.

File: analyze/index.ts (hunk 5 above)
Comment (davidmytton) [reply analyze/index.ts:148]: Fine by me to rename it to be more generic.

File: duration/index.ts (hunk 6 above)
Comment (davidmytton) [reply duration/index.ts:77]: I usually let the linter/formatter handle wrapping

File: headers/index.ts (hunk 7 above)
Comment (wooorm-arcjet) [reply headers/index.ts:6]: Yeah, it seems like a reiteration of the concept headers, which is also linked below at https://developer.mozilla.org/en-US/docs/Web/API/Headers. I don’t think it’s supposed to be in *our* docs of a small wrapper of the actual `Headers`.
I think what should be in our docs is whether someone needs `ArcjetHeaders` vs `Headers`!

File: inspect/index.ts (hunk 8 above)
Comment (wooorm-arcjet) [reply inspect/index.ts:63]: Because a more up to date version is on different places in the website (I linked 2).

I do think it’s very important to have that example, and more, but I am not so convinced they should be in the code (or in the readme), where it’s more easy to loose track of it (as happened), and less “correct” (this is Next-specific whereas the website can support different integrations/package managers/etc).

What are your thoughts?
Do you know of users who have enjoyed these fuller examples in their IDEs, attached to the code?

File: duration/index.ts (hunk 6 above)
Comment (wooorm-arcjet) [reply duration/index.ts:77]: I don’t think the formatters here touch this? Trunk doesn’t touch most of what I did here?

And, this is a bit more of a semantic thing. Like whether to use `<h1>` or `<p>`. Perhaps AIs can do that better now but it’s more of a conscious thing for me than just line wrapping at 80

File: inspect/index.ts (hunk 8 above)
Comment (davidmytton) [reply inspect/index.ts:63]: I added these examples a while back because 1) it provides a usage example in-context i.e. within the editor; 2) it helps LLMs reading the library docs. I use in-line examples all the time and find them very useful when they're there. I only did Next.js to get started, but I'd like to have them for every adapter.

It's up to us (@qw-in) to keep them updated with the docs, but I think it's an important DX improvement to keep them there.

Comment (wooorm-arcjet) [reply inspect/index.ts:63]: I do think examples are good in adapters, but this is `@arcjet/inspect`, so it felt a bit weird to have a full Next-specific example there.

Comment (davidmytton) [reply inspect/index.ts:63]: Next.js and Node.js are the most popular adapters, so let's have examples for those and then link to the docs for the others.

Comment (wooorm-arcjet) [reply inspect/index.ts:63]: @davidmytton To clarify, do you mean both examples in `@arcjet/inspect`?
Or, do you mean to use `@example`s in `@arcjet/next` and `@arcjet/node` for now?

Comment (davidmytton) [reply inspect/index.ts:63]: Both examples in `@arcjet/inspect`. It's a generic package so I think we should have examples for the most popular options. Most of the others are framework/language specific so we only need one example there.

Comment (wooorm-arcjet) [reply inspect/index.ts:63]: Done, added 6 examples!

I do want to stress that the existing examples were not working: there was one parameter `request`, but `res` was used for the response in the code. Plus, a `Request` type was used used without importing it, presumably `NextApiRequest`.

Just mentioning that it is hard to keep these correct!

File: analyze/index.ts [hunk 9]
```diff
@@ -92,13 +98,20 @@ function createCoreImports(detect?: DetectSensitiveInfoFunction): ImportObject {
...
+ *
+ * This considers different things on the `request` based on the passed
+ * `context.characteristics`.
+ *
+ * @param context
+ *   Context.
+ * @param request
+ *   Request.
+ * @returns
+ *   Promise to a SHA-256 fingerprint.
```
arcjet-rei: Making the English sound more natural:
```suggestion
 *   Promise for a SHA-256 fingerprint.
```

analyze/index.ts
arcjet-rei, line 107: I think "Request to be fingerprinted" is redundant but I think removing "Arcjet Analyze" drops valuable… context.

analyze/index.ts
```diff
@@ -141,7 +165,19 @@ export async function isValidEmail(
...
-// TODO(@wooorm-arcjet): docs.
+/**
+ * Detect whether a request is by a bot.
+ *
+ * @param context
+ *   Context.
+ * @param request
+ *   Request.
+ * @param options
+ *   Configuration.
```
arcjet-rei, line 173: Either document the supported options or drop a TODO to indicate you'll be filling that out later.

analyze/index.ts
```diff
@@ -161,12 +197,30 @@ export async function detectBot(
...
-// TODO(@wooorm-arcjet): docs.
+/**
+ * Detect sensitive info in a value.
+ *
+ * @param context
+ *   Context.
+ * @param value
+ *   Value.
+ * @param entities
+ *   Entities to detect.
```
arcjet-rei: This is inscrutable, and a couple instances of example parameters could help a lot here.

analyze/index.ts
```diff
@@ -161,12 +197,30 @@ export async function detectBot(
...
+ *   Function to detect sensitive info (optional).
+ * @returns
+ *   Promise to a result.
+ */
+// TODO(@wooorm-arcjet): expose `DetectSensitiveInfoFunction`, `SensitiveInfoEntities`, `SensitiveInfoResult`.
+// TODO(@wooorm-arcjet): less parameters, `5` is a lot.
 export async function detectSensitiveInfo(
   context: AnalyzeContext,
-  candidate: string,
+  value: string,
```
arcjet-rei, line 216: Why the switch away from "candidate" to the much more generic "value"? If you want to genericize the name, you should provide more clues what "value" is / how it is meant to be used in the JSDocs above.

body/index.ts
```diff
@@ -46,19 +63,30 @@ export interface ReadableStreamLike {
...
 
+/**
+ * Read the body of a stream.
+ *
+ * @param stream
+ *   Stream.
+ * @param options
+ *   Configuration (required).
+ * @returns
+ *   Promise to a concatenated body.
```
arcjet-rei, line 72: As above.
```suggestion
 *   Promise for a concatenated body.
```

decorate/index.ts
```diff
@@ -6,48 +6,40 @@ import {
...
-}
+// TODO(@wooorm-arcjet): rename to `Decorable`.
+/**
+ * Decorable value.
+ *
+ * Something that looks like `Headers` (Fetch),
+ * `OutgoingMessage` (Node.js), or
+ * `Response` (Fetch).
+ */
+export type ArcjetCanDecorate = HeaderLike | OutgoingMessageLike | ResponseLike;
```
arcjet-rei: This and a couple other things I'll flag later are more than simple documentation changes and it would be good to disentangle these interface cleanups from the documentation changes to simplify reviews.

decorate/index.ts
```diff
@@ -65,7 +57,7 @@ function isHeaderLike(value: ArcjetCanDecorate): value is HeaderLike {
 }
 
 function isResponseLike(value: ArcjetCanDecorate): value is ResponseLike {
-  if (typeof value.headers === "undefined") {
+  if (!("headers" in value) || typeof value.headers === "undefined") {
```
arcjet-rei: What's the reason for this belt-and-suspenders check? I can't think of a real-world situation where we'd need to check for both of these cases, which makes this feel like overly-restrictive typing to me.

decorate/index.ts
```diff
@@ -75,7 +67,10 @@ function isResponseLike(value: ArcjetCanDecorate): value is ResponseLike {
 function isOutgoingMessageLike(
   response: ArcjetCanDecorate,
 ): response is OutgoingMessageLike {
-  if (typeof response.headersSent !== "boolean") {
+  if (
+    !("headersSent" in response) ||
+    typeof response.headersSent !== "boolean"
+  ) {
```
arcjet-rei: Same question as above.

decorate/index.ts
```diff
@@ -159,14 +154,19 @@ function nearestLimit(
 }
 
 /**
- * Decorates an object with `RateLimit` and `RateLimit-Policy` headers based
- * on an {@link ArcjetDecision} and conforming to the [Rate Limit fields for
+ * Decorate something based on an Arcjet decision with rate limit headers.
```
arcjet-rei, line 157: Why drop the explicit link to `ArcjetDeicision`?

decorate/index.ts
```diff
@@ -159,14 +154,19 @@ function nearestLimit(
...
  * HTTP](https://ietf-wg-httpapi.github.io/ratelimit-headers/draft-ietf-httpapi-ratelimit-headers.html)
  * draft specification.
  *
- * @param value The object to decorate—must be similar to {@link Headers}, {@link Response} or
- * {@link OutgoingMessage}.
- * @param decision The {@link ArcjetDecision} that was made by calling `protect()` on the SDK.
+ * @param value
+ *   Decorable value.
+ * @param decision
+ *   Decision from `protect()`.
```
arcjet-rei, line 167: The old text is more informative than the new. Was it inaccurate? Why get rid of the extra information?

inspect/index.ts
```diff
@@ -44,57 +44,110 @@ function isActive(
...
+ * See [*Bot protection* on
+ * `docs.arcjet.com`](https://docs.arcjet.com/bot-protection/quick-start)
+ * for more info.
+ *
+ * Spoofed bot detection is part of advanced bot protection features which
+ * are not available on free plans but are available on the starter and
+ * business plans.
+ * See [*Bot verification* on
+ * `docs.arcjet.com`](https://docs.arcjet.com/bot-protection/reference#bot-verification)
+ * for more info.
```
arcjet-rei, line 63: Segmented / subscription-gated features should be called out as such up top. The old language also had the benefit of being clear and concise. I'll defer to @qw-in and / or @davidmytton as to what the best language to use here, but the current wording feels verbose and murky compared to the original.

ip/index.ts
```diff
@@ -76,7 +76,7 @@ function isTrustedProxy(
   return false;
 }
 
-abstract class CIDR {
+abstract class Cidr {
```
arcjet-rei: Cf. ref. our earlier discussion about changing capitalization, which seems especially pertinent here where it results in `class IPv4CDR extends Cidr`, which feels like an almost purposefully awkward inconsistency.

ip/index.ts
```diff
@@ -201,13 +201,19 @@ function isCIDR(address: string): address is `${string}/${string}` {
...
+ * Parse CIDR addresses and keep non-CIDR IP addresses.
+ *
+ * @param value
+ *   Value to parse.
+ * @returns
+ *   Parsed CIDR or given `value`.
+ */
+export function parseProxy(value: string): Cidr | string {
+  if (isCIDR(value)) {
+    return parseCIDR(value);
```
arcjet-rei: As above, the inconsistency feels very awkward.

arcjet-rei (approved):
When I see a PR entitled `docs: add JSDocs`, my expectation is that it will only include documentation changes, and not renamings and other refactorings, and especially not semantically meaningful (even if minor) changes. I'm approving the PR as-is because I think it is a substantial improvement, but please keep working on focusing PRs to make fewer different kinds of changes in each changeset.

Also, a consistent feeling I was left with was the more streamlined and genericized the documentation is, the less value it offers readers. I think the benefits for AI that @davidmytton describes also pertain to human readers as well. I know that when I'm hacking on a project or otherwise consuming a new dependency, it is very frustrating when I have to chase through a bunch of links to find the single source of truth specifying (for example) the configuration options for a function call. A little redundancy can go a long way in documentation.

For the most part, it feels like the changes to existing documentation are a (significant) net improvement, but I think in places you've simplified things too much. We'll just keep iterating until we get it right, I guess!

analyze/index.ts
```diff
@@ -121,18 +134,29 @@ export async function generateFingerprint(
...
+ * Check whether an email is valid.
+ *
+ * @param context
+ *   Context.
+ * @param value
+ *   Value.
+ * @param options
+ *   Configuration.
+ * @returns
+ *   Promise to a result.
```
arcjet-rei: Same as above:
```suggestion
 *   Promise for a result.
```

analyze/index.ts
wooorm-arcjet: I believe that `BotOptions` should have JSDocs but those come from WebAssembly where it is generated without docs. I added todos to do that.

analyze/index.ts
wooorm-arcjet: The type is called `ArcjetContext`, so TypeScript and other tools will turn this into something like `context: ArcjetContext - Context`.

I think there are improvements that can be made there, but I personally don’t think another `Arcjet` is one such improvement. I will do that if you want that.

analyze/index.ts
wooorm-arcjet: It is a complex API. It is hard to explain without looking at the type associated to `entities`. Then again, nobody is supposed to use this except for us. Anyway, I used more words. Let me know what you think and how you want to improve it.

analyze/index.ts
wooorm-arcjet: I think that is more clear with a function called `detectSensitiveInfo` with a `value: string` compared to a `candidate: string`, which is why I changed it.
I’d think that someone wants to detect sensitive info in some primary value of a string like shape.
Candidate confuses me.

I put that in the function description: “Detect sensitive info in a value.”

If that’s me and you think `candidate` is better I will change it.

decorate/index.ts
wooorm-arcjet: For TypeScript. But moved into https://github.com/arcjet/arcjet-js/pull/4865

decorate/index.ts
wooorm-arcjet: moved into https://github.com/arcjet/arcjet-js/pull/4865

decorate/index.ts
wooorm-arcjet: Moved into https://github.com/arcjet/arcjet-js/pull/4865

decorate/index.ts
wooorm-arcjet: This section is about the function, why would you use it, what it does.

The types of the parameters are of course very important but available in a specific section next to it (displayed differently depending on whether this will be read in an IDE or a readme)

decorate/index.ts
wooorm-arcjet: It reiterates the types, and readmes / IDEs / etc will display those types

inspect/index.ts
wooorm-arcjet: 🤔 It was a small `note:` deep down and now it is a giant section high up, saying which plans this does and does not work on. So... I feel like I agree with your points and tried to do just that.
So I’m curious to hear what you, @qw-in, @davidmytton suggest

ip/index.ts
wooorm-arcjet: rebased!

ip/index.ts
wooorm-arcjet: no longer!

analyze/index.ts
wooorm-arcjet: done

inspect/index.ts
davidmytton: Seems fine to me.

wooorm-arcjet:
@davidmytton you’re still marked as needing changes!

analyze/index.ts
```diff
@@ -92,13 +102,20 @@ function createCoreImports(detect?: DetectSensitiveInfoFunction): ImportObject {
...
- * across multiple requests.
- * @param context - The Arcjet Analyze context.
- * @param request - The request to fingerprint.
- * @returns A SHA-256 string fingerprint.
+ * Generate a fingerprint.
+ *
+ * Fingerprints can be used to identify the client across multiple requests.
+ *
+ * This considers different things on the `request` based on the passed
+ * `context.characteristics`.
```
qw-in, line 101: Link https://docs.arcjet.com/fingerprints/?

body/index.ts
```diff
@@ -1,18 +1,35 @@
-export type ReadBodyOpts = {
-  limit: number;
+// TODO(@wooorm-arcjet): rename to `Options`.
+/**
+ * Configuration.
+ */
+export interface ReadBodyOpts {
```
qw-in: Lets omit this type => interface change and focus on jsdoc if that's ok

env/index.ts
```diff
@@ -1,41 +1,111 @@
-export type Env = {
+/**
+ * Environment.
+ */
+// TODO(@wooorm-arcjet): rename to `Environment`.
+export interface Env {
```
qw-in: Lets focus on the jsdoc for this PR

wooorm-arcjet: it is now again a type

body/index.ts
wooorm-arcjet: it is now again a type

analyze/index.ts
wooorm-arcjet: linked

wooorm-arcjet:
@qw-in all done, thanks for the review!

wooorm-arcjet:
@arcjet-rei I see you gave a ✅ with comments. I responded to the comments. Do you still have further questions? Shall I mark them all as “resolved”?

analyze/index.ts
arcjet-rei: That makes sense, although I still think "value" is a little _too_ generic. Not enough to block it (or come up with something better off the top of my head), but I'll keep thinking about it.

arcjet-rei:
I've responded to and / or resolved all of my outstanding comments!

Issue #4553, opened 2025-07-03
Revisit examples that set top-level characteristics


blaine-arcjet:
Prior to the segmented cache, we had to specify `characteristics` in the top-level SDK options. This is no longer the case and they should be specified on an individual rule where possible.

We need to revisit the examples and docs to remove the workarounds.

wooorm-arcjet:
Noting here too that top-level characteristics are still used by the shield rule: GH-4430.

Issue #4430, opened 2025-06-16
Shield rules should accept characteristics


blaine-arcjet:
Currently, shield rules only consider the root characteristics on the SDK to generate a fingerprint. We need to change the backend to support these "local" characteristics then adjust the SDK to attach them to the protobuf (the same as how Rate Limit uses "local" but defaults to "global" characteristics).

PR #4753, opened 2025-07-29
refactor(examples): move `characteristics` to rules

wooorm-arcjet:
To solve GH-4553, I looked for `characteristics`.
I moved every case to a rule, when that type-checked. If there was no rule used that accepted them,
I assumed that they were not needed,
and removed them.

I only found top-level use inside the examples,
so I think this closes GH-4553.

### DIFF
--- a/examples/nextjs-14-nextauth-4/app/api/arcjet/route.ts
+++ b/examples/nextjs-14-nextauth-4/app/api/arcjet/route.ts
@@ -18,9 +18,6 @@ const authOptions = {
 // The arcjet instance is created outside of the handler
 const aj = arcjet({
   key: process.env.ARCJET_KEY!, // Get your site key from https://app.arcjet.com
-  // We specify a custom fingerprint so we can dynamically build it within each
-  // demo route.
-  characteristics: ["fingerprint"],
   rules: [
     // Protect against common attacks with Arcjet Shield
     shield({
@@ -40,14 +37,6 @@ export async function GET(req: NextRequest, res: Response) {
 
   let decision: ArcjetDecision;
   if (session && session.user && session.user!.email) {
-    // A very simple hash to avoid sending PII to Arcjet. You may wish to add a
-  // unique salt prefix to protect against reverse lookups.
-  const email = session.user!.email;
-  const emailHash = require("crypto")
-    .createHash("sha256")
-    .update(email)
-    .digest("hex");
-
     // Allow higher limits for signed in users.
     const rl = aj.withRule(
       // Create a token bucket rate limit. Other algorithms are supported.
@@ -59,10 +48,8 @@ export async function GET(req: NextRequest, res: Response) {
       })
     );
 
-    const fingerprint = emailHash; // Use the email hash as the fingerprint
-
     // Deduct 5 tokens from the token bucket
-    decision = await rl.protect(req, { fingerprint, requested: 5 });
+    decision = await rl.protect(req, { requested: 5 });
     console.log("Arcjet logged in decision", decision)
   } else {
     const fingerprint = ip(req);
@@ -79,7 +66,7 @@ export async function GET(req: NextRequest, res: Response) {
     );
 
     // Deduct 5 tokens from the token bucket
-    decision = await rl.protect(req, { fingerprint, requested: 5 });
+    decision = await rl.protect(req, { requested: 5 });
     console.log("Arcjet logged out decision", decision)
   }
 
--- a/examples/nextjs-app-dir-rate-limit/app/api/arcjet/route.ts
+++ b/examples/nextjs-app-dir-rate-limit/app/api/arcjet/route.ts
@@ -6,12 +6,12 @@ const aj = arcjet({
   // and set it as an environment variable rather than hard coding.
   // See: https://nextjs.org/docs/app/building-your-application/configuring/environment-variables
   key: process.env.ARCJET_KEY!,
-  // Limiting by ip.src is the default if not specified
-  characteristics: ["ip.src"],
   rules: [
     // Fixed window rate limit. Arcjet also supports sliding window and token
     // bucket.
     fixedWindow({
+      // Limiting by ip.src is the default if not specified
+      characteristics: ["ip.src"],
       mode: "LIVE", // will block requests. Use "DRY_RUN" to log only      
       window: "1m", // 1 min fixed window
       max: 1, // allow a single request (for demo purposes)
--- a/examples/nextjs-authjs-5/app/api/protected/route.ts
+++ b/examples/nextjs-authjs-5/app/api/protected/route.ts
@@ -4,7 +4,6 @@ import { auth } from "auth";
 // The arcjet instance is created outside of the handler
 const aj = arcjet({
   key: process.env.ARCJET_KEY!, // Get your site key from https://app.arcjet.com
-  characteristics: ["userId"], // Rate limit based on the Clerk userId
   rules: [
     // Protect against common attacks with Arcjet Shield
     shield({
@@ -12,6 +11,7 @@ const aj = arcjet({
     }),
     // Create a token bucket rate limit. Other algorithms are supported.
     tokenBucket({
+      characteristics: ["userId"], // Rate limit based on the Clerk userId
       mode: "LIVE", // will block requests. Use "DRY_RUN" to log only
       refillRate: 5, // refill 5 tokens per interval
       interval: 10, // refill every 10 seconds
--- a/examples/nextjs-authjs-5/middleware.ts
+++ b/examples/nextjs-authjs-5/middleware.ts
@@ -3,7 +3,6 @@ import { auth } from "auth";
 
 const aj = arcjet({
   key: process.env.ARCJET_KEY!, // Get your site key from https://app.arcjet.com
-  characteristics: ["userId"], // track requests by a custom user ID
   rules: [
     // Protect against common attacks with Arcjet Shield
     shield({
@@ -11,6 +10,7 @@ const aj = arcjet({
     }),
     // Create a token bucket rate limit. Other algorithms are supported.
     tokenBucket({
+      characteristics: ["userId"], // track requests by a custom user ID
       mode: "LIVE", // will block requests. Use "DRY_RUN" to log only
       refillRate: 5, // refill 5 tokens per interval
       interval: 10, // refill every 10 seconds
--- a/examples/nextjs-better-auth/app/api/auth/[...all]/route.ts
+++ b/examples/nextjs-better-auth/app/api/auth/[...all]/route.ts
@@ -17,7 +17,6 @@ import { NextRequest } from "next/server";
 // The arcjet instance is created outside of the handler
 const aj = arcjet({
   key: process.env.ARCJET_KEY, // Get your site key from https://app.arcjet.com
-  characteristics: ["userId"],
   rules: [
     // Protect against common attacks with Arcjet Shield. Other rules are
     // added dynamically using `withRule`.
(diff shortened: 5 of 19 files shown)


davidmytton:
Will have @qw-in review this (added to the JS SDK team to allow docs and examples reviews).

examples/nextjs-better-auth/app/api/auth/[...all]/route.ts
```diff
@@ -60,16 +59,6 @@ async function protect(req: NextRequest): Promise<ArcjetDecision> {
...
-  // If the user is logged in we'll use their ID as the identifier. This
-  // allows limits to be applied across all devices and sessions (you could
-  // also use the session ID). Otherwise, fall back to the IP address.
-  let userId: string;
-  if (session?.user.id) {
-    userId = session.user.id;
-  } else {
-    userId = ip(req) || "127.0.0.1"; // Fall back to local IP if none
-  }
-
```
Comment (qw-in) [inline examples/nextjs-better-auth/app/api/auth/[...all]/route.ts]: Seems like a shame to lose this functionality

wooorm-arcjet:
🤔 It could be that these still only accept global characteristics? I just saw https://github.com/arcjet/arcjet-js/issues/4430. I didn’t know that. Do you know if GH-4430 is correct and thus blocks GH-4553/this PR?

qw-in:
@wooorm-arcjet that sounds correct, though I think this PR can mostly land minus a few examples

examples/nextjs-14-nextauth-4/app/api/arcjet/route.ts
```diff
@@ -59,10 +48,8 @@ export async function GET(req: NextRequest, res: Response) {
...
     );
 
-    const fingerprint = emailHash; // Use the email hash as the fingerprint
-
     // Deduct 5 tokens from the token bucket
-    decision = await rl.protect(req, { fingerprint, requested: 5 });
+    decision = await rl.protect(req, { requested: 5 });
     console.log("Arcjet logged in decision", decision)
   } else {
     const fingerprint = ip(req);
```
blaine-arcjet, line 68:
```suggestion
```
Probably can remove an import too.

examples/nextjs-14-nextauth-4/app/api/arcjet/route.ts
```diff
@@ -59,10 +48,8 @@ export async function GET(req: NextRequest, res: Response) {
       })
     );
 
-    const fingerprint = emailHash; // Use the email hash as the fingerprint
-
     // Deduct 5 tokens from the token bucket
-    decision = await rl.protect(req, { fingerprint, requested: 5 });
```
blaine-arcjet, line 65: This should still take an `emailHash` fingerprint because the purpose is to demonstrate how you can rate limit based on an email to apply the token bucket across different IP sessions.

examples/nextjs-app-dir-rate-limit/app/api/arcjet/route.ts
```diff
@@ -6,12 +6,12 @@ const aj = arcjet({
...
   // See: https://nextjs.org/docs/app/building-your-application/configuring/environment-variables
   key: process.env.ARCJET_KEY!,
-  // Limiting by ip.src is the default if not specified
-  characteristics: ["ip.src"],
   rules: [
     // Fixed window rate limit. Arcjet also supports sliding window and token
     // bucket.
     fixedWindow({
+      // Limiting by ip.src is the default if not specified
+      characteristics: ["ip.src"],
```
blaine-arcjet: this is the default so it can be commented

examples/nextjs-authjs-5/app/api/protected/route.ts
```diff
@@ -4,14 +4,14 @@ import { auth } from "auth";
...
 const aj = arcjet({
   key: process.env.ARCJET_KEY!, // Get your site key from https://app.arcjet.com
-  characteristics: ["userId"], // Rate limit based on the Clerk userId
   rules: [
     // Protect against common attacks with Arcjet Shield
     shield({
       mode: "LIVE", // will block requests. Use "DRY_RUN" to log only
     }),
     // Create a token bucket rate limit. Other algorithms are supported.
     tokenBucket({
```
blaine-arcjet: It is probably better to show this as 2 different rules because the handler seems to be looking for a `req.auth.user.id` or falling back to an email. The caching should be segmented based on the fingerprint you are generating.

examples/nextjs-authjs-5/middleware.ts
```diff
@@ -3,14 +3,14 @@ import { auth } from "auth";
...
   key: process.env.ARCJET_KEY!, // Get your site key from https://app.arcjet.com
-  characteristics: ["userId"], // track requests by a custom user ID
   rules: [
     // Protect against common attacks with Arcjet Shield
     shield({
       mode: "LIVE", // will block requests. Use "DRY_RUN" to log only
     }),
     // Create a token bucket rate limit. Other algorithms are supported.
     tokenBucket({
+      characteristics: ["userId"], // track requests by a custom user ID
```
blaine-arcjet: Same as above

examples/nextjs-better-auth/app/api/auth/[...all]/route.ts
Comment (blaine-arcjet) [reply examples/nextjs-better-auth/app/api/auth/[...all]/route.ts]: @qw-in is correct. This should use `withRule` to apply the correct characteristic for said rule.

examples/nextjs-clerk-rate-limit/app/api/arcjet/route.ts
```diff
@@ -6,9 +6,6 @@ import ip from "@arcjet/ip";
 // The root Arcjet client is created outside of the handler.
 const aj = arcjet({
   key: process.env.ARCJET_KEY!, // Get your site key from https://app.arcjet.com
-  // We specify a custom fingerprint so we can dynamically build it within each
-  // demo route.
-  characteristics: ["fingerprint"],
```
blaine-arcjet: See other comments, these should be applied at the `withRule` correctly.

examples/nextjs-decorate/app/api-app/arcjet/route.ts
```diff
@@ -17,6 +15,8 @@ const aj = arcjet({
     // Fixed window rate limit. Arcjet also supports sliding window and token
     // bucket.
     fixedWindow({
+      // Limiting by ip.src is the default if not specified
+      characteristics: ["ip.src"],
```
blaine-arcjet: default so commented

examples/nextjs-decorate/pages/api/arcjet.ts
```diff
@@ -7,12 +7,12 @@ const aj = arcjet({
...
   // See: https://nextjs.org/docs/app/building-your-application/configuring/environment-variables
   key: process.env.ARCJET_KEY!,
-  // Limiting by ip.src is the default if not specified
-  //characteristics: ["ip.src"],
   rules: [
     // Fixed window rate limit. Arcjet also supports sliding window and token
     // bucket.
     fixedWindow({
+      // Limiting by ip.src is the default if not specified
+      characteristics: ["ip.src"],
```
blaine-arcjet: default so commented.

examples/nextjs-openai/app/api/chat/route.ts
```diff
@@ -25,12 +25,12 @@ const aj = arcjet({
   // and set it as an environment variable rather than hard coding.
   // See: https://nextjs.org/docs/app/building-your-application/configuring/environment-variables
   key: process.env.ARCJET_KEY,
-  characteristics: ["ip.src"], // track requests by IP address
   rules: [
     shield({
       mode: "LIVE", // will block requests. Use "DRY_RUN" to log only
     }),
     tokenBucket({
+      characteristics: ["ip.src"], // track requests by IP address
```
blaine-arcjet: default so remove?

examples/nextjs-permit/src/app/api/permissions/route.ts
```diff
@@ -17,7 +17,7 @@ export async function GET(req: Request) {
   }
 
   // Request a decision from Arcjet with the user's ID as a fingerprint
-  const decision = await aj.protect(req, { fingerprint: user.id });
+  const decision = await aj.protect(req);
```
blaine-arcjet: Still want to apply user.id as fingerprint. Can be defined on `slidingWindow` above if only applied to this handler.

examples/nextjs-permit/src/app/api/stats/route.ts
```diff
@@ -39,14 +39,9 @@ async function getClient() {
 }
 
 export async function GET(req: NextRequest) {
-  // Get the user's ID if they are logged in, otherwise use
-  // their IP address as a fingerprint
-  const user = await currentUser();
-  const fingerprint: string = user ? user.id : ip(req);
```
blaine-arcjet: `withRule` to fork the client.

examples/nextjs-server-actions/app/actions.ts
```diff
@@ -4,8 +4,6 @@ import arcjet, { request, validateEmail } from "@arcjet/next";
 
 const aj = arcjet({
   key: process.env.ARCJET_KEY!,
-  // Use the `uid` cookie that is set by the middleware to fingerprint requests
-  characteristics: ['http.request.cookie["uid"]'],
```
blaine-arcjet, line 8: This is required to make the example work. I think you need to be checking these changes end-to-end.

examples/nextjs-better-auth/app/api/auth/[...all]/route.ts
Comment (qw-in) [reply examples/nextjs-better-auth/app/api/auth/[...all]/route.ts]: May find this helpful: https://docs.arcjet.com/fingerprints/#ip-address--user-id

qw-in:
@wooorm-arcjet let me know if I can help with getting this one landed! Happy to take on making the changes if you'd like

wooorm-arcjet:
Thanks! I missed GH-4430. So half of this PR is incorrect.
I was thinking of fixing that first and then fixing this PR. But didn‘t get around to it.
We can also revert half of this PR first.
If you can do either quickly, I’m all :+1: to that!

examples/nextjs-server-actions/app/actions.ts
wooorm-arcjet: Are you sure? I just checked it and “it” worked 🤔
Arcjet only has a `validateEmail` email rule here, which doesn’t accept characteristics?
So, “it”, means checking the input box for valid/invalid emails?
Is `uid` a cookie set by Next?

examples/nextjs-app-dir-rate-limit/app/api/arcjet/route.ts
wooorm-arcjet: done

examples/nextjs-decorate/pages/api/arcjet.ts
wooorm-arcjet: done

examples/nextjs-permit/src/app/api/permissions/route.ts
wooorm-arcjet: Specified it at as a `characteristic` on `slidingWindow` above, and passing this value again, so that the types now pass

wooorm-arcjet:
Changed this PR to no longer touch anything when `shield` is used, as that uses global characteristics.

There are only a few examples where shield is not used.
This PR now passes `characteristics` to specific rules in those examples.
Where the global was `ip.src`, that is now also local, but commented out there.

examples/nextjs-server-actions/app/actions.ts
blaine-arcjet: I've just tested and confirmed that removing this breaks fingerprinting for Next.js server actions when running in production mode.

If you test with `npm run dev`, it will default the IP to `127.0.0.1` and log a message informing it is overridden.

If you test with `npm run build && npm run start`, Arcjet will fail because there is no IP override in production and the IP cannot be accessed in server actions.

wooorm-arcjet: Ah, fingerprints also use characteristics. Good catch.
Reverted, with an explanatory comment.
But that kinda implies that top-level characteristics are still quite useful 🤔

Issue #3350, opened 2025-02-26
SvelteKit adapter cannot access .env variables with Vite 6


blaine-arcjet:
While testing our SvelteKit example with Vite 6, I noticed that the values set in `.env` were not being detected by the SDK.

It seems that Vite 6 must have removed these from `process.env` but I don't think they provided any way to actually get them.

wooorm-arcjet:
I think this is solved (tested on Vite 6.3.something).

1. follow steps in readme of `examples/sveltekit` (including renaming `.env.example` to `.env`
2. add `ARCJET_BLOB=blub` to `.env` too just to be sure
3. add `console.log('xxx:', env)` on L4 of `src/lib/server/arcjet.ts`
4. ```
    xxx: {
      ...
      ARCJET_BLOB: 'blub',
    ```

blaine-arcjet:
I think you are mistaken by the issue. The Arcjet adapter for SvelteKit tries to access environment variables, such as `ARCJET_ENV`, `ARCJET_BASE_URL`, etc within the adapter itself. These are not available to the adapter due to some underlying changes in Vite (I think).

wooorm-arcjet:
My repro above points to something quite interesting.
Note how `env` from `'$env/dynamic/private'` is being logged, which has things from `.env`.
When changing that to `process.env`, there are no things from `.env`.

The problem then is: sveltekit loads `.env` and exposes it through its custom APIs, but doesn’t put stuff into the regular Node `process.env`.

This points to an underlying problem: arcjet core has different ways to pass options to it, regular options of which the API key is one, and options through `process.env` of which base url, development, and logging are examples.
Which becomes a problem when some tools will not do with `process.env` as what Node itself does.

wooorm-arcjet:
It is still unclear which values this issue is about. Can you please clarify what you tried, what worked, and what didn’t?

PR #989, opened 2024-06-17
fix(sveltekit): Load environment variables from process.env

blaine-arcjet:
I can't find the actual part of the SvelteKit code that causes the import of `"$env/dynamic/private"` inside a node_module to fail, but I fixed our example so the problem surfaced.

I also discovered https://github.com/sveltejs/kit/issues/12028 which states that a warning should be added that third-party libraries should not use this import. And https://github.com/nextauthjs/next-auth/issues/9809 + https://github.com/nextauthjs/next-auth/pull/10117 which highlight problems when it is used.

Perhaps we'll need to explore something like https://github.com/nextauthjs/next-auth/discussions/6247 but it seems that `process.env` is populated in SvelteKit so we can use it for now.

Fixes #982

Issue #982, opened 2024-06-15
SvelteKit SDK: Cannot find package '$env' imported from @arcjet/sveltekit/index.js


davidmytton:
Reported in https://github.com/arcjet/arcjet-js/issues/886#issuecomment-2169075274, the Sveltekit adapter is not working with both Node and Bun (which is the runtime of the original report).

```
"@arcjet/sveltekit": "^1.0.0-alpha.16"
```

## Reproduce - Node

```
node --version
v20.14.0
```

Follow the [SvelteKit setup guide](https://kit.svelte.dev/docs/creating-a-project):

```
npm create svelte@latest my-app
cd my-app
npm install
npm run dev
```

Then follow [our SvelteKit getting started guide](https://docs.arcjet.com/get-started/sveltekit) to install `npm i @arcjet/sveltekit` and create an API route at `src/routes/arcjet/+server.ts`

```ts
import { env } from "$env/dynamic/private";
import arcjet, { tokenBucket } from "@arcjet/sveltekit";
import { error, json, type RequestEvent } from "@sveltejs/kit";

const aj = arcjet({
  // Get your site key from https://app.arcjet.com
  // and set it as an environment variable rather than hard coding.
  // See: https://kit.svelte.dev/docs/modules#$env-dynamic-private
  key: env.ARCJET_KEY!,
  rules: [
    // Create a token bucket rate limit. Other algorithms are supported.
    tokenBucket({
      mode: "LIVE", // will block requests. Use "DRY_RUN" to log only
      characteristics: ["userId"], // track requests by a custom user ID
      refillRate: 5, // refill 5 tokens per interval
      interval: 10, // refill every 10 seconds
      capacity: 10, // bucket maximum capacity of 10 tokens
    }),
  ],
});

export async function GET(event: RequestEvent) {
  const userId = "user123"; // Replace with your authenticated user ID
  const decision = await aj.protect(event, { userId, requested: 5 }); // Deduct 5 tokens from the bucket
  console.log("Arcjet decision", decision);

  if (decision.isDenied()) {
    return error(429, { message: "Too many requests" });
  }

  return json({ message: "Hello world" });
}
```

Then `npm run dev` and loading the route errors:

```
npm run dev

> sk@0.0.1 dev
> vite dev

Re-optimizing dependencies because lockfile has changed

  VITE v5.3.1  ready in 418 ms

  ➜  Local:   http://localhost:5173/
  ➜  Network: use --host to expose
[...]

PR #4754, opened 2025-07-29
fix(sveltekit): read from `$env`, not `process.env`

wooorm-arcjet:
This should close GH-3350.
In there I tested that things from `.env` are put into `$env` by the framework but not `process.env`.
There is one case where `process.env` was used in the integration. This commit replaces its use.

### DIFF
--- a/arcjet-sveltekit/package.json
+++ b/arcjet-sveltekit/package.json
@@ -55,6 +55,9 @@
     "@arcjet/transport": "1.0.0-beta.9",
     "arcjet": "1.0.0-beta.9"
   },
+  "peerDependencies": {
+    "svelte": "^3.54.0 || ^4.0.0 || ^5.0.0-0"
+  },
   "devDependencies": {
     "@arcjet/eslint-config": "1.0.0-beta.9",
     "@arcjet/rollup-config": "1.0.0-beta.9",


wooorm-arcjet:
This relates to GH-3706 as well, and should solve one error there, but probably not the others

blaine-arcjet (changes requested):
Last time I looked this is not processed by Vite for dependencies. You should be able to use `git blame` to track the problems around this.

wooorm-arcjet:
You mean https://github.com/arcjet/arcjet-js/commit/375eaa9d41e2aa3e8534cdc826429b3a3f8e6909#diff-072eb728041c40d9204c82bb1940e0e6b885ff3ddacea9bdf88d88280bb5ac93 probably?

In the commit message you note that the example broke. But CI here shows that it seems to work now.

Anyway, then it is intentional that we shouldn’t use `process.env` or any of its variants in a library for SvelteKit.

qw-in:
The sveltekit publishing best practices touch on this I think: https://svelte.dev/docs/kit/packaging#Best-practices

blaine-arcjet:
> In the commit message you note that the example broke. 

In the commit message, I linked 3-4 other issues that state why this doesn't work, won't work, etc.

> But CI here shows that it seems to work now.

What do you mean by "seems to work"? The build succeeding doesn't mean this "works". The environment variable is not actually detected properly when you actually execute the example and run requests through. This can be made very obvious by trying to use `ARCJET_LOG_LEVEL` to increase the logging level to debug and running requests through.

I also vaguely remember there being a difference between examples loading the SDK via `file:../` protocol vs from npm. Vite might do weird things because Vite.

arcjet-rei (dismissed):
Given that even SvelteKit's own documentation is hilariously, tragically vague about what their own code is doing ("In prod, this behavior will depend on your adapter."), we are not landing this until someone has done the necessary digging to establish with certainty that this change is not going to cause unexpected behavior for SDK users. In practice that means that somebody has gone into the adapter source code on the Vite / SvelteKit side and figured out how and in which circumstances the adapters are mutating `process.env`.

wooorm-arcjet:
Earlier code used `$env/dynamic/private` (GH-982, GH-989).
This stopped working inside `node_modules/` because of a bug in Vite.
The code here was swapped to `process.env`. With a note: “we can use it for now.”
Vite 6 apparently stopped loading `.env` into `process.env` (GH-3350).
From what I see they support `VITE_` prefixed values and put them in `import.meta.env`: https://vite.dev/guide/env-and-mode.html#env-variables.

SvelteKit runs on different places, these places have different ways to set environment variables.
`process.env` is not always available.
Sometimes SvelteKit runs prerendered, where not all environment variables should be available.
SvelteKit thus provides its own way to access environment variables.
It puts things from `.env` files into that, not in `process.env`.

`$env/dynamic/private` works. In SvelteKit and in Vite.
This is what I noticed in https://github.com/arcjet/arcjet-js/issues/3350#issuecomment-3012204162 and PRd here.
I have tested this.
I have also tested `$env` in `node_modules/` with the old reproduction of David in GH-982.
It is also documented on the SvelteKit website. Emphasis mine:

> You should avoid using SvelteKit-specific modules like `$app/`environment in your packages ***unless you intend for them to only be consumable by other SvelteKit projects***.
>
> — https://svelte.dev/docs/kit/packaging#Best-practices

`arcjet-sveltekit` ***is*** supposed to only run in Sveltekit projects.

<details><summary>Longer steps to show it works</summary>

By adding onto this PR another patch:

```patch
--- a/arcjet-sveltekit/index.ts
+++ b/arcjet-sveltekit/index.ts
@@ -192,6 +192,8 @@ export default function arcjet<
   Simplify<ExtraProps<Rules> & CharacteristicProps<Characteristics>>
 > {
   const client = options.client ?? createRemoteClient();
+  console.log('sveltekit:node env:', process.env)
+  console.log('sveltekit:$env:', env)
 
   const log = options.log
     ? options.log
```

With a `.env` in `arcjet-js/examples/sveltekit`:

```
ARCJET_ENV=development
ARCJET_KEY=asdasd
ARCJET_LOG_LEVEL=debug
```

Then running:

```sh
npm run build
```

I see:

```
✓ 197 modules transformed.
sveltekit:node env: {
  ... no `ARCJET_`
}
sveltekit:$env: {
  ARCJET_ENV: 'development',
  ARCJET_KEY: 'asdasd',
  ARCJET_LOG_LEVEL: 'debug',
  ...
}
✦Aj WARN Arcjet will use 127.0.0.1 when missing public IP address in development mode
vite v6.3.5 building for production...
```

With `npm run dev`, you will see debug messages.

Testing different adapters, `@sveltejs/adapter-node` and `@sveltejs/adapter-static`:

```patch
+++ b/examples/sveltekit/svelte.config.js
@@ -1,4 +1,4 @@
-import adapter from '@sveltejs/adapter-auto';
+import adapter from '@sveltejs/adapter-node';
 import { vitePreprocess } from '@sveltejs/vite-plugin-svelte';
 import { csp } from '@nosecone/sveltekit';
```

…running `npm run build`/`npm run dev` also prints `ARCJET_` values and shows debug messages.

Additionally, running production.
Setting `ARCJET_ENV=production` in `.env`.
Then running `npm run build; npm run preview`. Curling into `curl -v -H "x-arcjet-suspicious: true" http://localhost:4173` (note the different protocol).
That still prints the `ARCJET_` values *and* debugs and runs in production.

</details>

> > In the commit message you note that the example broke.
>
> In the commit message, I linked 3-4 other issues that state why this doesn't work, won't work, etc.
>
> > But CI here shows that it seems to work now.
>
> What do you mean by "seems to work"?

You linked https://redirect.github.com/sveltejs/kit/issues/12028. One of 800+ open issues. Not by a maintainer. No comments to it.
I believe it is already solved because the docs do talk about it in the link from Quinn https://svelte.dev/docs/kit/packaging#Best-practices.
I also believe that issue has alternative solutions: we could document that a) people must not set `config.kit.env.privatePrefix`, b) support `PRIVATE_` prefixes, c) support arbitrary prefixes with a config value.

The other issues over on `nextauthjs/next-auth` has to do with how to put environment variables into prerendered code, things sent to the client.
They end up skipping those routes in https://redirect.github.com/nextauthjs/next-auth/pull/10339.
We do not test prerendering here.
We could skip things too.
I doubt we’d want to inject `ARCJET_KEY` into clients?

> Given that even SvelteKit's own documentation is hilariously, tragically vague about what their own code is doing ("In prod, this behavior will depend on your adapter."), we are not landing this until someone has done the necessary digging to establish with certainty that this change is not going to cause unexpected behavior for SDK users. In practice that means that somebody has gone into the adapter source code on the Vite / SvelteKit side and figured out how and in which circumstances the adapters are mutating process.env.

I cannot say anything for certain because we do not have exhaustive examples or tests and I do not have exhaustive knowledge of Svelte or SvelteKit.

The part on Vite: The linked OP issue is about Vite 6. `6.2.3` is used in the example. This PR works.

For SvelteKit: this PR reads from `"$env/dynamic/private"`, so private things. The default private prefix is `""`, so no prefix. That means everything is available there. If a user does:

```patch
+++ b/examples/sveltekit/svelte.config.js
@@ -9,6 +9,9 @@ const config = {
 	preprocess: vitePreprocess(),
 
 	kit: {
+		env: {
+			privatePrefix: 'PRIVATE_',
+		},
 		csp: csp(),
 		// adapter-auto only supports some environments, see https://kit.svelte.dev/docs/adapter-auto for a list.
 		// If your environment is not supported, or you settled on a specific environment, switch out the adapter.
```

Then nothing is available. Things become available when the `.env` is:

```
PRIVATE_ARCJET_ENV=production
PRIVATE_ARCJET_KEY=asdasdasd
PRIVATE_ARCJET_LOG_LEVEL=debug
```

Code for passing of `process.env` in Node is:

* https://github.com/sveltejs/kit/blob/1164ffa83d7b4812c812e4e2d6f63e7b1ea1ef98/packages/adapter-node/src/handler.js#L38
* https://github.com/sveltejs/kit/blob/1164ffa83d7b4812c812e4e2d6f63e7b1ea1ef98/packages/kit/src/runtime/server/index.js#L47-L61

blaine-arcjet:
This branch still crashes with the error presented in #982 which #989 attempted to work around. At the time, the issue could be replicated by removing the node adapter; however, this is no longer sufficient. It seems that Vite or SvelteKit now accounts for symlinks from the `file:` specifier so you need to use the following workaround to see the failure:

1. Build these changes
2. Run `npm pack` in the `arcjet-sveltekit` directory
3. Run `npm i ../../arcjet-sveltekit/arcjet-sveltekit-1.0.0-beta.9.tgz` in the `examples/sveltekit` directory (this will also change the `package.json` and `package-lock.json` file)
4. Use `ls` commands to ensure that `@arcjet/sveltekit` is not a symlink
5. Run `npm run dev -- --host`
6. Visit the URL presented

If you do all of the above on this branch, you'll see a 500 Internal Error page and you'll see the same error in #982 in your console:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '$env' imported from /workspaces/arcjet-js/examples/sveltekit/node_modules/@arcjet/sveltekit/index.js
```

wooorm-arcjet:
That’s a very good repro.....

I was at a loss for why `@auth/sveltekit` can do it:

* https://github.com/nextauthjs/next-auth/blob/39dd3b92de194c1a835f2d87631f4deb9d9fdf65/packages/frameworks-sveltekit/src/lib/actions.ts#L4
* https://github.com/nextauthjs/next-auth/blob/39dd3b92de194c1a835f2d87631f4deb9d9fdf65/packages/frameworks-sveltekit/src/lib/index.ts#L286
* (also on npm: https://www.npmjs.com/package/@auth/sveltekit?activeTab=code -> `dist/index.js` -> L284)

That file `index.js` exposes `SvelteKitAuth`, which is used in their examples: https://authjs.dev/reference/sveltekit. And that file load `$env` and `$app` etc.

Soooo, something isn’t making sense. Also look at your error: ERR_MODULE_NOT_FOUND. Rest of the stack trace shows the `at Object.getPackageJSONURL (node:internal/modules/package_json_reader:255:9)` or similar right? Node.js, package.json....

I had `@auth/sveltekit` and `@arcjet-js/sveltekit` next to each other. Replaced all of the code in their `index.js` with empty files except an export and a console.log of `$env`. One worked the other didn’t.
Then I started removing fields from `@auth/sveltekit/package.json` to see when it failed.

This worked. Guess what line made it crash?

```json
{
  "name": "@auth/sveltekit",
  "version": "1.10.0",
  "peerDependencies": {
    "svelte": "^3.54.0 || ^4.0.0 || ^5.0.0-0"
  },
  "type": "module",
  "exports": "./dist/index.js"
}
```

That’s right. The peer dependency.

Adding that to our `arcjet-sveltekit` makes things work.
Running your above steps on this branch now is ✅

arcjet-sveltekit/package.json
```diff
@@ -55,6 +55,9 @@
     "@arcjet/transport": "1.0.0-beta.9",
     "arcjet": "1.0.0-beta.9"
   },
+  "peerDependencies": {
+    "svelte": "^3.54.0 || ^4.0.0 || ^5.0.0-0"
```
wooorm-arcjet, line 59: we may want to narrow this down, we probably do not support every version? Last release of 4 was on 23 august 2024, last release of 3 on june 20 2023

arcjet-rei: I don't have a strong opinion on this, but keeping in mind how infrequently many of our users update even their version of our own SDK, we may want to err on the side of supporting more rather than fewer versions of dependencies. Also, I see no compelling reason to not follow SvelteKit's lead.

wooorm-arcjet: The current release line of `@svelte/kit` (`2.27.0`) started at `2.0.0` 2 years ago with a peer dep in its `package.json` on `"svelte": "^4.0.0 || ^5.0.0-next.0",`.

I think we can use that exact range too.
`@svelte/kit` probably uses `svelte` more than us, so if that works for them, probably also for us. So then I’d favor less peer dep warnings.
People will tell us if things don’t work in practice!

blaine-arcjet: > I think we can use that exact range too.

Yeah, this is specifically a SvelteKit adapter, so we want to peerDep on the version of Svelte that it does.

wooorm-arcjet:
Thanks for help with the digging all!

PR #4755, opened 2025-07-29
fix!: remove arbitrary base url from env in dev

wooorm-arcjet:
As we are getting closer to semver stable, proposing this as a PR!

Closes GH-90.

### DIFF
--- a/env/index.ts
+++ b/env/index.ts
@@ -57,37 +57,22 @@ const baseUrlAllowed = [
 ];
 
 export function baseUrl(env: Env) {
-  // TODO(#90): Remove this conditional before 1.0.0
-  if (isDevelopment(env)) {
-    if (env["ARCJET_BASE_URL"]) {
-      return env["ARCJET_BASE_URL"];
-    }
-
-    // If we're running on fly.io, use the Arcjet Decide Service hosted on fly
-    // Ref: https://fly.io/docs/machines/runtime-environment/#environment-variables
-    if (platform(env) === "fly-io") {
-      return "https://fly.decide.arcjet.com";
-    }
-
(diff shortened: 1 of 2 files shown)


env/index.ts
```diff
@@ -57,37 +57,22 @@ const baseUrlAllowed = [
...
-      return "https://fly.decide.arcjet.com";
-    }
+  // Use ARCJET_BASE_URL if it is set and belongs to our allowlist; otherwise
+  // use the hardcoded default.
+  if (
+    typeof env["ARCJET_BASE_URL"] === "string" &&
+    baseUrlAllowed.includes(env["ARCJET_BASE_URL"])
+  ) {
+    return env["ARCJET_BASE_URL"];
+  }
```
qw-in, line 68: Is it worth logging a warning if it's set and we don't use it?

wooorm-arcjet: An interesting idea, but:

a) it would probably result in repeated logs, I think this is called several times
b) in many cases we use our own logger, which has some benefits, but here we don’t have access to that

If we’d change our handling of env (relatedly I was just going through all that with https://github.com/arcjet/arcjet-js/pull/4754), to have some sort of “`readFromEnvironment(context: ArcjetContext, environment: Record<string, string>): ArcjetEnvironment`”, which would only read once and whose return value would be used everywhere, we could do that. And I’d kinda like that. Reading just once, not allowing mutation.

One more alternative: could we throw?

qw-in: Makes sense!

I'd rather not throw except maybe in development?

davidmytton: We never throw. None of our examples show handling errors that way, so this would be a breaking change which we also don't want to make ;)

PR #4802, opened 2025-08-05
feat: add filter rule

wooorm-arcjet:


arcjet/index.ts
```diff
@@ -350,11 +350,39 @@ function createValidator({
...
+ * @param path
+ *   Path to `value`.
+ * @param value
+ *   Value to validate.
+ * @returns
+ *   Nothing.
+ */
+function validateFilter(path: string, value: unknown): undefined {
+  if (typeof value === "string") {
+    // Perfect.
```
qw-in: Why not early return?

arcjet-fastify/index.ts
```diff
@@ -313,12 +313,11 @@ function toArcjetRequest<Properties extends PlainObject>(
       );
     }
   }
-
   const method = request.method ?? "";
   const host = headers.get("host") ?? "";
   let path = "";
   // Note: there may be a better way to detect `https`, no clue.
-  let protocol = request.server.initialConfig.https ? "https:" : "http:";
+  let protocol = request.server?.initialConfig.https ? "https:" : "http:";
```
qw-in: Since we're here maybe we can make an TODO(#) for the note?

arcjet/index.ts
```diff
@@ -568,12 +608,60 @@ export type SensitiveInfoOptions<Detect> =
...
 
+/**
+ * Filter.
+ */
+export interface Filter {
+  /**
+   * Human readable name of the filter;
+   * used for debugging and logging.
+   */
+  displayName?: string | null | undefined;
```
qw-in: I'm still of the opinion we should make this required - I think it would help the dashboard & logging etc. quite a lot

arcjet/index.ts
```diff
@@ -2077,6 +2165,237 @@ export function protectSignup<const Characteristics extends string[] = []>(
...
+    );
+  }
+  if (!allow && !deny) {
+    throw new Error(
+      "`filter` options error: either `allow` or `deny` must be specified",
+    );
+  }
+
+  const type = "FILTER";
+  const version = 0;
```
qw-in, line 2567: These can probably get hoisted out of fn scope?

arcjet/index.ts
```diff
@@ -2077,6 +2165,237 @@ export function protectSignup<const Characteristics extends string[] = []>(
...
+          // Invalid expressions will throw later when matching and whether they
+          // need remote or local data then doesn’t affect them.
+          .catch(function () {
+            return [];
+          })
+      );
+    }),
+  ).then(function (results) {
+    return new Set(results.flat());
+  });
```
qw-in: I didn't stare at it too hard but looks like maybe could be `Promise.allSettled()`?

examples/fastify/index.js
```diff
@@ -15,32 +15,18 @@ const arcjet = arcjetFastify({
...
-      mode: "LIVE", // Use `DRY_RUN` instead of `LIVE` to only log.
-    }),
-    // Protect against common attacks.
-    // See <https://docs.arcjet.com/shield/reference> for more info.
-    shield({
-      mode: "LIVE", // Use `DRY_RUN` instead of `LIVE` to only log.
+    filter({
+      allow: ['not vpn'],
+      mode: "LIVE",
     }),
```
qw-in: Something I see being confusing is the array of expressions. Is it `&&` or `||` logic? If we only supported _one_ expression that would be very clear due to the syntax I imagine right?

```tsx
allow: ["not vpn", "ip.src == 192.168.0.0/16"]
// VS
allow: "not vpn or ip.src == 192.168.0.0/16"
```

I think supporting a single expression (at least for now) is simpler DX and I'm guessing simpler to execute as well?

---

The other piece that I think is confusing (and I'm not sure I have a great solution) is _does this allow when evaluating to **true**? Or does this allow when evaluating to **false**?_

I'm personally guessing this filter will allow requests when the expression evaluates to **true** but I'm only ~80% certain from looking at the API. Do you have any ideas on how we could make it more obvious? (I'm looking for painfully obvious here)

qw-in (commented):
Some small code comments that you can take or leave 😄 

Big comments are:
1. The filter rule construction api
2. Could we drop `vpn` etc. support for now to simplify? Run 100% locally with no requests?

Additionally I still like the idea of using a tagged template literal for the expressions:

```
const exp: FilterExpression = wf`not vpn`
```

Reasons:
1. We have some control over string interpolation (I'm assuming string interpolation is an anti-pattern here)
2. We can add an explicit type to make clear this isn't just any string
3. It's a nice tag to do potential syntax highlighting in the future (like `gql` / `sql`)

Edit:

Another thought on tagged template literals is we could potentially use them as our composition primitive rather than the array of expressions:

```ts
const notVpn = wf`not vpn`
const isIntranet = wf`ip.src == 192.168.0.0/16`

filter({
  allow: wf`${notVpn} and ${isIntranet}`
})

```

arcjet/index.ts
wooorm-arcjet: that seems at odds with tagged templates + composition + your examples?

arcjet/index.ts
wooorm-arcjet: Hmm. I tried it a bit but it doesn’t feel better to me.
Having all the `const`s above is nice, but they also do work. At the end even one that kicks off a promise. That’s why I kept the more simple options-related stuff above. Then the more complex validation. Then the rest of the `const`s!
Acceptable?

arcjet/index.ts
wooorm-arcjet: ah, yeah, that can work!

examples/fastify/index.js
wooorm-arcjet: I changed the PR to support one thing being passed instead of an array!

This also is more of a broader discussion, whether `allow` even makes sense, I think. Much here is built around denying things.

Separate values is also useful because it allows names/descriptions:

```js
allow: [
  {displayName: "allow anyone not using a vpn", expression: "not vpn"},
  {displayName: "also allow the local network", expression: "ip.src == 192.168.0.0/16"}
]
```

…and separate values becomes more useful around larger expressions.

***

Yes, `allow` evaluates individual expressions and the first one that matches is used, which is equivalent to `or`.
`deny` is the same. That means your 2 examples **are** equal.

In my head its easier to think about whether things match, rather than the word `true` 🤔

arcjet-fastify/index.ts
wooorm-arcjet: I *think* it is correct (I wouldn’t know how to dynamically configure https? doesn‘t make much sense to do that right? https://fastify.dev/docs/latest/Reference/HTTP2/#secure-https). But I don’t know enough to say that for certain 🤷 I dunno. I can make one if you want!

wooorm-arcjet:
> anti-pattern

maybe, evaluating expressions is safe, but if you build expressions dynamically, and people could inject strings, then they could change whether expressions match. But at that point...
Instead, `country` from IP details seems to be an [ISO 3166-1 alpha 2](https://github.com/wooorm/iso-3166) code, and I could imagine people wanting to pull in lists and work on them and put them in there?

I *am* in favor of building things *on* these expressions, but us making rules above them, seems better than a sort of DSL above the DSL? It’s a bit “feeling” like on my part, so I should go more in depth, but then just like remote data, I don’t think that has to be here and now, can go somewhere else!

> Could we drop vpn etc. support for now to simplify? Run 100% locally with no requests?

Yup, it’s an important part to hack on for me now, but if this is otherwise :+1: from people I can move that to a different branch!

arcjet/index.ts
wooorm-arcjet: was just a *little* bit harder on my brain to add several returns but still keep track of what the remaining type of `value` is

arcjet-fastify/index.ts
qw-in: It's all good

arcjet/index.ts
qw-in: My thought is it's very helpful in the dashboard ui rather than the expression or some hash

Edit: I'll expand in a main comment as I think I see our confusion

qw-in:
> I could imagine people wanting to pull in lists and work on them and put them in there

The tagged templates can't actually stop them and I'd be curious to see if this comes up in feedback

I thought the main concern would be expression stability for caching?

> displayName

I was imagining each filter rule would have one top level name - rather than one per expression if that makes sense

> sort of DSL above the DSL

I can see that pov. It's a pretty common pattern for essentially string building of sql etc. I wouldn't recommend adding interpolation support in this pr was a future thought

> can move that to a different branch

I think this would be great 👍

wooorm-arcjet:
> I was imagining each filter rule would have one top level name - rather than one per expression if that makes sense

From what I gathered, just like with `detectBot` having a list of things to allow or deny, and I think other rules are similar, it is intentional to use such rules *once*. So not different `detectBot()` calls. And then also not different `filter()` calls 🤔

examples/fastify/index.js
davidmytton: I wanted to reiterate the idea of primitives and products here. I imagine filter rules being the primitive with the powerful expression language. It will likely be used by security engineers for precise tweaks rather than developers, and likely via remote rules than in code. 

I also want us to include several "products" that will configure the filter rule primitives for common use cases. `denyCountry()` could be one that you import and set up as a rule, with type-safe country codes as a parameter array. Deny or allow by IP (single or CIDR) could be another.

This gives us the powerful primitive, but still provide a great DX on the product.

blaine-arcjet:
> I was imagining each filter rule would have one top level name - rather than one per expression if that makes sense

This is correct. I talked to @morekid about this as well and we're planning to add `displayName` as a top-level configuration on every rule type. I still need to open the issue.

blaine-arcjet:
I'm going to move this to a Draft since @wooorm-arcjet and I are doing further discovery on Monday.

arcjet/index.ts
```diff
@@ -2077,6 +2156,144 @@ export function protectSignup<const Characteristics extends string[] = []>(
...
+      return ruleResult;
+    },
+    type,
+    validate(
+      _: ArcjetContext,
+      details: Partial<ArcjetRequestDetails>,
+    ): undefined {
+      if (!details.ip) {
+        throw new Error("Request filtering requires `ip` to be set");
+      }
```
blaine-arcjet:
```suggestion
```

arcjet/index.ts
```diff
@@ -2077,6 +2156,144 @@ export function protectSignup<const Characteristics extends string[] = []>(
...
+    },
+
+    version,
```
blaine-arcjet:
```suggestion
```

arcjet/index.ts
```diff
@@ -568,12 +603,56 @@ export type SensitiveInfoOptions<Detect> =
...
   | SensitiveInfoOptionsDeny<Detect>;
 
+/**
+ * Configuration to allow if a filter matches and deny otherwise.
+ */
+export type FilterOptionsAllow = {
+  /**
+   * One or more expressions.
+   */
+  allow: ReadonlyArray<string> | string;
```
blaine-arcjet:
```suggestion
  allow: ReadonlyArray<string>;
```

arcjet/index.ts
```diff
@@ -568,12 +603,56 @@ export type SensitiveInfoOptions<Detect> =
...
+export type FilterOptionsDeny = {
+  /**
+   * One or more expressions,
+   * must not be set if `deny` is set.
+   */
+  allow?: never;
+  /**
+   * One or more expressions.
+   */
+  deny: ReadonlyArray<string> | string;
```
blaine-arcjet:
```suggestion
  deny: ReadonlyArray<string>;
```

arcjet/index.ts
```diff
@@ -455,6 +478,18 @@ const validateShieldOptions = createValidator({
...
 
+/**
+ * Validate filter options.
+ */
+const validateFilterOptions = createValidator({
+  rule: "filter",
+  validations: [
+    { key: "mode", required: false, validate: validateMode },
+    { key: "allow", required: false, validate: validateFilters },
+    { key: "deny", required: false, validate: validateFilters },
```
blaine-arcjet:
```suggestion
    { key: "allow", required: false, validate: validateStringArray },
    { key: "deny", required: false, validate: validateStringArray },
```

arcjet/index.ts
```diff
@@ -350,6 +353,26 @@ function createValidator({
...
+function validateFilters(path: string, value: unknown): undefined {
+  if (Array.isArray(value)) {
+    for (const [idx, item] of value.entries()) {
+      validateString(`${path}[${idx}]`, item);
+    }
+  } else {
+    validateString(path, value);
+  }
+}
+
```
blaine-arcjet:
```suggestion
```

arcjet-fastify/index.ts
```diff
@@ -313,7 +313,6 @@ function toArcjetRequest<Properties extends PlainObject>(
       );
     }
   }
```
blaine-arcjet:
```suggestion
  }

```

Unrelated to diff

arcjet/index.ts
```diff
@@ -568,12 +603,56 @@ export type SensitiveInfoOptions<Detect> =
...
+ */
+export type FilterOptions = FilterOptionsAllow | FilterOptionsDeny;
+
 const Priority = {
   SensitiveInfo: 1,
   Shield: 2,
   RateLimit: 3,
   BotDetection: 4,
   EmailValidation: 5,
+  Filter: 6,
```
blaine-arcjet: Priority is probably 2

arcjet/index.ts
```diff
@@ -2077,6 +2156,144 @@ export function protectSignup<const Characteristics extends string[] = []>(
...
+  const allow: ReadonlyArray<string> = Array.isArray(options.allow)
+    ? options.allow
+    : options.allow
+      ? [options.allow]
+      : [];
+  const deny: ReadonlyArray<string> = Array.isArray(options.deny)
+    ? options.deny
+    : options.deny
+      ? [options.deny]
+      : [];
```
blaine-arcjet:
```suggestion
  const allow = options.allow ?? [];
  const deny = options.deny ?? [];
```

blaine-arcjet (commented):
I've provided some simplifications as commit-able suggestions.

arcjet/index.ts
wooorm-arcjet: okay, `ip` is no longer required

arcjet/index.ts
wooorm-arcjet: This introduces a type error, because `ArcjetRule` requires a `version` field. What do you want to do?

arcjet/index.ts
wooorm-arcjet: This stems from the discussion here: https://github.com/arcjet/arcjet-js/pull/4802#discussion_r2257606781. I think I agree with Quinn and, for filters, want to allow a single expression.

arcjet/index.ts
wooorm-arcjet: This stems from the discussion here: https://github.com/arcjet/arcjet-js/pull/4802#discussion_r2257606781. I think I agree with Quinn and, for filters, want to allow a single expression.

arcjet/index.ts
wooorm-arcjet: This stems from the discussion here: https://github.com/arcjet/arcjet-js/pull/4802#discussion_r2257606781. I think I agree with Quinn and, for filters, want to allow a single expression.

arcjet/index.ts
wooorm-arcjet: This stems from the discussion here: https://github.com/arcjet/arcjet-js/pull/4802#discussion_r2257606781. I think I agree with Quinn and, for filters, want to allow a single expression.

arcjet-fastify/index.ts
wooorm-arcjet: done

arcjet/index.ts
wooorm-arcjet: okay done.
Can you provide some reasoning for why 2 is the best value for filters?

arcjet/index.ts
wooorm-arcjet: This stems from the discussion here: https://github.com/arcjet/arcjet-js/pull/4802#discussion_r2257606781. I think I agree with Quinn and, for filters, want to allow a single expression.

arcjet/index.ts
blaine-arcjet: Rules are uniform. Changing the uniformity of rules is out of scope. Please open a follow up issue.

wooorm-arcjet: sad. done.

arcjet/index.ts
wooorm-arcjet: done.

arcjet/index.ts
wooorm-arcjet: done.

arcjet/index.ts
wooorm-arcjet: done.

arcjet/index.ts
wooorm-arcjet: done.

wooorm-arcjet:
@blaine-arcjet @arcjet-rei I think this is ready to land. I would appreciate a final review!

analyze/test/analyze.test.ts
```diff
@@ -280,3 +282,29 @@ test("isValidEmail", async function (t) {
 
   // TODO: test `allow` option.
 });
+
+test("matchFilters", async function (t) {
```
blaine-arcjet, line 286: These tests only handle `allowIfMatch == true` but perhaps we should have tests for `allowIfMatch == false`.

analyze/index.ts
```diff
@@ -238,3 +240,39 @@ export async function detectSensitiveInfo(
...
+      // @ts-expect-error: WebAssembly does not support readonly values.
+      expressions,
+      allowIfMatch,
+    );
+    // Ignore the `else` branch as we test in places that have WebAssembly.
+    /* node:coverage ignore next 4 */
+  }
+
+  context.log.debug("WebAssembly is not supported in this runtime");
+  return;
```
blaine-arcjet: I think this should either throw like the `detectSensitiveInfo` function or return an empty value like `detectBot`. Either case will cause the rule to be executed on the server, so it is mainly about the callsite and if it would be better to represent it as an error result or allowed result.

arcjet/index.ts
```diff
@@ -2077,6 +2136,128 @@ export function protectSignup<const Characteristics extends string[] = []>(
...
+        const result = await analyze.matchFilters(
+          context,
+          request_,
+          allow.length > 0 ? allow : deny,
+          allow.length > 0,
+        );
+
+        if (!result) {
+          throw new Error("WebAssembly is not supported in this runtime");
+        }
```
blaine-arcjet: It looks like you are throwing here so we should move this into the analyze function like `detectSensitiveInfo`.

arcjet/index.ts
```diff
@@ -2077,6 +2136,128 @@ export function protectSignup<const Characteristics extends string[] = []>(
...
+
+        ruleResult = new ArcjetRuleResult({
+          conclusion: result.allowed ? "ALLOW" : "DENY",
+          fingerprint: context.fingerprint,
+          reason: new ArcjetFilterReason(result.matchedExpressions[0]),
+          ruleId,
+          state,
+          ttl: 60,
```
blaine-arcjet: `ttl` is only set when DENY

arcjet/index.ts
```diff
@@ -2077,6 +2136,128 @@ export function protectSignup<const Characteristics extends string[] = []>(
...
+        });
+      }
+
+      context.cache.set(
+        ruleId,
+        context.fingerprint,
+        ruleResult,
+        ruleResult.ttl,
+      );
```
blaine-arcjet: Rules do not manage setting their own values into cache, it is handled by the protect logic because dry run results are not cached, etc.

arcjet/index.ts
```diff
@@ -2077,6 +2136,128 @@ export function protectSignup<const Characteristics extends string[] = []>(
...
+          ttl: 60,
+        });
+      } catch (error) {
+        ruleResult = new ArcjetRuleResult({
+          conclusion: "ERROR",
+          fingerprint: context.fingerprint,
+          reason: new ArcjetErrorReason(String(error)),
+          ruleId,
+          state,
+          ttl: 60,
```
blaine-arcjet:
```suggestion
          ttl: 0,
```
No ttl on errors

arcjet/index.ts
```diff
@@ -2077,6 +2136,128 @@ export function protectSignup<const Characteristics extends string[] = []>(
...
+          reason: new ArcjetFilterReason(result.matchedExpressions[0]),
+          ruleId,
+          state,
+          ttl: 60,
+        });
+      } catch (error) {
+        ruleResult = new ArcjetRuleResult({
+          conclusion: "ERROR",
+          fingerprint: context.fingerprint,
+          reason: new ArcjetErrorReason(String(error)),
```
blaine-arcjet:
```suggestion
          reason: new ArcjetErrorReason(error),
```
ErrorReason accepts an error value from a catch

arcjet/test/filter.test.ts
```diff
@@ -0,0 +1,461 @@
...
+      mode: "LIVE",
+    });
+    const first = await rule.protect(context, createRequest());
+    assert.equal(first.conclusion, "DENY");
+    assert.equal(first.state, "RUN");
+
+    const second = await rule.protect(context, createRequest());
+    assert.equal(second.conclusion, "DENY");
+    assert.equal(second.state, "CACHED");
+  });
```
blaine-arcjet: It doesn't set cache itself so this should probably test that it reads from a populated cache.

blaine-arcjet (commented):
This is very close! We just need to cleanup some stuff around caching/ttl and throwing behavior.

arcjet/index.ts
blaine-arcjet: The way I think about priority is that the most likely to DENY locally should be first and slow/server-only rules should be last. Since Filter expressions operate against the request data, it should match frequently.

arcjet/index.ts
blaine-arcjet: I was just trying to move properties to the top but it looks like you sorted alphabetically. This trips me up when combined with functions but that preference isn't a blocker.

arcjet/test/filter.test.ts
```diff
@@ -0,0 +1,461 @@
...
+    assert.throws(function () {
+      filter({
+        // @ts-expect-error: test runtime behavior of invalid `deny[]` value.
+        deny: [1],
+      });
+    }, /invalid type for `deny\[0]` - expected string/);
+  });
+
+  await t.test(
+    "should fail if `allow` and `deny` are both given",
```
arcjet-rei: Not to be pedantic, but this test does not ensure that there is a failure when `allow` and `deny` are both provided, given that the returned error is that there are no expressions to match in either one. Add another test case that includes a valid expression in both arrays?

arcjet-rei (commented):
I found one erroneous test case and am running into one test failure:

```
@arcjet/ip:test: ✖ should expose the public api (2.180583ms)
@arcjet/ip:test:   AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
@arcjet/ip:test:   + actual - expected
@arcjet/ip:test:
@arcjet/ip:test:     [
@arcjet/ip:test:       'default',
@arcjet/ip:test:   +   'findIp',
@arcjet/ip:test:       'parseProxy'
@arcjet/ip:test:     ]
@arcjet/ip:test:
@arcjet/ip:test:       at TestContext.<anonymous> (file:///Users/rei/Documents/projects/arcjet-code/arcjet-js/ip/test/index.test.js:6:16)
@arcjet/ip:test:       at async Test.run (node:internal/test_runner/test:1113:7)
@arcjet/ip:test:       at async TestContext.<anonymous> (file:///Users/rei/Documents/projects/arcjet-code/arcjet-js/ip/test/index.test.js:5:5)
@arcjet/ip:test:       at async Test.run (node:internal/test_runner/test:1113:7)
@arcjet/ip:test:       at async startSubtestAfterBootstrap (node:internal/test_runner/harness:358:3) {
@arcjet/ip:test:     generatedMessage: true,
@arcjet/ip:test:     code: 'ERR_ASSERTION',
@arcjet/ip:test:     actual: [ 'default', 'findIp', 'parseProxy' ],
@arcjet/ip:test:     expected: [ 'default', 'parseProxy' ],
@arcjet/ip:test:     operator: 'deepStrictEqual',
@arcjet/ip:test:     diff: 'simple'
@arcjet/ip:test:   }
```

I assume that this is related to the fact that IP categorization is still a work in progress, but noting it anyway.

On the whole, this is looking good and it feels like we're getting close. Ganbatte kudasai!

analyze/test/analyze.test.ts
wooorm-arcjet: good idea, done!

analyze/index.ts
wooorm-arcjet: I added an error, as that’s what now happens in arcjet core anyway, and then, is caught again.
Especially, with the addition of undetermined results, which this sort of is, an “allow” doesn’t feel exactly right.

arcjet/index.ts
wooorm-arcjet: done

arcjet/index.ts
wooorm-arcjet: done

arcjet/index.ts
wooorm-arcjet: done

arcjet/index.ts
wooorm-arcjet: done

arcjet/index.ts
wooorm-arcjet: done

arcjet/test/filter.test.ts
wooorm-arcjet: difficult to do because the `ruleId` is generated inside the rule.
Instead, I tested that `rule.protect` does not cache, but `aj.protect({rules: [rule]})` does!

arcjet/test/filter.test.ts
wooorm-arcjet: done

wooorm-arcjet:
> I found one erroneous test case and am running into one test failure: [...] I assume that this is related to the fact that IP categorization is still a work in progress, but noting it anyway.

This looks like a `.js` file left over from another branch, and switching to it.
`ip/test/index.ts` no longer exists on `main`, but the `.js` file was left over on your device!

When switching back from this branch to main, you’ll also notice that `arcjet/test/filter.test.js` remains.
A bit annoying. With Node adding support for reading TS files itself, we could stop generating `.js` files for test files: Related to https://github.com/arcjet/arcjet-js/issues/4477.

analyze/test/analyze.test.ts
wooorm-arcjet: done

arcjet/test/filter.test.ts
blaine-arcjet: Since the cache is pulled from the context, you could mock the cache layer, but I don't think it is necessary because your test works. 👍

blaine-arcjet (approved):
Awesome work! Glad to see this ready to land. One thing that I realized we should add to the rule-feature-checklist is an example. I only remembered this because I needed to tweak an example to test a filter rule locally.

wooorm-arcjet:
Thanks for the reviews everyone!

PR #4852, opened 2025-08-08
refactor(arcjet): use `interface`s to share fields

wooorm-arcjet:
To prepare for more JSDocs in `arcjet` (core),
I switch to interfaces.
Interfaces support a nice chain of extension which helps for our system of mutually exclusive options.

However, `interface`s are in some small ways more strict than putting some object into a `type`:
in certain cases they do not support indexing into them. This is a good thing but it introduces a type issue so I added some code to handle non-objects as options.
The code is useful because people *could* indeed accidentally pass a non-object.

### DIFF
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -329,13 +329,20 @@ function createValidator({
   rule: string;
   validations: ValidationSchema[];
 }) {
-  return (options: Record<string, unknown>) => {
+  return (options: unknown) => {
+    if (options === null || typeof options !== "object") {
+      throw new Error(`\`${rule}\` options error: expected an object`);
+    }
+
+    // Cast to allow arbitrary indexing.
+    const record = options as Record<string, unknown>;
+
     for (const { key, validate, required } of validations) {
-      if (required && !Object.hasOwn(options, key)) {
+      if (required && !Object.hasOwn(record, key)) {
         throw new Error(`\`${rule}\` options error: \`${key}\` is required`);
       }
 
-      const value = options[key];
+      const value = record[key];
 
       // The `required` flag is checked above, so these should only be validated
       // if the value is not undefined.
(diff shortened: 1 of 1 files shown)


blaine-arcjet (changes requested):
Interfaces are a type that says "an object containing *at least* these properties" but they can contain more. This is wrong because we take an exact options object.

blaine-arcjet:
@arcjet-rei asked me to add additional context here.

When your types are an overlay on an object ("at least these fields") then every change you make to them is breaking. For example, if a user specifies options with an extra field (`foobar: string`) and then you add that field to your interface such as `foobar: number`, you've now broken the user and cause their upgrade to be more difficult than it needs to be (and possibly have them abandon it).

When you use an exact object type, you know that adding an optional field will never break them because the previous object could not have that field.

arcjet-rei:
> To prepare for more JSDocs in arcjet (core), I switch to interfaces.

How does this change support that goal?

> Interfaces support a nice chain of extension which helps for our system of mutually exclusive options.

Helps in what ways?

wooorm-arcjet:
Incorrect: https://www.typescriptlang.org/play/?#code/JYOwLgpgTgZghgYwgAgJLmvJB5ADmYAexAGdkBvAKGWRMIFsIAlCARwFdgoIATAMWAQANjwBcyAEaFCQiHBCUAvpUpgAnrhQAVDRACCQ4HBJ4CxMgF4K1Wg2ZtO3foJHipMuQsUBuFTHYgCGYgyIhI+OiQsIgQABSE+ESk4pGYMaZJJACU4gE8EDCgvBTIypT+gcGhCOFgOpoGRiTxiebi9fqGxhnmOch5BUU8JWVhEBEY0Uix5HSMLBxcvALCYshgUOwQADS2jNhgABbQK67I8EIkEIpZKmP4HY3GM3P2i06naxtbu68Hx1BPuILlcbkA

> How does this change support that goal?

Less repeated JSDocs.

> Helps in what ways?

Less repeated JSDocs and copy/pasted code.

blaine-arcjet:
I am indeed incorrect about this difference. I assumed the difference was between `type` and `interface` but it is actually between object literals and variables holding objects. Variables holding objects can be in whatever shape and object literals are marked as an error when they include extra properties. It doesn't seem possible to restrict usage to object literals, so we need to consider all changes to options objects as possibly breaking.

> Less repeated JSDocs.

JSDocs should not be repeated. They should be catered to the rule they are specifying. An `allow` list for bot detection should not be documented the same was as an `allow` list for email validation. With the different rate limit algorithms, the fields often mean different things.

In the rare case something is uniform (e.g. `mode`), trying to DRY this code to reduce adding a doc to each one seems counterintuitive to actually being flexible in the design to adapt to changes a startup requires.

wooorm-arcjet:
> so we need to consider all changes to options objects as possibly breaking.

I think this is in the end the same as anything that potentially introduces/stops introducing a type error, being breaking. Which is why TS doesn’t follow semver.
You can extend that yo any change to any linter. Or even, anything that generates a string of HTML/markdown/etc.
And then you get at https://xkcd.com/1172/.

But that’s what TypeScript is *for*. It *should* introduce type errors when people do things wrong.
And folks can themselves choose how strict they are in their types:

```ts
const options1: Options = {someRequiredField: true, someOtherField: false}; // Type error for `someOtherField`
const options2 = {someRequiredField: true, someOtherField: false} satisfies Options;  // Type error for `someOtherField`
const options3 = {someRequiredField: true, someOtherField: false} as Options;
const options4 = {someRequiredField: true, someOtherField: false};

doThings(options1)
doThings(options2)
doThings(options3)
doThings(options4)

interface Options {
  someRequiredField: boolean
}

function doThings(options: Options): undefined { }
```

https://www.typescriptlang.org/play/?#code/MYewdgzgLgBCAOUCW4IEYBcMDyiWRgF4YBvCEAWwFMAlKgRwFckAnKgEwDEkqAbdrFBaMqAGhjlq2KAAsqLbnwEwAZgENeEKgF8A3DAD0BmABUAnvCox5LEC1V2YAA0lVpchT35OAUKEiwCMioAExEpK50TKwcivyCwmISlG6y8nHK6po6EmrIECo8EDh4qPqGxuaW1iy29iqOLinu6V7svv7QcKWQAMzhZClRzGxcbQki4q4tnkpYWVraMGrFuMGQun6ogT0QACwDkQwjseMwQpPJUmmz8aoai5s+7CAmMkhgAOYQABRB+OgAJTPV7vL6-f6hYEvN4fb5-Xa9aGguEQ3Z7YE+D5QeTqYBWNYA0g+GBXWjHGJjOYwABGIBAvCoajAPm0Ph8KkYYGA6xgMLB8MhkCwhNQgKwXPYVEKYA4pBgbKAA

> JSDocs should not be repeated. They should be catered to the rule they are specifying. An allow list for bot detection should not be documented the same was as an allow list for email validation.

Yes, indeed!

> With the different rate limit algorithms, the fields often mean different things.

Indeed. In this PR, for rate limiting, `mode` and `characteristics` are shared. The other fields are specific.

> In the rare case something is uniform (e.g. mode), trying to DRY this code to reduce adding a doc to each one seems counterintuitive to actually being flexible in the design to adapt to changes a startup requires.

I didn’t think it was that rare, if it was once or twice then sure but it is most rules, and there will be more rules...
This PR exists because I am documenting the APIs in readmes. For that, JSDocs, and for that, some small changes so I can better explain the APIs to humans. APIs are also on the website.

Many PRs are small on their own, seemingly insignificant. Sure. But I want to explain to humans that a rule accepts these 3 shared options, and then either `allow` but not `deny`, or `deny` but not `allow`.

PR #4854, opened 2025-08-08
docs(nosecone): add derivative works to readme

wooorm-arcjet:
I noticed that some parts of `nosecone` were inspired by 2 other tools. As their licenses are included in the source code I think that it is a good idea to mention them in the readme too.

Related-to: GH-4654.

### DIFF
--- a/nosecone/README.md
+++ b/nosecone/README.md
@@ -72,7 +72,7 @@ Derivative work based on
 licensed under
 [MIT][github-csp-typed-directives-license] © Joshua Hemphill and
 Tecnico Corporation.
-Our work picks the couple types the we need.
+Our work picks the couple types that we need.
 It is also based on
 [`helmetjs/helmet`][github-helmet] licensed under
 [MIT][github-helmet-license] © Evan Hahn and Adam Baldwin.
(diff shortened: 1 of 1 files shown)


nosecone/README.md
```diff
@@ -67,5 +67,20 @@ console.log(response);
 
 [Apache License, Version 2.0][apache-license] © [Arcjet Labs, Inc.][arcjet]
 
+Derivative work based on
+[`josh-hemphill/csp-typed-directives`][github-csp-typed-directives]
+licensed under
+[MIT][github-csp-typed-directives-license] © Joshua Hemphill and
+Tecnico Corporation.
+Our work picks the couple types the we need.
```
qw-in: Wording doesn't quite make sense here

qw-in (approved):

nosecone/README.md
wooorm-arcjet: Hmm. Good point. `few` instead of `couple`? Do you a good suggestion?

Oh also typo, `the` -> `that`

qw-in: Yeah just the typo fix would sort it 👍

wooorm-arcjet: Fixed!

PR #4856, opened 2025-08-08
refactor(redact): reuse exposed types

wooorm-arcjet:
Some of the types from `@arcjet/redact-wasm` are not used anywhere. Other ones were redefined in `@arcjet/redact` itself. This commit removes the duplication and removes the (completely internal) unused API surface from `@arcjet/redact-wasm`.

Some of the exposed API surface did not specify return types either. This adds return types so that we can verify that the API matches what we want instead of returning arbitrary objects (with `const`s).

### DIFF
--- a/redact/index.ts
+++ b/redact/index.ts
@@ -1,14 +1,17 @@
 import { initializeWasm } from "@arcjet/redact-wasm";
-import type { SensitiveInfoEntity } from "@arcjet/redact-wasm";
+import type {
+  RedactedSensitiveInfoEntity as RedactedSensitiveInfoEntityWasm,
+  RedactSensitiveInfoConfig,
+  SensitiveInfoEntity,
+} from "@arcjet/redact-wasm";
 
 /**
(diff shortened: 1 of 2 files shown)


redact/index.ts
```diff
@@ -1,14 +1,17 @@
...
  */
-export type ArcjetSensitiveInfoType =
-  | "email"
-  | "phone-number"
-  | "ip-address"
-  | "credit-card-number";
+export type ArcjetSensitiveInfoType = Exclude<
+  SensitiveInfoEntity["tag"],
+  "custom"
+>;
```
qw-in, line 14: I'm not sure if this is public, but the computed version typically is a worse dx experience in code editors

wooorm-arcjet: It’s a good point.
There is some more redefinition also in `arcjet` core.
So my primary motivation is that defining things 3 times will *probably* get us into trouble when we add more things.
Especially because these are generated in the Wasm.

PR #4858, opened 2025-08-08
docs(stable-hash): add derivative work to readme

wooorm-arcjet:
I noticed that `stable-hash` was inspired by `feross/buffer`. As its license is included in the source code I think that it is a good idea to mention that in the readme too.

Related-to: GH-4654.
Related-to: GH-4854.


qw-in:
Oh, it might be good to add many of these PRs to a "SUSTAINING: js-sdk" GitHub project or similar

PR #4859, opened 2025-08-08
refactor(protocol): remove unneeded type utility

wooorm-arcjet:
This is shorter in almost all cases.
And, faster and simpler to be explicit.

### DIFF
--- a/protocol/index.ts
+++ b/protocol/index.ts
@@ -8,8 +8,6 @@ export type * from "./well-known-bots.js";
 // Re-export the Bot categories from the generated file
 export { categories as botCategories } from "./well-known-bots.js";
 
-type RequiredProps<T, K extends keyof T> = Required<Pick<T, K>> & Omit<T, K>;
-
 /**
  * Mode of a rule.
  */
@@ -1035,7 +1033,7 @@ export class ArcjetIpDetails {
    * @returns
    *   Whether the IP address has latitude info.
    */
-  hasLatitude(): this is RequiredProps<this, "latitude" | "accuracyRadius"> {
+  hasLatitude(): this is { accuracyRadius: number; latitude: number } {
     return typeof this.latitude !== "undefined";
   }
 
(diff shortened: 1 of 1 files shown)


qw-in (approved):
It looks like we may lose some safety here in that we need to keep these assertion types in sync (for instance if `timezone` changed to a `TimeZone` type or similar). Maybe TypeScript still checks? I'm not sure

That said I think the simplification is still worth it 👍🏻

qw-in:
Looks like we do lose the safety: https://www.typescriptlang.org/play/?#code/FAYwNghgzlAECyBPAwpGsDexi1gMwEsBTMAEwH4AuWAOwFcBbAIyICcBubXAC2gDFiZAIwAKAJTUALtwJxZmfINLUok1gRoBzWAF9MOXLFZFJdVjVhq6RAzuw6gA

Is there a different way to achieve similar safety?

wooorm-arcjet:
Right!
I would importantly clarify that for a *user* this is not different.
Indeed, for us, we could mess up: TS does not check assertions/predicates.
I think that point is more an argument *against* functions: a person too can check for `timezone` to exist, and then TS does its thing.
However, codepilot did a good job.
And with good tests (not existing yet, but would be in `protocol.test.ts`) we could catch it

qw-in:
Maybe this is an opportunity to add those tests to make sure codepilot did a good job 😜

wooorm-arcjet:
Noted!

blaine-arcjet:
This change decouples the assertions from the original definition, so when you change the original definition, you will not get type hints that you need to change these assertions.

We should revert this change.

arcjet-rei:
For future reference, the thing I / we should have caught when reviewing this change is that replacing the explicit type guards with structural typing leads to this decoupling between guards and definitions, and that is a problem because it introduces opportunities for us to inadvertently introduce breaking changes.

PR #4873, opened 2025-08-11
refactor(logger): use more specific type than `unknown`

wooorm-arcjet:
The first parameter of the log methods is currently typed as unknown. However we do know that it is one of two types: a plain object (merging object) or a `string` (message).
We can use the more specific type that we use instead of an `unknown`.

### DIFF
--- a/logger/index.ts
+++ b/logger/index.ts
@@ -62,17 +62,17 @@ function getMessage(
 }
 
 function getOutput(
-  mergingObject: unknown,
+  messageOrObject: Record<string, unknown> | string,
   message: unknown,
   interpolationValues: unknown[],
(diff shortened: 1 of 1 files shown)


logger/index.ts
```diff
@@ -147,12 +147,12 @@ export class Logger {
     ...interpolationValues: unknown[]
   ): void;
   debug(
-    mergingObject: unknown,
+    messageOrObject: Record<string, unknown> | string,
```
wooorm-arcjet, line 150: note for reviewers: there are overloads above that are more strict: either the record, or a string, must be passed.

qw-in (approved):

PR #4875, opened 2025-08-11
chore(sprintf): expose named export, deprecate default

wooorm-arcjet:
This commit adds a named export for `sprintf`.

Related-to: GH-4860.


wooorm-arcjet:
Not everything needs an ADR.

PR #4876, opened 2025-08-11
chore(ip): expose named export, deprecate default

wooorm-arcjet:
This commit adds a named export for `findIp`.

Related-to: GH-4723.
Related-to: GH-4860.
Related-to: GH-4875.


wooorm-arcjet:
Not everything needs an ADR.

PR #4877, opened 2025-08-11
feat(ip): expose `Cidr` type

wooorm-arcjet:
This exposes the `Cidr` type which is useful in other places, such as integrations.

Related-to: GH-4723.

### DIFF
--- a/ip/index.ts
+++ b/ip/index.ts
@@ -1023,6 +1023,6 @@ function findIp(
 /**
  * One of the CIDR ranges.
  */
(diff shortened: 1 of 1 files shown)


qw-in (approved):
Should we export the v4 & v6 variants as well?

wooorm-arcjet:
🤔 not yet I think.

I needed the type so that I can use the function and put its return values in other types.
I think that’s a reasonable expectation for people around APIs: that they can relatively easily access input/output types.
But I don’t yet see a good use case for narrower needs? So I’d say: when we get there!

PR #4878, opened 2025-08-11
chore(nosecone): expose named export, deprecate default

wooorm-arcjet:
This commit adds a named export for `nosecone`.
The default export is still supported but deprecated.

Related-to: GH-4860.
Related-to: GH-4875.
Related-to: GH-4876.

PR #4880, opened 2025-08-11
chore(nosecone-sveltekit): expose named export, deprecate default

wooorm-arcjet:
This commit adds a named export for `nosecone`.

Related-to: GH-4723.
Related-to: GH-4860.
Related-to: GH-4875.
Related-to: GH-4878.
Related-to: GH-4879.


wooorm-arcjet:
Not everything needs an ADR.

PR #4883, opened 2025-08-11
refactor(redact)!: improve types around custom entities

wooorm-arcjet:
The `redact` function accepts two type parameters.
The second type parameter `CustomEntities` is not used. Looking at the tests (or the examples in #4982). It is always `string`. Never something more specific.
The first type parameter, `Detect`, is an entire function, whose variadic part is that unused `CustomEntities`.
Still, it is accepted as a `const` and used in `RedactOptions`, where it does work to influence the `entities` and `replace` options.
However, it is not *needed* to put an entire function into a type parameter, if only a small part is variadic. This is not a real issue. Just a tiny bit smelly.

Still, there are some cases where type errors happen for valid use cases. This PR changes the type parameters and internal code to allow these valid use cases. With tests.

Closes GH-4982.

### DIFF
--- a/redact/index.ts
+++ b/redact/index.ts
@@ -13,59 +13,53 @@ export type ArcjetSensitiveInfoType = Exclude<
   "custom"
 >;
 
-type DetectSensitiveInfoEntities<T> = (
-  tokens: string[],
-) => Array<ArcjetSensitiveInfoType | T | undefined>;
-
-type ValidEntities<Detect> = Array<
-  // Via https://www.reddit.com/r/typescript/comments/17up72w/comment/k958cb0/
-  // Conditional types distribute over unions. If you have ((string | undefined)
-  // extends undefined ? 1 : 0) it is evaluated separately for each member of
-  // the union, then union-ed together again. The result is (string extends
-  // undefined ? 1 : 0) | (undefined extends undefined ? 1 : 0) which simplifies
-  // to 0 | 1
-  undefined extends Detect
-    ? ArcjetSensitiveInfoType
-    : Detect extends DetectSensitiveInfoEntities<infer CustomEntities>
-      ? ArcjetSensitiveInfoType | CustomEntities
-      : never
->;
-
 /**
- * Redact sensitive information.
- *
- * @param entity
- *   Entity to redact.
- * @param plaintext
- *   The plaintext string to redact.
- * @returns
- *   Redacted string or nothing.
+ * Types of sensitive information that can be detected which also allows
+ * custom tags (arbitrary strings) while still retaining autocompletion and other IDE features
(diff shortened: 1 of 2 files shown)


blaine-arcjet (changes requested):
I'm pretty sure this breaks the type inference between Detect and Replace. These types were written so the types returned by Detect are propagated into the type signature of Replace.

wooorm-arcjet:
No. Try it out.

wooorm-arcjet:
@blaine-arcjet could you make a video of what did/didn’t work for you?
When working on this, I did check autocomplete and other IDE things.
From the OP:
> There is now good autocomplete when writing options. And a bug was caught in the tests, where a value was replaced in a replace function, but that value was not searched for in entities.

wooorm-arcjet:
I have confirmed that the bug from GH-1455 is no longer present here too. So if these types are reused in `arcjet` core, that issue can be closed. Example reproducing that problem (on main) and showing that this PR solves that:

```ts

const result1 = redact("asd", {
  entities: ["custom", "credit-card-number"],
  detect: (tokens: string[]) => {
      if (token === "test@example.com") {
        return "custom";
      }
    });
  },
  contextWindowSize: 1,
});

const result2 = redact("asd", {
  // Type error for `custom`
  entities: ["custom", "credit-card-number"],
  detect(tokens: string[]) {
      if (token === "test@example.com") {
        return "custom";
      }
    });
  },
  contextWindowSize: 1,
});
```

(and: swapping ~~`tokens: string[]`~~ for `tokens` in both examples also type checks without errors on this PR)

wooorm-arcjet:
Sweet thanks! #1455 is not closed *yet* with this PR. I meant to say that, that bug, was similarly present here and is no longer here.

There are several similar types:

* https://redirect.github.com/arcjet/arcjet-js/blob/cc76e1cf778e8d6d69a3023d61164908d39299ce/redact-wasm/index.ts#L56
* https://redirect.github.com/arcjet/arcjet-js/blob/cc76e1cf778e8d6d69a3023d61164908d39299ce/redact/index.ts#L16
* https://redirect.github.com/arcjet/arcjet-js/blob/cc76e1cf778e8d6d69a3023d61164908d39299ce/analyze-wasm/types.ts#L41
* https://redirect.github.com/arcjet/arcjet-js/blob/cc76e1cf778e8d6d69a3023d61164908d39299ce/arcjet/index.ts#L864

Some repetition is not bad per se, but I think it could be in 2 places: the rule is slightly different from manually calling `redact`. Or maybe 1 place. We’ll see. So a follow up PR would close #1455!

wooorm-arcjet:
Updates the OP to make more sense. Merging, thanks for the review!

PR #4884, opened 2025-08-11
chore(protocol): `latitude` and `longitude` exist together

wooorm-arcjet:
While reading through this, I assumed this was a bug. I cannot imagine a `latitude` (and `accuracyRadius`) to exist but no `longitude`, and inverse?
Especially because these *are* checking for `accuracyRadius`, and there is also a `hasAccuracyRadius` which *does* have a predicate that if it exists,
both `latitude` and `longitude` exist too.

This is also what I saw in the data,
though I have not seen too much data.

### DIFF
--- a/protocol/index.ts
+++ b/protocol/index.ts
@@ -878,23 +878,29 @@ export class ArcjetIpDetails {
 
   /**
    * Check if the IP address has geo `latitude` info.
-   * This also implies that `accuracyRadius` is available.
+   * This also implies that `accuracyRadius` and `longitude` are available.
    *
    * @returns
    *   Whether the IP address has latitude info.
    */
-  hasLatitude(): this is RequiredProps<this, "latitude" | "accuracyRadius"> {
+  hasLatitude(): this is RequiredProps<
+    this,
+    "accuracyRadius" | "latitude" | "longitude"
+  > {
     return typeof this.latitude !== "undefined";
   }
 
   /**
    * Check if the IP address has geo `longitude` info.
-   * This also implies that `accuracyRadius` is available.
+   * This also implies that `accuracyRadius` and `latitude` are available.
    *
    * @returns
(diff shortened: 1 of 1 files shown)


qw-in:
I don't have enough context to give a helpful review here, sorry

arcjet-rei (commented):
We should probably offer a `hasCoordinates` method that presumes the existence of `this { latitude, longitude, accuracyRadius }`. Since we don't have a way of knowing how  / if the existing methods are in use, we can't remove them, but also we don't know what use cases people might have for `hasLatitude` and `hasLongitude`, and I'm not sure what value we're adding by ensuring that both are present if one is.

I disagree that the current behavior is a bug, unless the bug is that we're offering these methods at all.

wooorm-arcjet:
> I disagree that the current behavior is a bug

Did you see the comment about `hasAccuracyRadius` which says that if `accuracyRadius` exists, both `latitude` and `longitude` exist too? These two functions *also* say that `accuracyRadius` exists.
That seems like an impossible situation to me?

> the bug is that we're offering these methods at all.

Right

blaine-arcjet (changes requested):
As you've previously stated, the connection between type assertions is not guaranteed. If you are going to do this, then you need to check all values. It seems better to implement the extra function that @arcjet-rei mentioned and check all 3 values.

wooorm-arcjet:
The current code, which you wrote, doesn’t have those checks for all values.
Which is why there are currently runtime bugs. For example, not all `asn` fields are present.
Changing runtime code and new functions is a separate discussion and PR.
For example, I prefer deprecating all this unneeded API surface.

wooorm-arcjet:
As this is turning into a bigger discussion, I marked this as WIP and made it into a discussion over at https://github.com/arcjet/arcjet-js/issues/5087.

PR #4889, opened 2025-08-11
chore(protocol): add `hasAsn`, deprecate `hasASN`

wooorm-arcjet:
There is also an `isVpn` here.
This lets people who see that casing use the same casing for ASN information.

Related-to: GH-4478.

### DIFF
--- a/protocol/index.ts
+++ b/protocol/index.ts
@@ -1145,7 +1145,7 @@ export class ArcjetIpDetails {
     asnType: string;
     asn: string;
   } {
-    return typeof this.asn !== "undefined";
+    return this.hasAsn();
   }
 
   /**


protocol/index.ts
```diff
@@ -1129,14 +1129,33 @@ export class ArcjetIpDetails {
...
+   *   Whether the IP address has ASN info.
+   */
+  hasASN(): this is {
+    asnCountry: string;
+    asnDomain: string;
+    asnName: string;
+    asnType: string;
+    asn: string;
+  } {
+    return typeof this.asn !== "undefined";
```
qw-in: Why not call the other method?

qw-in (approved):

protocol/index.ts
wooorm-arcjet: ohhh, yeah, also possible. I did that!

PR #4913, opened 2025-08-12
fix(protocol): Remove change that disjoined IP details and assertions (#4859)

blaine-arcjet:
This reverts commit 0a638ae8e2d680fbe3c33556f3011f4ad733f5ad.

As stated at https://github.com/arcjet/arcjet-js/pull/4859#issuecomment-3181025930, this change made the assertions disjoint from the initial definition. This means that a change in the definition is not reflected in TypeScript failing.

### DIFF
--- a/protocol/index.ts
+++ b/protocol/index.ts
@@ -8,7 +8,9 @@ export type * from "./well-known-bots.js";
 // Re-export the Bot categories from the generated file
 export { categories as botCategories } from "./well-known-bots.js";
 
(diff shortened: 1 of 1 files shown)


blaine-arcjet:
The build failures were due to https://github.com/arcjet/arcjet-js/pull/4722 which both @qw-in and I had a problem with, but it landed nonetheless. I've changed the helper to avoid reverting that change also, which means duplicating some code that was contained inside the TS stdlib but achieves the same goal.

arcjet-rei (approved):
Tests passing, so this LTGM now. I appreciate the minimal change to avoid having a multiple revert. 🚢

PR #4971, opened 2025-08-20
ci: lock actions to shas

wooorm-arcjet:
We now [block unpinned actions](https://github.blog/changelog/2025-08-15-github-actions-policy-now-supports-blocking-and-sha-pinning-actions/), so this pins the remaining ones.


davidmytton (approved):
LGTM. Will also manually merge because Trunk won't be able to do it.

PR #4981, opened 2025-08-21
feat(env): add support for trailing slash

wooorm-arcjet:
@morekid ran into an issue where a trailing slash was not allowed.
This surfaced because now we are not allowing unknown urls in development.
Perhaps that is a good thing.

Looking through some of the examples, I did see more trailing slashes: https://github.com/search?q=org%3Aarcjet%20ARCJET_BASE_URL&type=code.

We can either change those, or allow a slash, which is this PR.


davidmytton:
This is fine by me. I'm not sure how the port appeared in the local URL though - it shouldn't be needed because https://decide.arcjet.orb.local is a load balancer than proxies to the API. So I'd suggest removing the port number as well as allowing slashes.

wooorm-arcjet:
I tracked it down to always being there: from https://github.com/arcjet/arcjet-js/pull/83.

On allowing ports: can we accept *any* port on these known allowed URLs? (I think so, but, asking to confirm).

blaine-arcjet:
I did also see a few references with the trailing slash in the search that @wooorm-arcjet linked. I think it is fine to allow these, but do we want to just add the trailing-slash items to the list instead of writing logic around it?

blaine-arcjet:
I attached #4984 to this PR. However, I believe we need to verify that Connect RPC does the normalization and discuss whether the code or expanded allowlist should be used.

davidmytton:
We only support HTTPS (443) in prod and the Orbstack setup forwards HTTPS correctly, so my suggestion is to mirror prod, and remove the port.

For trailing slash, I like being flexible for users. Even if we document everything without a trailing slash (my preference), it's an easy mistake to make we should allow.

wooorm-arcjet:
> However, I believe we need to verify that Connect RPC does the normalization

These values with trailing slashes, documented in several places, did work before, so that suggests to me that it works.
The normalization in connect is at:

https://github.com/connectrpc/connect-es/blob/05a426bdede05630c0253d0e0318eaab4348dde5/packages/connect/src/protocol/create-method-url.ts#L38

Tested at:

https://github.com/connectrpc/connect-es/blob/05a426bdede05630c0253d0e0318eaab4348dde5/packages/connect/src/protocol/create-method-url.spec.ts#L51

blaine-arcjet:
Based on @arcjet-rei's approval on #5035, I think we're going with that one. I'll close this out.

PR #4997, opened 2025-08-26
docs(arcjet): add JSDocs

wooorm-arcjet:
Closes GH-4962.

This adds JSDocs to all types that are exposed.
It importantly documents all fields, parameters, etcetera on those types too.

This is useful if one for example writes `detectBot({allow: ""})`, where TypeScript now shows documentation for that allow option. To not be too verbose, functions themselves only describe their parameters, not the fields *in* those parameters.

### DIFF
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -455,81 +455,397 @@ const validateShieldOptions = createValidator({
   validations: [{ key: "mode", required: false, validate: validateMode }],
 });
 
+/**
+ * Configuration for the token bucket rate limit rule.
+ *
+ * @template Characteristics
+ *   Characteristics to track a user by.
+ */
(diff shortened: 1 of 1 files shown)


arcjet-rei (approved):
This looks fine to me. The completion for the various options / properties works for me with my LSP, but with these changes it can be a little challenging to read straight through the source files, so you might want to include explicit cross-references to the various configuration types.

wooorm-arcjet:
Thanks for the review!

> so you might want to include explicit cross-references to the various configuration types.

Can you expand on this? Are you thinking of referencing from an `AllowOptions` to a `DenyOptions`?

wooorm-arcjet:
Thanks again for the review Rei, landed! Do let me know what type of cross-references you’re thinking of!

Issue #4808, opened 2025-08-06
Support overriding `ip` workaround


wooorm-arcjet:
There’s a tip on the website here to allow overriding an IP for IP info in dev: https://docs.arcjet.com/troubleshooting#override-development-ip.

From some testing and a quick glance at the code in adapters for `findIp`, that seems to not work in astro, bun, deno, fastify, node, remix, sveltekit

blaine-arcjet:
The docs are wrong here. This override behavior is not a desired design goal and is a workaround on platforms that provide an `ip` fields on their "request" object.

davidmytton:
I'm reopening this as a different name because it's useful to be able to get IP data back in development. Overriding the IP is a workaround to achieve this right now, but as noted it only works in specific adapters. We should have a better way to do this so users can see exactly what data they'll get back from IP analysis.

blaine-arcjet:
@davidmytton Is this something that would exist in the SDK or would it be an API route or some tool in the dashboard?

qw-in:
In my 1:1 with @blaine-arcjet we had a quick chat about potential solutions. We considered an environment variable but landed on something like an `x-arcjet-ip-override: 1.1.1.1` header being the most ergonomic and flexible. It can be configured per-request but doesn't change the SDK API between development & production the same way setting `request.ip` does.

A downside is that this can be difficult to set in browsers but I still think it is overall the most flexible approach. [Playwright already supports adding additional headers](https://playwright.dev/docs/api/class-browsercontext#browser-context-set-extra-http-headers) and I'm sure there are browser settings or extensions that could do the same.

A potential follow on idea could be to add an Arcjet integration that supports this to the [Astro dev toolbar](https://docs.astro.build/en/guides/dev-toolbar/) or even build an Arcjet chrome extension (probably a devtools panel).

Implementation note from @blaine-arcjet is that this should not be included in `@arcjet/ip`. Instead each adapter would check `ARCJET_ENV === "development" && headers.has("x-arcjet-ip-override")` and short-circuit the ip lookup call.

wooorm-arcjet:
Given the explanation of why this is needed, on the website: “In local development, Arcjet defaults to `127.0.0.1` so the IP analysis metadata will be empty.”

(and the code, which is explained above [here](https://github.com/arcjet/arcjet-js/blob/3e601d7929c5e3a13160a9c61925465887a80a7b/arcjet-fastify/index.ts#L307-L315))

[...]

PR #5019, opened 2025-08-27
fix: support passing `ip` to `protect` in more integrations

wooorm-arcjet:
This adds support in types and in actual code for passing `ip` in `request` to `protect`.
This applies to the astro, bun, deno, fastify, and node integrations. The nest and next integrations already support this. Remix and Sveltekit have custom ways to pass IPs in and do not receive a regular `request` so I left them as-is.

This method is documented on the website at: <https://docs.arcjet.com/troubleshooting#override-development-ip>.

Closes GH-4808.

### DIFF
--- a/arcjet-astro/internal.ts
+++ b/arcjet-astro/internal.ts
@@ -146,6 +146,19 @@ export type ArcjetOptions<
 >;
 
 /**
+ * Request object that also supports overriding IP addressess for Arcjet’s IP
+ * geolocation and VPN and proxy detection.
+ *
+ * @link https://docs.arcjet.com/troubleshooting#override-development-ip
+ */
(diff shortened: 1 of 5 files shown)


arcjet-astro/internal.ts
```diff
@@ -229,7 +242,7 @@ export function createArcjetClient<
     const url = new URL(request.url);
     let ip = findIp(
       {
-        ip: clientAddress,
+        ip: request.ip || clientAddress,
```
qw-in, line 245: Should this only happen in development?

blaine-arcjet:
As mentioned on #4808. The docs are documenting a workaround and not a desired feature.

Issue #4561, opened 2025-07-04
Dry run of `sensitiveInfo` results in `DENY`


wooorm-arcjet:
> This is a bug, DRY_RUN mode should never produce a DENY decision.
> — @blaine-arcjet in https://github.com/arcjet/arcjet-js/pull/4550#discussion_r2183784972

From the test case:

```js
    const [rule] = sensitiveInfo({ allow: [], mode: "DRY_RUN" });
    const result = await rule.protect(
      {
        ...exampleContext,
        async getBody() {
          return "127.0.0.1 test@example.com 4242424242424242 +353 87 123 4567";
        },
      },
      exampleDetails,
    );
    assert.equal(result.conclusion, "DENY");
```

blaine-arcjet:
I've dug into this some more and it looks like the DRY_RUN override happens in the core logic. This allows the Rule Results to provide their actual decision while the top-level conclusion is still `ALLOW`. I *think* this is the correct behavior because users may want to inspect dry run executions to figure out if they would have triggered a DENY (perhaps via a logging platform).

wooorm-arcjet:
That *does* sound correct.
That behavior does only happen here for the local results.
Are you confident the decide api has the same behavior?

blaine-arcjet:
> Are you confident the decide api has the same behavior?

I am not confident. I remember changing something at one point, but don't remember what/why. Please open an internal issue to evaluate/revisit the Decide service works in regard to this 🙏

wooorm-arcjet:
Done, closing this here!

PR #5020, opened 2025-08-27
fix(arcjet): do not deny a dry run

wooorm-arcjet:
The local bot, email, and sensitive info rules can return a deny. Regardless of whether `mode` is `"DRY_RUN"`.
That should not happen.

Closes GH-4561.


blaine-arcjet:
Drafting this due to some further investigation.

Issue #4560, opened 2025-07-04
Allow no options for rules, default to dry run


wooorm-arcjet:
> I don't have a strong opinion on this. We could make all rules accept no options and then default to like DRY_RUN + "allow none". Do you want to make a follow up issue so other folks can chime in?
> — @blaine-arcjet in https://github.com/arcjet/arcjet-js/pull/4550#discussion_r2183779707

Obviously, have to check per rule to make sure if that really makes sense.

(Also, this was born out of allowing `shield()` if `shield({})` is allowed in the types, may be possible to change just that).

davidmytton:
Agreed on that being a default. It's what the user expects, but doesn't suddenly cause harm because it's in dry run mode.

blaine-arcjet:
Shield is going to get configuration in the future and then this value will be required like all other rules. This would become a breaking change when we do that. To avoid that breaking change in the future, I think we should keep it requiring an empty object if you don't specify a mode.

PR #5021, opened 2025-08-27
feat(arcjet): allow calling `shield` w/o options

wooorm-arcjet:
The shield rule accepts an options object with a single optional field `mode`, defaulting to `DRY_RUN`.
As all fields in the options object are optional, humans who want that, now have to pass an empty object: `shield({})`.
It is nicer for humans to be able to omit that and write `shield()`.

Closes GH-4560.


blaine-arcjet:
As I mentioned on #4560, we don't want to make a change that will cause a breaking change in the future, even if it makes the rule definition awkward currently.

Issue #4591, opened 2025-07-10
Deprecate `@arcjet/tsconfig`


wooorm-arcjet:
Issue title is what I currently think is the best solution. But first there’s some things to discuss. Alternative solutions are possible.

We currently publish `@arcjet/tsconfig` from here, the JavaScript SDK.
It contains 3 configs: `base`, `react-library`, and `nextjs`.
We use `base` here in the JS SDK.
The monorepo and docs and separate examples sometimes use `base` and sometimes `nextjs`.
`react-library` is as far as I can see unused.
There are examples here in the JS SDK, and probably in other places that do not use these configs but use entirely custom ones.

There’s I think some problems with that:

1. I cannot make the `tsconfig` stricter here, that would affect examples and monorepo that need to then loosen (them being stricter is probably a good idea, but that has different priorities than the JS SDK being stricter).
2. `react-library` and `nextjs` tsconfigs are maintained here but not used or tested here, only used in other places
3. a public package `@arcjet/tsconfig`, `@astro/tsconfig`, `@nextjs/tsconfig`, to users *sounds* like they should use it and extend from it in their arcjet/astro/nextjs projects. But this package is, as far as I know, not recommended for people to install

It is of course nice to not have duplication of tsconfigs everywhere. But I think that can be solved in other ways.

wooorm-arcjet:
Related to GH-4590.

PR #5022, opened 2025-08-27
chore(tsconfig): remove `@arcjet/tsconfig`

wooorm-arcjet:
See the related issue for more reasoning.

After this lands it remains to deprecate the package on npm. That can be done when the next SDK is released.
At that same version, we can publish an empty package and immediately deprecate it.
If anyone uses it, they will not automatically pull it in with semver (because we’re at beta releases) but when they pull it in they will see the deprecation.

Closes GH-4591.


qw-in (changes requested):
Looks like bun is unhappy:

Image

wooorm-arcjet:
@qw-in Bun now passes!

PR #5035, opened 2025-08-28
feat(env): Support trailing slashes on ARCJET_BASE_URL

### LINKED ISSUE #4984: ARCJET_BASE_URL should allow trailing slash
Assuming that Connect RPC normalizes baseUrls that have a trailing-slash, we should allow users to specify URLs with trailing slashes.

We could write code to trim a trailing slash or we could expand the allowlist to include trailing-slash URLs.

blaine-arcjet:
Closes #4984

Alternative to #4981

I prefer less moving parts and am hesitant to normalize the environment variable the user gives us (instead deferring that to Connect RPC). Given that, I'd rather we enumerate all the URLs that we support with and without a trailing slash.

### DIFF
--- a/env/index.ts
+++ b/env/index.ts
@@ -117,6 +117,12 @@ const baseUrlAllowed = [
   "https://fly.decide.arcjet.com",
   "https://fly.decide.arcjettest.com",
   "https://decide.arcjet.orb.local",
+  // Allow trailing slashes
+  "https://decide.arcjet.com/",
+  "https://decide.arcjettest.com/",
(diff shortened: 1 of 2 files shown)


wooorm-arcjet:
Either are fine to me, but I do slightly prefer the other, it makes sure we always have a variable without slash, regardless of whether Connect or an alternative supports it. And, dryer code.

arcjet-rei (approved):
Explicit, includes tests, LGTM :shipit:

PR #5038, opened 2025-08-29
docs(next): refactor JSDocs

wooorm-arcjet:
This mostly refactors the existing JSDocs to have similar information to all other integrations.

Closes GH-4958.

### DIFF
--- a/arcjet-next/index.ts
+++ b/arcjet-next/index.ts
@@ -1,4 +1,4 @@
-import type { NextApiResponse } from "next";
+import type { NextApiRequest, NextApiResponse } from "next";
 import { NextResponse } from "next/server.js";
 import { headers, cookies } from "next/headers.js";
 import type {
@@ -11,6 +11,7 @@ import core from "arcjet";
 import type {
   ArcjetDecision,
   ArcjetOptions as CoreOptions,
+  ArcjetRuleResult,
   Primitive,
   Product,
   ArcjetRequest,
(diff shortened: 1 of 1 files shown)


wooorm-arcjet:
@qw-in I noticed that `createMiddleware` is defined here but seems to be undocumented on the website.
Should it be there? Should it be deprecated here?

arcjet-next/index.ts
```diff
@@ -331,62 +434,69 @@ export interface ArcjetNext<Props extends PlainObject> {
...
+   *         mode: "LIVE",
+   *         max: 100,
+   *         window: "60s",
+   *       }),
+   *     );
+   *
+   *   export async function GET(req: NextRequest) {
+   *     const decision = await aj.protect(req);
+   *     if (decision.isDenied()) {
+   *       throw new Error("Forbidden");
```
qw-in, line 494: Should this not be a returned response? Unfortunately the `forbidden` helper is still experimental

wooorm-arcjet:
@qw-in responding on the [review note](https://github.com/arcjet/arcjet-js/pull/5038#discussion_r2316544659) here so it remains visible while I merge:

Looked into it a little, but didn’t find a clear answer. For server components, the recommendation is to return different things, but they don’t have access to `response`: https://nextjs.org/docs/app/getting-started/error-handling#server-components. In the future, `forbidden` sounds good, or otherwise, perhaps we need to make the example bigger (showing the client handling too) and verify that it properly works.

Logged as https://github.com/arcjet/arcjet-js/issues/5057

qw-in:
> createMiddleware is defined here but seems to be undocumented on the website.
Should it be there? Should it be deprecated here?

Sorry I completely missed this somehow! It predates my time at Arcjet but I suspect it should be added to the docs

blaine-arcjet:
> Should this not be a returned response? Unfortunately the forbidden helper is still experimental

I believe @qw-in is correct with that example since it is an API handler. Throw only really works in the case of Server Actions.

PR #5039, opened 2025-08-29
docs(node): add JSDocs

wooorm-arcjet:
Closes GH-4959.

### DIFF
--- a/arcjet-node/index.ts
+++ b/arcjet-node/index.ts
@@ -110,18 +110,36 @@ type PlainObject = {
   [key: string]: unknown;
 };
 
+/**
+ * Configuration for {@linkcode createRemoteClient}.
+ */
 export type RemoteClientOptions = {
+  /**
+   * Base URI for HTTP requests to Decide API (optional).
+   *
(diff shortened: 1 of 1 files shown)


arcjet-node/index.ts
```diff
@@ -110,18 +110,36 @@ type PlainObject = {
...
+   * is known and allowed) and the standard production API otherwise.
+   */
   baseUrl?: string;
+
+  /**
+   * Timeout in milliseconds for the Decide API (optional).
+   *
+   * Defaults to `500` in production and `1000` in development.
+   */
   timeout?: number;
```
qw-in, line 130: Out of scope opinion:

We should rename to `timeoutMs`

qw-in (approved):

arcjet-node/index.ts
blaine-arcjet: For the sake of argument, this should probably take a string representing a duration (`500ms`, `1s`) which is what we support on rate limit configuration.

wooorm-arcjet: Logged this as https://github.com/arcjet/arcjet-js/issues/5056

Issue #5016, opened 2025-08-27
astro: types of custom (serializable) rules should match `arcjet` core rules


wooorm-arcjet:
Take for example:

https://github.com/arcjet/arcjet-js/blob/3e601d7929c5e3a13160a9c61925465887a80a7b/arcjet-astro/index.ts#L278-L285

The `options` parameters are marked as optional in `arcjet-astro/index.ts`.

However, they are required in `arcjet/index.ts`:

https://github.com/arcjet/arcjet-js/blob/3e601d7929c5e3a13160a9c61925465887a80a7b/arcjet/index.ts#L1889

The solution I believe, to prevent people from making mistakes, is to remove the `?`s in `arcjet/index.ts`.

blaine-arcjet:
I think I had a reason for this but cannot remember. However, I believe the option validation is handled by Astro (via Zod schema validation), which is the something we want to retain since users don't need to use TypeScript with Astro.

PR #5052, opened 2025-09-01
fix(astro): fix types, zod validation

wooorm-arcjet:
Previously, in the types, `options` were optional, but they are required.

Zod validation additionally marked the *entire* options objects as optional. But it is instead *fields* that are (sometimes) optional.

To make sure the types are correct, I used `satisfies IntegrationRule<>`. To make sure Zod is correct, you can use something like:

```ts
const example: z.infer<typeof validateSlidingWindowOptions> = {
  interval: 123
  max: 123,
};
```

I verified with the Astro example, and noticed that it said every email I provided was invalid. That led me to think something was wrong. So instead of denying most emails with an empty `allow: []`, this now allows free emails (such as gmail) and emails w/o gravatar (such as my work email).

Closes GH-5016.


qw-in:
How difficult would it be to add tests for this?

wooorm-arcjet:
Good to have tests for adapters. That was tracked in GH-769. I made subissues for each adapter to track the work, for Astro, that’s GH-5060. And, made a PR to solve it, GH-5069. When that lands, I can show that this PR improves things.

wooorm-arcjet:
This is now exhaustively tested!

arcjet-astro/index.ts
```diff
@@ -435,8 +430,10 @@ export function filter(options: FilterOptions) {
  * @returns
  *   Astro integration Sensitive information rule to provide to the SDK in the `rules` field.
  */
-export function sensitiveInfo(options?: SensitiveInfoOptions<never>) {
-  return { type: "sensitiveInfo", options } as const;
+export function sensitiveInfo(options: SensitiveInfoOptions<never>) {
+  return { type: "sensitiveInfo", options } as const satisfies IntegrationRule<
+    Array<string>
```
qw-in, line 437: Typically I've been seeing this syntax in this repo

```suggestion
    string[]
```

arcjet-astro/index.ts
```diff
@@ -20,145 +20,132 @@ const validateProxies = z.array(z.string());
...
   .object({
-    baseUrl: z.string(),
-    timeout: z.number(),
+    baseUrl: z.string().optional(),
+    timeout: z.number().optional(),
   })
   .strict()
   .optional();
 
 const validateShieldOptions = z
```
qw-in, line 31: These could potentially be coupled together a bit too:

```ts
const shieldOptionsSchema = z
  .object({
      mode: validateMode.optional(),
  })
  .strict() satisfies z.ZodType<ShieldOptions>;
```

See [playground](https://www.typescriptlang.org/play/?ssl=13&ssc=34&pln=1&pc=1#code/JYWwDg9gTgLgBAbzgLzgXzgMyhEcBEyEAJvgNwBQFMAnmAKZwCyNASvQMbTFwC8iFOHAB2AQxD0AXHADOMKMGEBzSmir0AHpFhwuwuczaduAZQ4ALeiFEBxCCT4oAdBABGAK04wAFAkEjxKWc5BWVvAEoAGgo0cNlRGGAZTGB6GWcALRIAFTp6AB4Wdi4oYgA+SgpNbXg9AyLjUrNLawAhUR5+ZBcPL19-GVx6CBhLKGlukMUlCOjY+MTk1PTurOJchkKjEvKyIA)

arcjet-astro/index.ts
```diff
@@ -350,8 +337,10 @@ function integrationRuleToClientRule<Characteristics extends readonly string[]>(
  * @returns
  *   Astro integration Shield rule to provide to the SDK in the `rules` field.
  */
-export function shield(options?: ShieldOptions) {
-  return { type: "shield", options } as const;
+export function shield(options: ShieldOptions) {
+  return { type: "shield", options } as const satisfies IntegrationRule<
+    Array<string>
+  >;
 }
```
qw-in, line 346: Does it make more sense to be explicit about these types here? `satisfies` could "leak" accidental extra fields into the function signatures

```suggestion
export function shield(options: ShieldOptions): IntegrationRule<{ type: shield; options: ShieldOptions }> {
  return { type: "shield", options };
}
```

arcjet-astro/index.ts
```diff
@@ -20,145 +20,132 @@ const validateProxies = z.array(z.string());
...
   .object({
     rateLimit: validateSlidingWindowOptions,
     bots: validateBotOptions,
     email: validateEmailOptions,
   })
-  .strict()
-  .optional();
+  .strict();
 
 type IntegrationRule<Characteristics extends readonly string[]> =
```
qw-in, line 139: Probably out of scope for this PR but a slightly odd type to my 👀.

I wonder about:

```ts
type ShieldRule = { type: "shield"; options: ShieldOptions };
type RateLimitRule<TCharacteristics extends /* ... */> = { /* ... */ } | { /* ... */ };
type IntegrationRule<TCharacteristics extends /* ... */> = 
  | ShieldRule
  | RateLimitRule<TCharacteristics>;
```

Then lower down `shield` could return `ShieldRule` without any `Array<string>` I think.

Oh, and maybe namespacing them all makes sense: `AstroShieldRule`

arcjet-astro/index.ts
wooorm-arcjet: I saw both and tried to stick with one, my somewhat strong preference is `Array<string>`:

* reads naturally to me, “array of strings”
* same as above especially with `readonly string[]` vs `ReadonlyArray<string>`: the array is readonly, not the string
* `[`, `]`  are reused for tuples in TS, which are a nuanced difference from regular arrays, so instead I like being explicit
* uses regular type parameter syntax, more similar to things like `Set<string>`, `Map<string, string>`, etc

It would be nice to have more rules for such things tho!
If it is intentional to use `[]` syntax then that’s also okay!

And, type parameters can have defaults too, so as `Characteristics extends Array<string>`, we could also have: `Characteristics extends Array<string> = Array<string>`, then we can omit those here!

arcjet-astro/index.ts
wooorm-arcjet: Great idea! But `exactOptionalProperties` prevents that currently, because `mode` can be explicit `undefined` here but not in `arcjet` core yet:

https://github.com/arcjet/arcjet-js/blob/e6d9980977af2df067df9ce774e9bee04aa87d69/arcjet/tsconfig.json#L3-L4.

Added a note to do this!

arcjet-astro/index.ts
wooorm-arcjet: Yes, indeed, I saw some further room for improvement there!

arcjet-astro/index.ts
wooorm-arcjet: Wait, isn‘t the point of `satisfies` that it doesn’t do that?

Yes, indeed, I saw this too. Having a type for an explicit result (instead of having that source of truth in the big enum but reusing those types in the enum). Using that type here as return types. Then no `satisfies` and no `const`s are needed here!

wooorm-arcjet:
Thanks for the reviews everyone!

Issue #31, opened 2023-11-06
Figure out if we can avoid cast in `@arcjet/next`


blaine-arcjet:
TypeScript seems to get confused in the `protect()` function of `@arcjet/next` because we are manipulating generics. We currently just cast to an `ArcjetRequest<Rules>` but we should investigate if we can avoid the cast.

PR #5054, opened 2025-09-02
refactor: simplify `withClient` types in adapters

wooorm-arcjet:
This removes a gnarly type problem, which was solved with a cast, by simplifying the types and well, sadly, still having a type cast.

But the new type cast is on a much smaller value, the default. And it kinda makes sense that TS says `{}` is not `Props`. And the rest of the code is simpler now anyway, so I think it’s good to have.

(I think this is enough to close 31)

Closes GH-31.

### DIFF
--- a/arcjet-astro/internal.ts
+++ b/arcjet-astro/internal.ts
@@ -184,11 +184,11 @@ export interface ArcjetAstro<Props extends PlainObject> {
    * This will analyze the request locally where possible and otherwise call
    * the Arcjet decision API.
    *
-   * @param ctx
-   *   Additional context for this function call.
    * @param request
    *   Details about the {@linkcode ArcjetRequest} that Arcjet needs to make a
    *   decision.
+   * @param rest
+   *   Additional properties required for running rules against a request.
    * @returns
    *   Promise that resolves to an {@linkcode ArcjetDecision} indicating
    *   Arcjet’s decision about the request.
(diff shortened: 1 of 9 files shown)


arcjet-rei:
Because I keep bringing it up in standup, I took another look at this PR today, and also discussed it a bit with @blaine-arcjet. Having dug into it a bit more, this is a difficult change to evaluate definitively.

- As @wooorm-arcjet mentions, it does simplify the types, but…
- it still includes a cast (which is, as @wooorm-arcjet mentions, a much narrower cast).
- While #31 is focused on the cast to `ArcjetRequest<Rules>`, the fact that there _is_ still a cast still feels a little janky.
- It is difficult to test changes that are more or less entirely related to (inferred) types in signatures.
- As such, it's difficult to gauge whether this is a breaking change for downstream users.

Since it _is_ a potential breaking change, it seems to me that unless we can remove the cast altogether, this isn't a substantial enough improvement to justify the time necessary to establish that the change isn't going to break things for users.

I think the removal of the hack and the overall simplification of the signatures makes this a change worth keeping, but I don't think it's ready to merge until we can come up with a way to eliminate all casts. As such, I'm going to convert this to a draft for now (so it doesn't keep coming up in reviews), but I will also noodle on it some more and see if I can come up with a solution that preserves the simplification, eliminates the cast, and keeps the type checker happy.

Issue #759, opened 2024-05-15
Support for Cloudflare Workers


davidmytton:
Allow us to support applications built on Cloudflare workers. There's a lot of useful data in the request context [`cf` property](https://developers.cloudflare.com/workers/runtime-apis/request/#the-cf-property-requestinitcfproperties) we could make use of.

blaine-arcjet:
This is blocked by Connect RPC's requirement to construct a client with a baseUrl. We want to avoid reconstructing a client on each request to the worker.

blaine-arcjet:
Related: Connect doesn't work well on Cloudflare, see https://github.com/connectrpc/connect-es/issues/577

davidmytton:
I spent a bit of time researching where we're at with support for Cloudflare workers. 

> This is blocked by Connect RPC's requirement to construct a client with a baseUrl. We want to avoid reconstructing a client on each request to the worker.

Seems there's a few reasons for this:

1. We want to load our WASM module into the worker outside of the request context so it can be instantiated once rather than on every request. This is why we recommend creating the Arcjet client outside of the request handler for our other adapters.
2. With https://github.com/arcjet/arcjet-js/pull/1201 we'd like to have a persistent connection to improve latency across requests.
3. The DX is better when you can define base rules in a client and then `withRule` any additional rules you might want to combine.

With Cloudflare's [move from Service Workers to ES Modules](https://developers.cloudflare.com/workers/reference/migrate-to-module-workers/), environment variables are no longer provided in the global scope. Instead, they're [passed as a binding to the handler](https://developers.cloudflare.com/workers/reference/migrate-to-module-workers/#environment-variables-in-es-modules-format). Trying to access them as a global will error.

Also, creating the connect client requires a base URL, which we configure dynamically based on the environment variables.

**Loading WASM**

Cloudflare officially supports WASM in workers. They have [some history with slow worker startup](https://community.cloudflare.com/t/fixed-cloudflare-workers-slow-with-moderate-sized-webassembly-bindings/184668), but it seems they've implemented a fix for this. [From 2020](https://community.cloudflare.com/t/fixed-cloudflare-workers-slow-with-moderate-sized-webassembly-bindings/184668/16):

[...]

PR #5077, opened 2025-09-04
feat(stable-hash, transport): support `workerd`

wooorm-arcjet:
Related-to: GH-759.
Closes GH-5076.


blaine-arcjet (changes requested):
It's great to support the `workerd` target in our packages! This is a small step towards supporting Cloudflare Workers, even though it will take us a lot of time to make the changes necessary to fully support them due to the environment variable restrictions and inability to construct an SDK client outside of the handler.

I want us to have separate files for `edge-light` and `workerd` targets, even though it is a perforation of files.

With specific names, we can specialize them when the time comes to change whatever without needing to split the file at that time. It also allows the files to be named the same as the target instead of `dom` (we don't run/support the DOM even though you are referencing `fetch` from the WHATWG spec) and `edge` (too generic).

blaine-arcjet (approved):
Awesome!

Issue #5101, opened 2025-09-05
Use `jco-transpile`


wooorm-arcjet:
Since there is now a `jco-transpile` package with few dependencies (for 2 weeks), and that we’ve been stuck on an older `jco` to prevent package bloat, we can attempt to use it.

Related-to: bytecodealliance/jco#500.

Issue #4342, opened 2025-06-09
CI: Restrict jco


blaine-arcjet:
We stopped upgrading jco because they introduced a ton of dependencies for the ComponentizeJS feature. We should restrict this in dependabot until we figure out how to decouple their transpile & componentize code.

wooorm-arcjet:
It looks like there is already an ignore?

https://github.com/arcjet/arcjet-js/blob/9dd8eff3e14bf569492f11827ab778936e5a1041/.github/dependabot.yml#L41-L44

wooorm-arcjet:
If that doesn’t work, should it be `dependency-name: "@bytecodealliance/jco"`?

blaine-arcjet:
> If that doesn’t work, should it be dependency-name: "@bytecodealliance/jco"?

Yeah, you are correct that the name is wrong in the config file. It should be the fully scoped package name.

PR #5102, opened 2025-09-05
build(analyze-wasm, redact-wasm): use `jco-transpile`

wooorm-arcjet:
Since there is now a `jco-transpile` package with few dependencies (for 2 weeks), and that we’ve been stuck on an older `jco` to prevent package bloat, this PR attempts to unblock us by switching to the new package.

It seems to work! Their package is at `0.0.4` though, and the types a bit broken, but things pass. All options except for `no-wasi-shim` that were used are also in `jco-transpile`. I think that’s something that is wrapped around `transpile`. It could be some shim is injected though, I do not know enough about webassembly to verify whether some shim is in here.

Related-to: bytecodealliance/jco#500.
Related-to: GH-3446.
Related-to: GH-4342.
Closes GH-5101.

### DIFF
--- a/rollup-config/tsconfig.json
+++ b/rollup-config/tsconfig.json
@@ -3,7 +3,9 @@
     "checkJs": true,
     "declarationMap": true,
     "declaration": true,
-    "emitDeclarationOnly": true
+    "emitDeclarationOnly": true,
+    // Deno types cause problems.
+    "skipLibCheck": true
   },
   "extends": "../tsconfig.base.json",
   "include": ["index.js"]


wooorm-arcjet:
Ah, Next/Webpack fails on some code that’s generated:

```
const isNode = typeof process !== 'undefined' && process.versions && process.versions.node;
let _fs;
async function fetchCompile (url) {
  if (isNode) {
    _fs = _fs || await import('node:fs/promises');
    return WebAssembly.compile(await _fs.readFile(url));
  }
  return fetch(url).then(WebAssembly.compileStreaming);
}
```

🤔

(from: https://github.com/bytecodealliance/jco/blame/cbaacb8fc9395366702b0bcf8dc07a99c85bc1ce/crates/js-component-bindgen/src/intrinsics/mod.rs#L472)

wooorm-arcjet:
The transpile options today are:

```
 *   name: string,
 *   instantiation?: 'async' | 'sync',
 *   importBindings?: 'js' | 'optimized' | 'hybrid' | 'direct-optimized',
 *   map?: Record<string, string>,
 *   asyncMode?: string,
 *   asyncImports?: string[],
 *   asyncExports?: string[],
 *   asyncWasiImports?: string[],
 *   asyncWasiExports?: string[],
 *   validLiftingOptimization?: bool,
 *   tracing?: bool,
 *   nodejsCompat?: bool,
 *   tlaCompat?: bool,
 *   base64Cutoff?: bool,
 *   js?: bool,
 *   minify?: bool,
 *   optimize?: bool,
 *   namespacedExports?: bool,
 *   outDir?: string,
 *   multiMemory?: bool,
 *   experimentalIdlImports?: bool,
 *   optArgs?: string[],
```

So we could investigate whether we could use more of those

analyze-wasm/build.ts
blaine-arcjet, line 1: I haven't dug into this in-depth yet but my feeling was that we'd put this into our pipeline as a Rollup plugin.

wooorm-arcjet: I’m a bit 🤷 on rollup: it currently doesn’t do much, it helps with WebAssembly, injects the current version, and builds `.d.ts` files. We could do more there. Or, perhaps we can do without it, and cut those dependencies?

In this case, these 2 duplicate scripts only needs to run when generating new WebAssembly, right?
So it *could* be a single script in the top-level monorepo?

Generating the version in a file could also be a simple node script.

And TS can generate types too (and we soon wouldn’t need it for our tests)

blaine-arcjet: Rollup will continue to be used because it is the basis of the zero-dependency philosophy. We will remove all non-arcjet packages and then we will produce a single bundle for each SDK/Adapter that will contain the packages in this repo. Trying to move things out of the rollup pipeline will make this harder in the future.

analyze-wasm/.gitignore
```diff
@@ -130,6 +130,8 @@ dist
 .pnp.*
 
 # Generated files
+build.d.ts
+build.js
```
blaine-arcjet:
```suggestion
```

redact-wasm/.gitignore
```diff
@@ -130,6 +130,8 @@ dist
 .pnp.*
 
 # Generated files
+build.d.ts
+build.js
```
blaine-arcjet:
```suggestion
```

redact-wasm/wasm/arcjet_analyze_bindings_redact.component.js
```diff
@@ -25,310 +24,931 @@ function instantiate(getCoreModule, imports, instantiateCore = WebAssembly.insta
...
+    this.#lock = null;
+    existingLock.resolve(existingLock.ticket);
+  }
+  
+  isExclusivelyLocked() { return this.#lock !== null; }
+  
+}
+
+if (!Promise.withResolvers) {
+  Promise.withResolvers = () => {
```
blaine-arcjet, line 573: I don't think we want to shim `Promise` in Arcjet user's runtime. This may be a non-starter unless we fix upstream.

redact-wasm/wasm/arcjet_analyze_bindings_redact.component.js
```diff
@@ -25,310 +24,931 @@ function instantiate(getCoreModule, imports, instantiateCore = WebAssembly.insta
...
+      reject = rej;
+    });
+    return { promise, resolve, reject };
+  };
+}
+
+const _debugLog = (...args) => {
+  if (!globalThis?.process?.env?.JCO_DEBUG) { return; }
+  console.debug(...args);
+};
```
blaine-arcjet, line 587: I think this is going to cause warnings in Next.js targeting the Edge runtime. We had a problem before where static analysis was warning on any `process.env` access.

package-lock.json
```diff
@@ -789,193 +787,22 @@
...
-      "dev": true,
-      "license": "Apache-2.0",
-      "optional": true,
-      "os": [
-        "win32"
-      ],
-      "bin": {
-        "wizer-win32-x64": "wizer"
-      }
+    "node_modules/@bytecodealliance/jco-transpile/node_modules/@bytecodealliance/preview2-shim": {
```
blaine-arcjet: The wasi-p2 shim is still pulled in as a dependency so I worry that this is actually shimming WASI which we don't want to expose in our components since WASI allows for exiting the sandbox (which would be a security risk).

blaine-arcjet (commented):
I like the approach here. After reviewing the implementation, I wanted to look at the codegen since we've been behind and I think I found some non-starters. We will likely need to contribute fixes upstream before we can land this change if we don't want to apply more workarounds after transpile but before file-write.

Notably:
1. It looks like wasi-p2 shim is being applied. We'll need a way to disable.
2. The `_debugLog` looks at `process.env` which will need to be removed.
3. `Promise` is shimmed, which would modify Arcjet users' runtimes.
4. Titus mentioned the types are not generated as well as they used to be.
5. ...Maybe other stuff I missed

package-lock.json
wooorm-arcjet: Passing `wasiShim: false` does not affect the output.
That is likely because it seems to be off by default: https://github.com/bytecodealliance/jco/blob/b3bc6266cfa521093afe8e563670a0b39d35715d/packages/jco-transpile/src/transpile.ts#L236

redact-wasm/wasm/arcjet_analyze_bindings_redact.component.js
wooorm-arcjet: There seems to not be no option to turn debug logging off.
That could be a way to go.

Another one is, there is an option to turn off/on node compat. Defaulting to off, it injects [this code](https://github.com/bytecodealliance/jco/blob/b3bc6266cfa521093afe8e563670a0b39d35715d/crates/js-component-bindgen/src/intrinsics/mod.rs#L533) when on.
Perhaps, when rendering the [debug log](https://github.com/bytecodealliance/jco/blob/b3bc6266cfa521093afe8e563670a0b39d35715d/crates/js-component-bindgen/src/intrinsics/mod.rs#L665-L675), that option could be looked at, to not inject `process.env` stuff?

redact-wasm/wasm/arcjet_analyze_bindings_redact.component.js
wooorm-arcjet: Yes, I don’t see a reason why a library should do this.

[It](https://github.com/bytecodealliance/jco/blob/b3bc6266cfa521093afe8e563670a0b39d35715d/crates/js-component-bindgen/src/intrinsics/mod.rs#L677) could be turned into a ponyfill function. Or alternatively, an option to turn it off?

wooorm-arcjet: see https://github.com/bytecodealliance/jco/pull/1156

Issue #5107, opened 2025-09-05
Update `package-lock.json`s in examples


wooorm-arcjet:
We do use tools to update them, but those tools do not pick up changes in `file:` based links.
So, this has to be manual.

PR #5108, opened 2025-09-05
deps(example): update `package-lock.json`s

wooorm-arcjet:
Closes GH-5107.


arcjet-rei (approved):
Is there a reason that `examples/nextjs-app-dir-rate-limit` is still depending on `nextjs@15.4.0-canary.67"? Otherwise, this looks to me. It would be cool if we could configure Renovate (or Dependabot, I suppose, although that seems less likely) to deal with these `package-lock.json` updates automatically.

Everything else LGTM.

wooorm-arcjet:
Weird, I didn’t see that, good catch.

No clue why that’s there. Would you prefer me to unlock that here? Or unlock later?

arcjet-rei:
If there's not a reason for it to be like that, change it now, please and thank you!

blaine-arcjet:
> Is there a reason that examples/nextjs-app-dir-rate-limit is still depending on nextjs@15.4.0-canary.67?

Yes, this example uses an experimental Next.js feature which I don't think shipped yet. This canary needed to be updated and not replaced with a stable version.

PR #5111, opened 2025-09-05
deps(examples): update `tailwind` in `nextjs-authjs-5`

wooorm-arcjet:
Closes GH-5109.

/cc @qw-in would especially appreciate a look from you on this: I don’t know tailwind.
The migration tool here crashed, but in GH-5113 it mostly worked.

PR #5113, opened 2025-09-05
deps(examples): update `tailwind` in `nextjs-react-hook-form`

wooorm-arcjet:
Closes GH-5111.

/cc @qw-in would especially appreciate a look from you here too.

Note that here the tailwind migration did everything, except that it failed with peer dependencies, I had to install things manually.


examples/nextjs-react-hook-form/app/page.tsx
```diff
@@ -36,15 +36,15 @@ export default function IndexPage() {
...
+          className={buttonVariants({ variant: "outline-solid" })}
         >
           Arcjet Documentation
         </Link>
         <Link
           href="https://app.arcjet.com"
           target="_blank"
           rel="noreferrer"
-          className={buttonVariants({ variant: "outline" })}
+          className={buttonVariants({ variant: "outline-solid" })}
```
qw-in, line 47: The migration did this?

examples/nextjs-react-hook-form/package.json
```diff
@@ -14,15 +14,17 @@
     "@hookform/resolvers": "^5.2.1",
     "@radix-ui/react-label": "^2.1.7",
     "@radix-ui/react-slot": "^1.0.2",
+    "@tailwindcss/postcss": "^4.1.13",
     "class-variance-authority": "^0.7.1",
     "clsx": "^2.1.1",
-    "lucide-react": "0.536.0",
+    "eslint-plugin-tailwindcss": "^3.18.2",
```
qw-in, line 20: Why the move to dependencies?

examples/nextjs-react-hook-form/app/page.tsx
wooorm-arcjet: yeah!

examples/nextjs-react-hook-form/package.json
wooorm-arcjet: I think `npm install` does that, sorting.
It does adhere to tabs/spaces and such, but it does do sorting!

examples/nextjs-react-hook-form/app/page.tsx
wooorm-arcjet: well something did, I did *not* do that manually

wooorm-arcjet: it does break though, reverting.
I though migration without crashing was a good thing, but, it does things that don’t work :(

wooorm-arcjet:
There are no styles in this one, something doesn’t work. I don’t know what’s up with tailwind + radix + shadcn + next in this example.
My idea after fighting a whole day with dependencies, is to have less dependencies 😅
Perhaps someone else could try this one. Or, we could try and remove some things!

Issue #5114, opened 2025-09-05
`remix-express` example has no styles


wooorm-arcjet:
I was going through pinned dependencies, which often dealt with tailwind (e.g., https://github.com/arcjet/arcjet-js/pull/5113).

In the `remix-express` example, tailwind is also used, but already upgraded.
However, opening the example shows just a black screen with some white text.
No indication that there’s an input with text you can edit, and a button you can click.

I also noticed that 1Password tried to enter an email address, because the input has a `name="email"`.

I am unsure whether the example ever looked great: whether few styles are a bug or just never implemented.

PR #5115, opened 2025-09-05
fix(examples): add styles to `remix-express`

wooorm-arcjet:
* add styles
* use textarea instead of input, it’s not for just an IP address
* use a more wordy example instead of just an IP address
* use `name="text"` instead of `name="email"` so 1password and the like don’t attempt to autofill or store
* also attempted to update other dependencies, which didn’t work, but I made sure everything that’s possible is up to date, and left some notes

Closes GH-5114.

Before:

Screenshot 2025-09-05 at 4 08 41 PM

After:

Screenshot 2025-09-05 at 4 08 21 PM

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -559,11 +559,13 @@ updates:
     ignore:
       # Remix 2 doesn't support Express 5 so the dependencies below are
       # restricted.
+      # The peer dependency comes from `@remix-run/express`.
+      # See also: <https://github.com/remix-run/remix/discussions/10438>
       - dependency-name: express
         versions: [">=5"]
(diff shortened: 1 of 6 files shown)


qw-in:
cc @arcjet/design-team

morekid:
I've made some adjustments (https://github.com/arcjet/arcjet-js/pull/5115/commits/018552b668d89e24cd759162183d2f4228c96ec3) to align this to Arcjet's style as much as possible.
Go to go for me with this change.

image

image

wooorm-arcjet:
Thanks! Looks great to me ✅  but I cannot use the GH UI to ✅ as this is my PR, @qw-in?

PR #5170, opened 2025-09-11
feat(arcjet): add IP-related fields to filters

wooorm-arcjet:
Make sure to expand `arcjet/test/filter.test.ts`.

Closes arcjet/arcjet#5649.


protocol/index.ts
```diff
@@ -187,32 +187,53 @@ export class ArcjetReason {
...
 /**
  * Filter reason.
  */
 export class ArcjetFilterReason extends ArcjetReason {
   /**
-   * Expression that matched or `undefined` if no expression matched.
+   * Expressions that matched.
    */
-  expression?: string | undefined;
+  matchedExpressions;
```
qw-in, line 212: Should these have types? I assume currently inferred from the constructor

protocol/proto/decide/v1alpha1/decide_pb.d.ts
```diff
@@ -973,10 +973,27 @@ export declare class ShieldReason extends Message<ShieldReason> {
  */
 export declare class FilterReason extends Message<FilterReason> {
   /**
-   * @generated from field: string matched_expression = 1;
+   * Deprecated: Use the `matched_expressions` field instead.
```
qw-in, line 976:
```suggestion
   * Deprecated: Use the `matchedExpressions` field instead.
```

arcjet/test/filter.test.ts
```diff
@@ -353,6 +349,327 @@ test("expressions", async function (t) {
...
+        const [rule] = filter({ allow: ['ip.src.region == ""'] });
+        const result = await rule.protect(createContext(), createRequest());
+        assert.equal(result.conclusion, "ALLOW");
+        assert(result.reason.isFilter());
+        assert.deepEqual(result.reason.matchedExpressions, []);
+        assert.deepEqual(result.reason.undeterminedExpressions, [
+          'ip.src.region == ""',
+        ]);
+      });
+    });
```
qw-in, line 611: These tests look quite similar - thoughts on code sharing here?

protocol/index.ts
wooorm-arcjet: Inferred from the constructor indeed! So I thought I’d do without!

protocol/proto/decide/v1alpha1/decide_pb.d.ts
wooorm-arcjet: This is generated, so we cannot change it here. And, as the field uses a `_` there, I’m not sure what’s best 🤔

arcjet/test/filter.test.ts
wooorm-arcjet: Yeah...

My thinking:

a) I want tests on data that is close to reality
b) I want tests on the API that people use in reality

But the APIs that people use do not have hooks to inject some data (and that’s not necessarily bad).
That may be possible in `analyze`, `analyze-wasm`, but then it’s not how people use it anymore.
So I ended up with testing here that the fields do not error and end up as `undetermined`.

Some are strings and some are booleans (and perhaps there will be more types in the future), so that would mean looping over different arrays and string building. It’s possible but I ended up as 🤷 and just listed them.
Perhaps in the future there’s a way to inject data, then we can add `match` / `no match`es too?

Do you have a strong preference?

qw-in: No strong preference - makes sense!

protocol/index.ts
qw-in: I guess it'll be added in the `.d.ts` file? I'd prefer explicit here - seeing the bare declaration tripped my up a second. Not a strong preference

wooorm-arcjet: Yes, it’s completely typed/type safe!
So it’s a bit like writing types for a callback function, even though it’s already inferred.

I can see the argument for adding them too though; I do like return types for APIs/packages 🤔 🤷

Issue #4547, opened 2025-07-02
React Router v7 "Framework" mode support


qw-in:
Given the suggested upgrade path from Remix is to React Router v7, it would be helpful for anyone using the `@arcjet/remix` SDK if we had support for `react-router@7`.

I looked into this some for https://github.com/arcjet/example-remix/issues/75#issuecomment-2918328087, and it appears that a framework-specific adapter (`@arcjet/react-router`) would be required (`@arcjet/node` is not sufficient - can't get `ip` and web standards `Request` type).

I _think_ the adapter would be almost identical to our current `@arcjet/remix` but with different peer dependencies. It seems for example the same [`getLoadContext` pattern](https://github.com/arcjet/example-remix/blob/ad8ee93cfaafbceb19c52e81c2b0b0059dfcabcc/server.js#L6) is used in react-router, although it returns a `Map` now for improved typing. I can't find fantastic documentation for this, but here is a [decision log](https://github.com/remix-run/react-router/blob/86b9182dd61b900ecda2aadb74b0dc24d159cc42/decisions/0014-context-middleware.md) and [the definition in the source of `@react-router/express`](https://github.com/remix-run/react-router/blob/86b9182dd61b900ecda2aadb74b0dc24d159cc42/packages/react-router-express/server.ts#L46).

There is also an upcoming (currently unstable) middleware feature that could impact this. The best information I've found on that currently is in [the changelog](https://reactrouter.com/start/changelog#middleware-unstable). Notably it appears to change the way `context` works.

davidmytton:
We'll need to maintain Remix support as well because [Remix v3 is coming](https://remix.run/blog/wake-up-remix).

wooorm-arcjet:
So, does that mean an `arcjet-react-router` next to `arcjet-remix`?

davidmytton:
Yeah I think so.

davidmytton:
Might be interesting to experiment with the new [middleware](https://reactrouter.com/how-to/middleware) and [context API](https://reactrouter.com/how-to/middleware#context-api) to pass Arcjet data down to loaders / actions.

PR #5199, opened 2025-09-16
feat: add React Router integration

wooorm-arcjet:
This adds an integration that wraps Arcjet core
for React Router.
It has tests and docs.
There are also two examples.
One specifically to use the future middleware API.

What remains is that React Router does not provide IP information.
Every of its adapters (with for example cloudflare or amazon or express or node) does IPs differently.
If users deploy outside of known platforms such as amazon, cloudflare, fly, heroku, render, vercel, etc, then they can set `context.ip` or a header (such as `x-client-ip`).

Closes GH-4547.


README.md
```diff
@@ -55,6 +55,10 @@ Arcjet security features for protecting JS apps:
 - [Remix][quick-start-remix]
 - [SvelteKit][quick-start-sveltekit]
 
+<!--
+TODO: list `react-router` when done.
```
qw-in, line 59: Is this for pre-merge?

arcjet-react-router/README.md
```diff
@@ -0,0 +1,278 @@
...
+## Use
+
+In a route such as `routes/home.tsx` do something like:
+
+```tsx
+import arcjet, { fixedWindow, shield } from "@arcjet/react-router";
+
+// …
+
+const aj = arcjet({
```
qw-in, line 77: We typically suggest defining this once in `lib` or similar

arcjet-react-router/README.md
```diff
@@ -0,0 +1,278 @@
...
+-->
+
+## API
+
+This package exports the identifier
+[`createRemoteClient`][api-create-remote-client].
+The default export is [`arcjet`][api-arcjet].
+It also exports all identifiers from `arcjet` core.
+
+This package exports the [TypeScript][] types
```
qw-in, line 125: Missing link I think

arcjet-react-router/index.ts
```diff
@@ -0,0 +1,348 @@
...
+ */
+export function createRemoteClient(
+  options?: RemoteClientOptions | null | undefined,
+): Client {
+  const settings = options ?? {};
+  const baseUrl = settings.baseUrl ?? baseUrlFromEnvironment(process.env);
+
+  return createClient({
+    baseUrl,
+    // @ts-expect-error: TODO: add.
```
qw-in: For this pr?

arcjet-react-router/index.ts
```diff
@@ -0,0 +1,348 @@
...
+      ip = "127.0.0.1";
+    } else {
+      // In production, warn for every request.
+      state.log.warn(
+        "Cannot find client IP address; if this is a development environment, set the `ARCJET_ENV` environment variable to `development`; in production, provide `context.ip` or an `x-client-ip` (or similar) header",
+      );
+    }
+  }
+
+  const url = new URL(details.request.url);
```
qw-in, line 334: Are we sure `.url` always contains a hostname? This could throw

arcjet-react-router/package.json
qw-in, line 1: I'm no expert on `package.json` wizardry but does it not make sense to list `react-router` somewhere?

examples/react-router-middleware/app/routes/home.tsx
```diff
@@ -0,0 +1,81 @@
...
+    </Form>
+  );
+}
+
+// This action happens on POST requests, which have a body, which contains form data.
+// So this is probably where the Sensitive info rule kicks in.
+export async function action(actionArguments: Route.ActionArgs): Promise<{ message: string }> {
+  const decision = actionArguments.context.get(arcjetDecisionContext);
+
+  if (decision && decision.isDenied()) {
```
qw-in:
```suggestion
  if (decision?.isDenied()) {
```

qw-in (commented):
Didn't get all the way through but some initial comments. Will test it out soon!

arcjet-react-router/README.md
wooorm-arcjet: last line! https://github.com/arcjet/arcjet-js/blob/71a74969dfe93e1f90207caa05c7eba617a5bf99/arcjet-react-router/README.md#L278

README.md
wooorm-arcjet: no, my idea was that it can be done after: nice to first publish this before writing docs on the website, after which that can be linked here!

arcjet-react-router/index.ts
wooorm-arcjet: Would have to be generated in the protobuf stuff. Better to generate inside this PR indeed, but could be later

arcjet-react-router/package.json
wooorm-arcjet: the remix adapter doesn’t have a peer on that, things should work without it; but probably good to have one, I added `"react-router": ">=7"`.

examples/react-router-middleware/app/routes/home.tsx
wooorm-arcjet: ah, good idea, done!

arcjet-react-router/README.md
wooorm-arcjet: Hmm, I hope to do without tips on how best to structure apps here.
With these `## Use` examples in readmes I try to be as short as possible, that a user can get something running that works after installing, before reading on to more info (example: https://github.com/arcjet/arcjet-js/tree/main/arcjet-remix#use).

There is this tip: https://redirect.github.com/arcjet/arcjet-js/blob/682dc47350d45f2ae12ba08c0379a0b93ed14f44/arcjet-react-router/index.ts#L172-L176.

Though I do also hope that folks, when they construct `aj` with the same arguments several times, DRY up their code themselves already?

Maybe this is more for some overarching docs page on the website or main `arcjet-js` readme?

arcjet-react-router/index.ts
wooorm-arcjet: I don’t think this particular case throws because `request` is a Fetch request, on which, the `url` field is a serialized URL, right? https://developer.mozilla.org/en-US/docs/Web/API/Request/url. I don’t imagine browsers constructing an unparseable URL?
The remix adapter doesn’t wrap this either!

wooorm-arcjet:
Thanks for the review Quinn!

arcjet-react-router/README.md
qw-in: Ah, shorthand

arcjet-react-router/index.ts
wooorm-arcjet: done on this branch now btw!

qw-in:
> Thanks for the review Quinn!

No worries, still want to get to this but my week is a little hectic. If you need a review before then maybe worth asking around 😅

arcjet-rei:
Code changes look good; I'll come back later and pull down the branch and run the examples to make sure that all looks good.

wooorm-arcjet:
Sweeet!

wooorm-arcjet:
Thanks for the review Rei!

PR #5202, opened 2025-09-16
chore(ci): Remove dependabot groups on examples and apply cooldown to direct deps

blaine-arcjet:
While we transition to Renovate, this limits the updates Dependabot provides and restricts them to 7 day cooldown period.


wooorm-arcjet:
@blaine-arcjet there was some info in this file on which examples needed locked dependencies. It may still be good to have that info somewhere

Issue #5209, opened 2025-09-18
Detect credit card numbers with dashes


wooorm-arcjet:
We do not currently detect credit cards with dashes in them. We do for phone numbers, so we should implement this for credit card checks as well.

PR #5210, opened 2025-09-18
feat(redact-wasm): ignore dashes in card numbers

wooorm-arcjet:
This generates new WebAssembly into `@arcjet/redact-wasm` which supports this feature.
This also adds tests for card numbers to make it visible what is and is supported in `@arcjet/redact-wasm`

Closes GH-5209.


blaine-arcjet (changes requested):
The detection code is also used in the `sensitiveInfo` rule so we should also include updates to analyze-wasm in this PR.

PR #5211, opened 2025-09-19
build: enable `ignore-scripts`

wooorm-arcjet:
Related reading:

* <https://socket.dev/blog/pnpm-10-0-0-blocks-lifecycle-scripts-by-default>
* <https://pnpm.io/supply-chain-security>

Adding `ignore-scripts` turns of *all* pre/post (lifecycle) scripts. In dependencies that are installed and also in the packages developed here. Hence this now documents to run `npm run build` after `npm ci`.

Personally, I have added `ignore-scripts` to all my packages several years ago, and always use `npm test` to make sure *everything* is perfect: properly built, linted, and tested.


davidmytton:
Makes sense to me - I don't see why we'd need any scripts to run, we're not using complex deps.

examples/express-launchdarkly/README.md
```diff
@@ -0,0 +1,108 @@
+<a href="https://arcjet.com" target="_arcjet-home">
```
qw-in: I think this might need a rebase

PR #5212, opened 2025-09-19
feat(filters): add limit to 10 expressions, 1024 bytes

wooorm-arcjet:
Closes arcjet/arcjet#5820.


qw-in (approved):
Code looks good to me!

Should we consider matching [cloud flares limit of 4096?](https://developers.cloudflare.com/ruleset-engine/rules-language/expressions/#maximum-rule-expression-length)

wooorm-arcjet:
Better to start small, and expand if needed! I could see it happening, though, I dunno, `1024`, seems pretty big to me! 😅

davidmytton:
Yeh I think start smaller and see if we get asked to increase it. Minor difference in performance, but every little bit counts.

qw-in:
Sounds good! Have you measured say 4x1024 vs 1x4096? Mostly out of curiosity

PR #5265, opened 2025-10-03
rename exampple dirs to match arcjet package names

luigimorel:
As mentioned in #72, I have renamed the example directories to match the arcjet package names. 

@blaine-arcjet , should I go ahead and consolidate packages into a single dir? For example express-* can all be moved into express/ and then constituent directories, e.g bots, validate-email. Refer to screenshot attached. 
image


luigimorel:
I would also want to be advised. Is it a requirement to have javascript files emitted? For example in `examples/express/bots`, can I not emit them? 
See tsconfig 

```ts 
{
  "compilerOptions": {
    "checkJs": true,
    "lib": ["dom.iterable", "dom", "esnext"],
    "module": "node16",
    "strict": true,
    "noEmit": true //does not emit js files
  },
  "include": ["**/*.js"]
}
```

luigimorel:
final examples directory 
image

luigimorel:
@wooorm-arcjet  these files seem to be too many for copilot to review. wanna do it with me over a call?

wooorm-arcjet:
Hey Luigi! Thanks for working on things!

Your work is rather different than what we had in mind. Which makes sense: there isn’t that much info in the issues.
We were also not completely done with discussing them. So what’s in the issues and in your PR, isn’t exactly what we’ll end up with.

Main points of feedback: a) this is too big, b) the goal was to *merge*/*remake* examples so that there is say one for express which shows all the features together, instead of making folders for express with subfolders per feature.

We do appreciate open source contributions, but it would be nice to first have you sort of explain what you plan on doing in a few bullet points on issues, after which we can clarify stuff. Then we can prevent unneeded work from happening!

This also reminds me, we should have better contributing docs, there isn’t really anything for external contributors yet. I opened an issue here: https://github.com/arcjet/.github/issues/1!

luigimorel:
Understood. 

Here's what I'm thinking. I feel like you still need what I have so far done. 

I'll explain. 

We have an `examples`  folder. There's a technology called node (it's a `dir`). `node` has it's `basic` example and also a framework example, e.g `hono`. 
To the `basic` example, you can add all the features that Arcjet has to offer. Same for the `hono` example as well. 

Then do the same for all other supported technologies.  

I agree, the PR is big. My apologies. You can just close it. Or, I can just continue and finish what ever you wanted to accomplish.

blaine-arcjet:
@luigimorel I'm going to close this out based on feedback in @wooorm-arcjet's comment.

PR #5269, opened 2025-10-06
fix: support ESM types in dependencies

wooorm-arcjet:
While this repo uses ESM everywhere, this was not really enabled in TypeScript, as that used `moduleResolution: "node"` (an alias for `node10`), which only really supports `require` and importing files.

See: <https://www.typescriptlang.org/tsconfig/#moduleResolution>.

This surfaces for example in Astro using export maps to point people to their Zod in GH-5266. That probably surfaces because of withastro/astro@ad26fa22.

As the word is `node`, we could use `nodenext` (alias `node16`), but as that would require that `module: "esnext"` be changed to something Node-specific too, using the modern `moduleResolution: "bundler"` is likely a better idea. That value does more than is needed for export maps, but it at least does not force more Node-specific configuration.

The next problem is that `@rollup/plugin-typescript` has bugs. which are not very visible because they use a stale-bot to close issues. See rollup/plugins#1583.
For that reason, these `moduleResolution` and `module` fields are explicit in `rollup-config/index.js` too.

The `bundler` module resolution algorithm slightly changes the handling of the internal types, which is why `analyze-wasm` and `redact-wasm` now reference the `wasm.d.ts` files, which importantly ***does not*** leak into the generated `.d.ts` files.

Related-to: GH-5266.

### DIFF
--- a/analyze-wasm/index.ts
+++ b/analyze-wasm/index.ts
@@ -1,3 +1,7 @@
+// Support `?js` and such:
+//
+/// <reference types="./wasm.js" />
+
 import { instantiate } from "./wasm/arcjet_analyze_js_req.component.js";
 import type { ImportObject } from "./wasm/arcjet_analyze_js_req.component.js";
 
--- a/redact-wasm/index.ts
+++ b/redact-wasm/index.ts
@@ -1,3 +1,7 @@
+// Support `?js` and such:
+//
+/// <reference types="./wasm.js" />
+
 import { instantiate } from "./wasm/arcjet_analyze_bindings_redact.component.js";
 import type { ImportObject } from "./wasm/arcjet_analyze_bindings_redact.component.js";
 import type { ArcjetRedactCustomRedact } from "./wasm/interfaces/arcjet-redact-custom-redact.js";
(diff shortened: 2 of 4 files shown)


analyze-wasm/index.ts
```diff
@@ -1,3 +1,7 @@
+// Support `?js` and such:
+//
+/// <reference types="./wasm.js" />
```
blaine-arcjet, line 3: Not blocking because this seems to work, but why did these lose the automatic `wasm.d.ts`?

wooorm-arcjet:
> Not blocking because this seems to work, but why did these lose the automatic wasm.d.ts?

I think the change is Rollup/`bundler`-specific: I believe I did not see it on regular TypeScript/`nodenext`.
The type error surfaced in `analyze`/`redact`, not `analyze-wasm`/`redact-wasm`.
My guess is that with `bundler` (through Rollup), the `.ts` file of dependencies are read?
In which case `wasm?js` fails?

I ran into other errors with Rollup, stuff not visible with regular TypeScript (in IDE). It was rather complex to find something that worked with minimal changes.

This state of 2 different type checks with different workings is perhaps not great in the future too. Given also Node being able to read `.ts` files through type stripping, there are some alternative paths forward, such as that `tsc` can be used for type checking source code and `test/` and `build/` stuff. There is also: we could use `.js` with JSDoc (generating `.d.ts`), drop the `.ts` files, that’s what I do, less build step helps making sure things such as paths work!

PR #5270, opened 2025-10-06
fix(fastify): better `protocol` detection, minimal request interface

wooorm-arcjet:
Updating `fastify` keeps failing the example as that is pinned to a different one, which are not exactly the same.

Other integrations define their own request as a minimum viable interface. This PR adds that too for Fastify.

While doing that, I did notice that `protocol` was also on there, and the current detection was a bit weird, so this now uses `request.protocol` if available (which it always should be according to the types).

Related-to: GH-5237.

### DIFF
--- a/arcjet-fastify/index.ts
+++ b/arcjet-fastify/index.ts
@@ -167,7 +167,10 @@ export interface ArcjetFastifyRequest {
   /**
    * Underlying connection of the incoming request.
    */
-  socket: { encrypted?: boolean | undefined; remoteAddress?: string | undefined };
+  socket: {
+    encrypted?: boolean | undefined;
+    remoteAddress?: string | undefined;
+  };
 
   /**
    * URL of the incoming request.
--- a/arcjet-fastify/index.ts
+++ b/arcjet-fastify/index.ts
@@ -24,7 +24,6 @@ import type {
   ExtraProps,
 } from "arcjet";
 import arcjetCore from "arcjet";
-import type { FastifyRequest } from "fastify";
 
 // TODO(@wooorm-arcjet): using `export all` will leak things in the public API,
 // resulting in unneeded breaking changes,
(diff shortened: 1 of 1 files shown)


arcjet-fastify/index.ts
```diff
@@ -338,8 +379,12 @@ function toArcjetRequest<Properties extends PlainObject>(
...
   const host = headers.get("host") ?? "";
   let path = "";
-  // Note: there may be a better way to detect `https`, no clue.
-  let protocol = request.server.initialConfig.https ? "https:" : "http:";
+  let protocol =
+    request.protocol === "https" || request.protocol === "http"
+      ? request.protocol + ":"
+      : request.server.initialConfig?.https
+        ? "https:"
+        : "http:";
```
qw-in, line 390: This looks like a logic change which I have mixed feeling about landing in a types change. They both seem like good changes tho!

blaine-arcjet: I think I'd prefer to call this `fix(fastify)` and include these together, since this is a small PR and we're making fixes _related_ to the minimum request interface.

arcjet-fastify/index.ts
```diff
@@ -24,7 +24,6 @@ import type {
   ExtraProps,
 } from "arcjet";
 import arcjetCore from "arcjet";
-import type { FastifyRequest } from "fastify";
```
blaine-arcjet, line 27: Does removing this allow us to drop fastify from devDeps? Probably not because it'll still be a peerDep but I wanted to check.

blaine-arcjet (approved):
Besides addressing @qw-in's concern, this looks good to me.

arcjet-fastify/index.ts
wooorm-arcjet: Swapped this PR title/body around to focus on the fix!

wooorm-arcjet:
Continuing the sub-conversation from Blaine at https://github.com/arcjet/arcjet-js/pull/5270#discussion_r2408003516 (while resolving it to land this):

Yes, we could remove the peer (direct?) dep on the framework, as it is no longer mentioned in published `.js` and `.d.ts` files.
This is also the case for some other integrations, as `@arcjet/react-router`.
I think currently the “host framework” is mostly now listed in `peerDependencies` of integrations?
We could have a policy/ADR on this, or it could remain informal.

blaine-arcjet:
> I think currently the “host framework” is mostly now listed in peerDependencies of integrations?
> We could have a policy/ADR on this, or it could remain informal.

I think I only added the peerDep when we had imports (even if it was just a type import). Originally, I didn't even rely on the devDep matching the peerDep but that failed in some cases so I started adding peerDep + matching devDep. I'm fine if we want to signal supported versions with peerDeps but I would prefer less peerDeps when possible because they cause both user and maintainer frustrations.

wooorm-arcjet:
I’m :+1: to either strategy

Issue #5223, opened 2025-09-25
Simplify and streamline examples


wooorm-arcjet:
Some dependencies such as `tailwind`, `radix-ui`, `lucide-react` make it harder to maintain but don’t add much (similar arguments as #5220).

Instead, I think it would be good to streamline the examples so that they look like each other (with some minimal regular CSS) and show similar features.
So that humans can compare exactly what’s different for Arcjet between integrations.
This relates a lot to arcjet/examples#5 and also to the comments on the parent issue GH-72.

davidmytton:
I think most of those come in from a default framework install e.g. Next.js installs various things and includes graphics as well. Agreed on slimming them down to avoid dependencies that aren't relevant to our SDK codebase.

morekid:
I think it's best to properly brainstorm examples in general and mock a design, layout and feature showcase approach that is consistent. See https://www.notion.so/arcjet/2025-08-28-Design-agenda-25c9ddf35eaf8024b8fadf1b26243c97?source=copy_link#25c9ddf35eaf8057aae0c6682f201acf.

We can then think about how to make the implementation more lightweight and maintainable once we have properly decided how these should look. In scope for this also a consideration to make the DS public so that we can use that.

PR #5273, opened 2025-10-07
deps(examples): regenerate package locks

wooorm-arcjet:
This regenerates package locks for bun and node examples.
This pulls in patch releases that match our semver ranges: it does not update semver ranges in `package.json`s.

There was one case where `@arcjet/next` was pulled from remote instead of locally. That is now changed.

For some reason, sometimes, package locks keep having old stuff in there.
I searched for `arcjet/[^"]+": "(?!1\.0\.0-beta\.13|workspace:|file:|\.\.\/)` to find our old stuff and nuked respective `node_modules/`/`package-lock.json`s.


blaine-arcjet:
> There was one case where `@arcjet/next` was pulled from remote instead of locally.

This was by design; however, I cannot say it was a good design. Perhaps we should write down somewhere that all examples in this repository use file paths for arcjet/arcjet-js dependencies to validate changes to main, while arcjet/examples uses published dependencies?

examples/nextjs-bot-categories/package-lock.json
```diff
@@ -1612,39 +1649,295 @@
...
-      },
-      "funding": {
-        "url": "https://opencollective.com/eslint"
-      }
-    },
-    "node_modules/acorn": {
-      "version": "8.14.1",
-      "resolved": "https://registry.npmjs.org/acorn/-/acorn-8.14.1.tgz",
-      "integrity": "sha512-OvQ/2pUDKmgfCg++xsTX1wGxfTaszcHVcTctW4UJB4hibJx2HXxxO5UmVgyjMa+ZDsiaf5wWLXYpRWMmBI0QHg==",
+    "node_modules/@unrs/resolver-binding-android-arm-eabi": {
```
blaine-arcjet: How did `@unrs/*` get included? I've been trying to drop deps that pull that in.

wooorm-arcjet: unrs-resolver -> eslint-import-resolver-typescript -> eslint-config-next -> nextjs-bot-categories

wooorm-arcjet:
Yes, lockfiles are regenerated, because npm does not maintain its package-locks correctly. Though these updates are also what a user sees when running `npm install`.
Which ones do you not want to update?
Sure, I welcome you updating too!

examples/nextjs-react-hook-form/package-lock.json
blaine-arcjet, line 1: Document decision to use exclusively arcjet-js filesystem paths for examples in arcjet-js repository.

examples/nextjs-bot-categories/package-lock.json
blaine-arcjet: This did not need to be updated so I've removed the unnecessary lockfile updates. These deps will eventually get removed as part of #5223

blaine-arcjet:
> Yes, lockfiles are regenerated, because npm does not maintain its package-locks correctly.

Since it is just a json file, you can chop out weird things and run `npm i --package-lock-only` to regenerate just what you want included.

> Though these updates are also what a user sees when running npm install.

We never recommend package-less `npm install`. All our examples say to use `npm ci` which uses the lockfiles provided and our tooling ensures the lockfiles are known-good versions. We do not indiscriminately update an entire lockfile.

> Which ones do you not want to update?  Sure, I welcome you updating too!

I've removed the regenerated file changes and only updated the arcjet-js package definitions. It is worth noting that this is not necessary when using filesystem package paths, so we probably don't need this type of PR in the future. Previously the arcjet-js packages would be updated as part of Dependabot updates and perhaps Renovate will do the same when implemented.

examples/nextjs-react-hook-form/package-lock.json
blaine-arcjet: #5290

PR #5304, opened 2025-10-15
build: remove unused env variable

wooorm-arcjet:
I see no hits for `ARCJET_RUNTIME` on GH in the `arcjet` org.


blaine-arcjet (approved):
Good catch. This was removed long ago.

PR #5305, opened 2025-10-15
fix(nuxt): stash body for `h3` functions

### LINKED ISSUE #5351: Nuxt Adapter: Protect consumes body
As discovered at https://github.com/arcjet/examples/pull/53#discussion_r2430460783,

> If you read the request body before calling #protect, #protect can't see the request body.

> If you try to read the request body after calling #protect, the readFormData call stalls.

wooorm-arcjet:
Closes #5351

Related-to: h3js/h3#616.
Related-to: arcjet/examples#53.

### DIFF
--- a/arcjet-nuxt/internal.ts
+++ b/arcjet-nuxt/internal.ts
@@ -348,6 +348,9 @@ function createGetBody(state: State, event: ArcjetH3Event) {
           expectedLength,
           limit: 1048576,
         });
+        // Set it to integrate with most H3 utilities.
+        // See: <https://github.com/h3js/h3/pull/616>.
+        event.node.req.body = body;
         return body;
       }
 


arcjet-nuxt/internal.ts
```diff
@@ -348,6 +348,9 @@ function createGetBody(state: State, event: ArcjetH3Event) {
           expectedLength,
           limit: 1048576,
         });
+        // Set it to integrate with most H3 utilities.
+        // See: <https://github.com/h3js/h3/pull/616>.
+        event.node.req.body = body;
```
blaine-arcjet, line 353: is this stream drained?

blaine-arcjet: The PR linked is confusing, perhaps we want to permalink this code: https://github.com/oscartbeaumont/h3/blob/845eecaaaeffdb415321f76744e31d55074f37b8/src/utils/body.ts#L46 which seems to indicate that a non-stream body will be passed to `Buffer.from()` and then wrapped in another promise for the framework to use.

An additional question is whether something can read the body before Arcjet and add attached the buffered body at this same property. If so, we should look for it first before trying to read again.

wooorm-arcjet: Basically, yeah; no longer readable, is the error.

wooorm-arcjet: This code above *is* taking that `body`, if it is there, set by some tool. But not all the complex other fields.
We *could* use `readBody` from `h3` instead of `@arcjet/body`, if we want to depend on that, to improve compatibility with those tools, if that dependency is worth it. Or we could similarly support all those fields.

However, in the related example using the deprecated `readFormData`, that does not set it. That would be an upstream much like the above PR but the inverse. I tried it for a little bit but couldn’t quickly get it working 🤔

blaine-arcjet:
I'm going to convert this to draft because I think we need to consider all the scenarios (such as someone reading before vs after, etc). We'll want to add something to the example to show this working since adapters don't have tests.

wooorm-arcjet:
Closing as the issue is closed as not planned: https://github.com/arcjet/arcjet-js/issues/5351#issuecomment-3784344415

PR #5309, opened 2025-10-17
test(redact-wasm): add some end to end tests

### LINKED ISSUE #5352: End-to-end tests for redact-wasm
We should have some JS tests that test the behavior of executing the Wasm logic

wooorm-arcjet:
As there are now some card number tests,
this adds the others.

Note in particular that paren groups for phone numbers don’t add up.

Closes #5352


wooorm-arcjet:
Thanks for the review!

PR #5362, opened 2025-11-03
fix!: handle phone numbers with paren groups while restricting groups

wooorm-arcjet:
* fix phone numbers w/ paren groups (`(020) 334 4522` and `+1 (555) 555-5555` are now detected)
* change to slightly restrict number groups, previously `1 iPhone2` was seen as one group, these are now separate tokens
* change filter parse errors (now prefixed with `warn:`)
* remove a duplicate test

Closes GH-5363.


arcjet/test/filter.test.ts
```diff
@@ -177,7 +177,7 @@ test("filter: `protect`", async function (t) {
     assert(result.reason instanceof ArcjetErrorReason);
     assert.equal(
       result.reason.message,
-      'Filter parsing error (1:1):\nhttp.blob ~ "Chrome"\n^^^^^^^^^ unknown identifier\n',
+      'warn: Filter parsing error (1:1):\nhttp.blob ~ "Chrome"\n^^^^^^^^^ unknown identifier\n',
```
wooorm-arcjet: these are the prefixes introduced to lower severity

redact-wasm/test/index.test.ts
```diff
@@ -268,16 +268,8 @@ test("card number", async function (t) {
...
-      {
-        end: 19,
-        identifiedType: { tag: "credit-card-number" },
-        original: "4242-4242 4242-4242",
-        redacted: "<Redacted credit card number #0>",
-        start: 0,
-      },
-    ]);
+  await t.test("should not work w/ dashes and spaces", async function () {
+    assert.deepEqual(wasm.redact("4242-4242 4242-4242", emptyOptions), []);
```
wooorm-arcjet, line 272: this is due to slightly stricter merging of number groups: it’s no longer one single token

redact-wasm/test/index.test.ts
```diff
@@ -578,19 +568,6 @@ test("phone number", async function (t) {
...
-    async function () {
-      assert.deepEqual(
-        wasm.redact("1.2.3.4", {
-          ...emptyOptions,
-          entities: [{ tag: "phone-number" }],
-        }),
-        [],
-      );
-    },
-  );
```
wooorm-arcjet, line 593: exact same test is above

arcjet/test/filter.test.ts
blaine-arcjet: We can/should filter these in the JS bindings. I can put together a draft and regenerate into this PR.

redact-wasm/test/index.test.ts
blaine-arcjet: I think this is a breaking change then

wooorm-arcjet: I considered it not intentional that `xxx1 2yyy` was seen as one token in arcjet/arcjet#6089. I left the non-first tokens as-is but made the 1st token stricter. If you want to make the 1st token looser again, how loose? Allowing `-`? Other values?

blaine-arcjet: I didn't say we needed to change it, just that it was a breaking change. I updated the title of the PR.

Issue #5382, opened 2025-11-06
`@arcjet/ip`: problem when proxy string looks like a CIDR


wooorm-arcjet:
`@arcjet/ip` accepts `proxies` as an array of `string` and `Cidr`.
It does not attempt to parse `string`s as CIDRs but treats them as static IPs, which it compares to actual IPs, which will never match.
An example of this mistake is at https://docs.arcjet.com/ip#proxy-filtering.
That CIDR range would never match, it has to be passed through `parseProxy` first.

One solution is to call `parseProxy` *inside* `findIp`. That may be slow as it would be called a lot.
Another solution is to throw an error if a proxy contains `/`

davidmytton:
Can you update the docs to correct the example please?

blaine-arcjet:
`findIP` is called within Arcjet handlers and we try to avoid throwing inside them. I think we should try adding the parsing logic because the Arcjet SDKs pre-parse the proxy, so they'd skip the parsing inside this function.

wooorm-arcjet:
As integrations already pass things through `parseProxy`, the branch would not be hit

PR #5385, opened 2025-11-06
fix(ip): throw on proxy CIDR range in string form

wooorm-arcjet:
It is also possible to throw this error in `isTrustedProxy`. The downside of that is that the check then occurs for every potential IP, and that the error would not happen if no potential IP is found. This verifies the input options once.

An alternative approach is to call `parseProxy` instead of throwing. The downside of that is that it would be faster if users do that initially and once.

Closes GH-5382.


blaine-arcjet:
Converting to draft because I don't think we want to throw here (see discussion issue).

Issue #5380, opened 2025-11-06
SDKs: support `ip` field


wooorm-arcjet:
Users have all kinds of setups. They may notice that Arcjet does not find the correct IP address, log headers, and know which value they trust and want to use.
A good solution would be to raise issues, configure `platform` and `proxies` correctly, but that is involved.
Instead, an escape hatch can be allowed and documented.

GH-4808 is like that, but with a header, and importantly only for development.
It is a good idea to support a header in development: that allows users to test different things without redeploying, and there’s no danger of malicious users.
I think it’s good to use a *very different* method in production, as in this case, there are malicious users, which can send headers.

As `ip` is already documented on the website, while not working in every SDK, I think that’s the best way to go. It doesn’t always have to be the same field everywhere, but it would help to streamline things. Like GH-5019.

blaine-arcjet:
Adapters work on the specific "request" for a framework or runtime. This is not available on many platforms and that is why it is not applied

PR #5386, opened 2025-11-06
feat: support `ip` field in all integrations

wooorm-arcjet:
Some already support `request.ip`.
This brings that same field to all integrations. So that it works the exact same way for everyone.
Alternatively, it is possible to do different things per integration.

Closes GH-5380.


blaine-arcjet:
I've marked the issue as not planned with the reason so closing this.

Issue #5390, opened 2025-11-07
`@arcjet/env`: add support for `firebase` platform


wooorm-arcjet:
From https://github.com/arcjet/arcjet-js/issues/5383#issuecomment-3498892720, it is possible to infer `firebase` based on `FIREBASE_CONFIG`.

PR #5393, opened 2025-11-07
feat(ip): add `firebase` platform

wooorm-arcjet:
CI fails now and should work after GH-5392.

Closes GH-5390.


blaine-arcjet:
Merged into https://github.com/arcjet/arcjet-js/pull/5392

PR #5394, opened 2025-11-07
feat(ip): support cidr strings in proxies

wooorm-arcjet:
Closes GH-5382.

### DIFF
--- a/ip/index.ts
+++ b/ip/index.ts
@@ -858,7 +858,11 @@ export function findIp(
   if (Array.isArray(rawProxies)) {
     for (const cidrOrIp of rawProxies) {
       if (typeof cidrOrIp === "string") {
-        proxies.push(isCidr(cidrOrIp) ? parseCidr(cidrOrIp) : cidrOrIp);
+        proxies.push(parseProxy(cidrOrIp));
+      }
+
+      if (isIpv4Cidr(cidrOrIp) || isIpv6Cidr(cidrOrIp)) {
+        proxies.push(cidrOrIp);
       }
     }
   }
--- a/ip/test/ip.test.ts
+++ b/ip/test/ip.test.ts
@@ -279,6 +279,21 @@ test("`findIp`", async (t) => {
     );
   });
 
+  await t.test("should filter an invalid value in a proxy", function () {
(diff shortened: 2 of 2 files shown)


davidmytton:
Nice. No need to fix the docs example with this 😎 We can still use your IP concepts doc though to explain how the SDK detects the IPs since this is becoming more complex.

ip/index.ts
```diff
@@ -852,7 +852,19 @@ export function findIp(
...
   options?: Options | null | undefined,
 ): string {
-  const { platform, proxies } = options || {};
+  const { platform, proxies: rawProxies } = options || {};
+  const proxies: Array<Cidr | string> = [];
+
+  if (rawProxies) {
+    for (const cidrOrIp of rawProxies) {
+      proxies.push(
+        typeof cidrOrIp === "string" && isCidr(cidrOrIp)
```
blaine-arcjet: Should we move `typeof cidrOrIp === "string"` into the `isCidr` helper function?

blaine-arcjet: Actually, I just realized that these come from the user-facing API so JS users could provide a non-string. We should check strings before pushing onto the proxies list at all.

blaine-arcjet: We also need to check that `rawProxies` is an array (which may change depending on the resolution of our other open issues).

ip/index.ts
```diff
@@ -852,7 +852,21 @@ export function findIp(
...
   options?: Options | null | undefined,
 ): string {
-  const { platform, proxies } = options || {};
+  const { platform, proxies: rawProxies } = options || {};
+  const proxies: Array<Cidr | string> = [];
+
+  if (Array.isArray(rawProxies)) {
+    for (const cidrOrIp of rawProxies) {
+      if (typeof cidrOrIp === "string") {
+        proxies.push(parseProxy(cidrOrIp));
```
blaine-arcjet, line 861: There's a small chance to throw in the CIDR parsing if the user specifies the CIDR incorrectly, but I think that is a far enough edge case that we don't need to catch & filter

blaine-arcjet (approved):

Issue #5395, opened 2025-11-07
SDKs: support `trustedHeader` field


wooorm-arcjet:
ENG-46

blaine-arcjet:
Adding a configuration option such as `trustedIpHeader` has too many edge cases and gotchas that users might stumble into, such as a spoofed header on a public (yet hard to guess) endpoint. I'm going to close this because I believe that #4265 solves the problem in a more robust way that reduces the chance to read from a spoofed header.

PR #5396, opened 2025-11-07
feat: support `trustedIpHeader` field

wooorm-arcjet:
Closes GH-5395.

### DIFF
--- a/arcjet-astro/index.ts
+++ b/arcjet-astro/index.ts
@@ -17,7 +17,7 @@ const resolvedVirtualClientId = "\0ARCJET_VIRTUAL_CLIENT";
 
 const validateMode = z.enum(["LIVE", "DRY_RUN"]);
 const validateProxies = z.array(z.string());
-const validateTrustedHeader = z.string().optional();
+const validateTrustedIpHeader = z.string().optional();
 const validateCharacteristics = z.array(z.string());
 const validateClientOptions = z
   .object({
@@ -221,7 +221,7 @@ type ArcjetIntegrationOptions<Characteristics extends readonly string[]> = {
    * It can contain IPv4 or IPv6 addresses.
    * Proxies are filtered out.
    */
-  trustedHeader?: string | null | undefined;
+  trustedIpHeader?: string | null | undefined;
 };
 
 function validateAndSerialize<
@@ -649,7 +649,7 @@ export function createRemoteClient(options?: RemoteClientOptions | undefined) {
 export default function arcjet<Characteristics extends readonly string[]>(
   options: ArcjetIntegrationOptions<Characteristics> = { rules: [] },
 ): AstroIntegration {
-  const { rules, characteristics, client, proxies, trustedHeader } = options;
+  const { rules, characteristics, client, proxies, trustedIpHeader } = options;
   const arcjetImports = new Set();
   const arcjetRules: string[] = [];
   for (const rule of rules) {
@@ -666,8 +666,8 @@ export default function arcjet<Characteristics extends readonly string[]>(
     ? `proxies: ${validateAndSerialize(validateProxies, proxies)},`
     : "";
 
-  const trustedHeaderInjection = trustedHeader
-    ? `trustedHeader: ${validateAndSerialize(validateTrustedHeader, trustedHeader)},`
+  const trustedIpHeaderInjection = trustedIpHeader
+    ? `trustedIpHeader: ${validateAndSerialize(validateTrustedIpHeader, trustedIpHeader)},`
     : "";
 
   const clientInjection = client
@@ -756,7 +756,7 @@ export default function arcjet<Characteristics extends readonly string[]>(
                           ],
                           ${characteristicsInjection}
                           ${proxiesInjection}
-                          ${trustedHeaderInjection}
+                          ${trustedIpHeaderInjection}
                           ${clientInjection}
                         })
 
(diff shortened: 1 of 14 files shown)


arcjet-astro/index.ts
```diff
@@ -208,6 +209,20 @@ type ArcjetIntegrationOptions<Characteristics extends readonly string[]> = {
    * (optional, example: `["100.100.100.100", "100.100.100.0/24"]`).
    */
   proxies?: string[];
+  /**
+   * Name of (lowercase) HTTP request header that you trust (such as
```
davidmytton: Can we normalize it so it doesn't matter whether they enter it in lower case or not? I imagine some people just copying it in from the headers e.g. `X-Real-Ip` and then things not working.

arcjet-astro/index.ts
```diff
@@ -208,6 +209,20 @@ type ArcjetIntegrationOptions<Characteristics extends readonly string[]> = {
...
+   * This value is *preferred* over IP addresses provided by the
+   * framework and IP addresses found in other headers based on the platform,
+   * in both development and production.
+   *
+   * It can point to a regular IP (such as `x-client-ip`) or a list of IPs
+   * (such as `x-forwarded-for`).
+   * It can contain IPv4 or IPv6 addresses.
+   * Proxies are filtered out.
+   */
+  trustedHeader?: string | null | undefined;
```
davidmytton: What do you think about naming this `trustedIpHeader` instead? I can't see where we might want to trust another header in future, but this would make it more descriptive.

arcjet-astro/index.ts
wooorm-arcjet: done!

arcjet-astro/index.ts
wooorm-arcjet: :+1: done!

blaine-arcjet (changes requested):
If I am reading this PR correctly, this option doesn't correlate an IP or CIDR range with a header, so users can configure their SDK with an active attack vector. Here's the example: if they are unaware that using a Cloudflare proxy in front of a Vercel application leaves the public Vercel URL accessible, someone can submit requests via the public Vercel URL and spoof the custom header.

The only time we can trust a "trusted header" is if the connecting IP matches a set of known IPs that are allowed to add the header.

davidmytton:
Where possible we're detecting the platform e.g. the new Firebase support in https://github.com/arcjet/arcjet-js/issues/5390, but there are cases where it's too difficult to do it without making assumptions about the setup. This provides a config option so users who know what they're doing can override our settings. It's secure by default, and users can override if they need to.

blaine-arcjet:
> This provides a config option so users who know what they're doing can override our settings. It's secure by default, and users can override if they need to.

But user's don't know what they are doing with this option, as specified in my example. Instead, this needs to map between trusted IPs and the header that will be set. For the cloudflare example, that would be something like:

```js
{
  "173.245.48.0/20": "cf-connecting-ip",
  "103.21.244.0/22": "cf-connecting-ip",
  "103.22.200.0/22": "cf-connecting-ip",
  "103.31.4.0/22": "cf-connecting-ip",
  "141.101.64.0/18": "cf-connecting-ip",
  "108.162.192.0/18": "cf-connecting-ip",
  "190.93.240.0/20": "cf-connecting-ip",
  "188.114.96.0/20": "cf-connecting-ip",
  "197.234.240.0/22": "cf-connecting-ip",
  "198.41.128.0/17": "cf-connecting-ip",
  "162.158.0.0/15": "cf-connecting-ip",
  "104.16.0.0/13": "cf-connecting-ip",
  "104.24.0.0/14": "cf-connecting-ip",
  "172.64.0.0/13": "cf-connecting-ip",
  "131.0.72.0/22": "cf-connecting-ip",
}
```

If they are managing their own, the server will likely have a specific IP, etc.

davidmytton:
That doesn't solve the problem, it just moves the config to a different field. If they could specify all the IPs then they'd do that in the existing proxies config. As [we have discussed](https://linear.app/arcjet/issue/ENG-46/create-an-api-to-map-proxy-address-to-a-different-header), where the list of IPs isn't available or is very long e.g. Firebase/GCP, configuring them all is too difficult. 

We can document the risks and leave it to the user to decide what they want to do when there is an unusual setup.

blaine-arcjet:
> If they could specify all the IPs then they'd do that in the existing proxies config.

The existing proxy config does not work with platform detection because platform detection is *designed* to pull the exact connecting IP. If the platform detection couldn't give us the exact connecting IP, there would be no point in having it.

We should **not** add configuration options that explicitly allow IP spoofing attacks. Instead, the trusted proxies need to map a trusted header. That is the only way to avoid allowing a spoofed header between 2 public services.

blaine-arcjet:
> where the list of IPs isn't available or is very long e.g. Firebase/GCP

Firebase is solved by platform detection, GCP gives you CIDR ranges in the UI if I remember the report from the past.

blaine-arcjet:
Drafting while we look at #5398

wooorm-arcjet:
This also solves for example:

- when users deploy to platforms that are not detected from `env` such as Cloudflare
  — an alternative is to make `platform` configurable
- when users deploy to platforms other than Firebase, Fly, Render, or Vercel, in which case Arcjet will look at 11 headers, ±10 of which can be set by malicious users
  — an alternative is to make platform configurable and to support these headers/platforms more narrowly
- when users deploy to new/other platforms
  — alternative same as above/below
- when users put other services or custom services in front of where they deploy
  — an alternative is a `Service` design which allows arbitrary headers, though, these tools should probably be well-behaving and set `x-forwarded-for`/`forwarded` correctly, for which `proxies` as strings is enough

blaine-arcjet:
Closing this because #5395 is not planned due to the edge cases and gotchas that users may introduce into their application with the option.

Issue #4265, opened 2025-05-30
Consider an API to map proxy addres to a different header


blaine-arcjet:
When an application is fronted by a proxy like Cloudflare, the connecting socket address will be in Cloudflare's CIDR ranges. We support filtering these via the `proxies` configuration on our SDK, but then we don't have the connecting client's IP. Cloudflare will attach the client IP as `cf-connecting-ip`, so we could map back to that; however, we can't do this for all users because we don't know their specific proxy setup.

Perhaps we could expand the proxy configuration to allow re-mapping of headers, such as:

```
arcjet({
  // ... config
  proxies: {
    "cf-connecting-ip": [
      // Cloudflare CIDR ranges
    ]
  }
})
```

qw-in:
I'll port my internal comment over and modify it a bit:

```
arcjet({
  // ... config
  proxies: cloudflareProxy(flyio())
})
```

Could also make it generic:

```
arcjet({
  proxies: headerFromCIDR({
    header: "cf-connecting-ip",
    cidr: "Cloudflare CIDR ranges",
  })
})
```

blaine-arcjet:
If we add `platform` as a configurable property on the SDK, we could provide something like:

```ts
function cloudflareProxy() {
  return {
    platform: "cloudflare",
    proxies: [
        // All their CIDR ranges but those could fall out of date
    ]
  }
}
```
And then it would be used like:

```ts
arcjet({
  // ... config
  ...cloudflareProxy()
})
```

One thing to highlight is that we don't really want to try to maintain the list of Cloudflare CIDR ranges. They can change over time (and even the way you fetch them could change and break old SDKs).

wooorm-arcjet:
@qw-in do you remember the actual problem that user ran into?
What were they using?
Was `platform` detected correctly from `env`?
Why was the `proxies` option not enough for them?

qw-in:
> do you remember the actual problem that user ran into?

I've mostly seen this when someone has the cloudflare proxy in front of their server and Arcjet can't get the ip of a request

> What were they using?

I assume you are asking about hosting platform or framework and it varies. The issue is the proxy though

> Was platform detected correctly from env?

Probably? But it would not be detecting say "cloudflare"

> Why was the proxies option not enough for them?

Cloudflare doesn't include the ip address in `x-forwarded-for` (which is what the proxies config reads). Instead it is in `cf-connecting-ip`

wooorm-arcjet:
[...]

Issue #5381, opened 2025-11-06
Create `@arcjet/proxies` (or so)


wooorm-arcjet:
Several services that people (can) put in front of Arcjet provide lists of CIDR ranges that they use.
These lists don’t seem to change much. Cloudflare changes about once per year.
To prevent each user from crawling that data, this data can be provided in a package.
In a package, the data can also be molded into a format specifically made for Arcjet, depending on GH-5378 and GH-5379.

IPs from Google: <https://www.gstatic.com/ipranges/goog.json>
IPs from CF: <https://www.cloudflare.com/ips-v4/>, <https://www.cloudflare.com/ips-v6/>.

Alternative names are `@arcjet/ip-data`, `@arcjet/services`.
Alternatively can also live in `@arcjet/ip`.

PR #5398, opened 2025-11-07
draft: add services

wooorm-arcjet:
Added this to `arcjet-react-router` because that has tests.

Closes GH-4265.


arcjet-react-router/test/index.test.ts
```diff
@@ -395,6 +395,94 @@ test("`default`", async function (t) {
...
+              "2405:8100::/32",
+              "2405:b500::/32",
+              "2606:4700::/32",
+              "2803:f800::/32",
+              "2a06:98c0::/29",
+              "2c0f:f248::/32",
+            ],
+            platform: "cloudflare",
+          },
+        ],
```
blaine-arcjet: I like this *a lot*! I wonder if we can change the API to something like [my example](https://github.com/arcjet/arcjet-js/pull/5396#issuecomment-3503499875) to make this easy to configure different services that all point to 1 application:

```js
services: {
    "103.21.244.0/22": "cloudflare",
    "103.22.200.0/22": "cloudflare",
    "103.31.4.0/22": "cloudflare",
    "104.16.0.0/13": "cloudflare",
    "104.24.0.0/14": "cloudflare",
    "108.162.192.0/18": "cloudflare",
    "131.0.72.0/22": "cloudflare",
    "141.101.64.0/18": "cloudflare",
    "162.158.0.0/15": "cloudflare",
    "172.64.0.0/13": "cloudflare",
    "173.245.48.0/20": "cloudflare",
    "188.114.96.0/20": "cloudflare",
    "190.93.240.0/20": "cloudflare",
    "197.234.240.0/22": "cloudflare",
    "198.41.128.0/17": "cloudflare",
    "2400:cb00::/32": "cloudflare",
    "2405:8100::/32": "cloudflare",
    "2405:b500::/32": "cloudflare",
    "2606:4700::/32": "cloudflare",
    "2803:f800::/32": "cloudflare",
    "2a06:98c0::/29": "cloudflare",
    "2c0f:f248::/32": "cloudflare",
    // some google range
    "192.104.160.0/23": "gcp"
    // some range I manage
    "192.168.0.1": "my-own-platform"
}
```

The problem with the above that it relies exclusively on supported platforms and doesn't allow me to configure my custom proxy. If these platforms pointed to headers, we'd be able to support custom configurations:

```js
services: {
    "103.21.244.0/22": "cf-connecting-ip",
    "103.22.200.0/22": "cf-connecting-ip",
    "103.31.4.0/22": "cf-connecting-ip",
    "104.16.0.0/13": "cf-connecting-ip",
    "104.24.0.0/14": "cf-connecting-ip",
    "108.162.192.0/18": "cf-connecting-ip",
    "131.0.72.0/22": "cf-connecting-ip",
    "141.101.64.0/18": "cf-connecting-ip",
    "162.158.0.0/15": "cf-connecting-ip",
    "172.64.0.0/13": "cf-connecting-ip",
    "173.245.48.0/20": "cf-connecting-ip",
    "188.114.96.0/20": "cf-connecting-ip",
    "190.93.240.0/20": "cf-connecting-ip",
    "197.234.240.0/22": "cf-connecting-ip",
    "198.41.128.0/17": "cf-connecting-ip",
    "2400:cb00::/32": "cf-connecting-ip",
    "2405:8100::/32": "cf-connecting-ip",
    "2405:b500::/32": "cf-connecting-ip",
    "2606:4700::/32": "cf-connecting-ip",
    "2803:f800::/32": "cf-connecting-ip",
    "2a06:98c0::/29": "cf-connecting-ip",
    "2c0f:f248::/32": "cf-connecting-ip",
    // some google range
    "192.104.160.0/23": "whatever-gcp-uses",
    // some range I manage
    "192.168.0.1": "my-custom-header"
}
```

This, in combination with #5381 would allow users to do:

```js
services: {
    ...cloudflare(),
    // some google range
    ...gcp(),
    // some range I manage
    "192.168.0.1": "my-custom-header"
}
```

But that begs the question of why we name this `services` instead of just expanding the type definition of `proxies` to allow this:

```js
proxies: {
    ...cloudflare(),
    // some google range
    ...gcp(),
    // some range I manage
    "192.168.0.1": "my-custom-header"
}
```

blaine-arcjet: Even barring the improvements from #5381 (if we wanted to delay it), we can document that our users can use this function to map all Cloudflare IPs to the header:

```js
async function cloudflare() {
    const ips = [
        ...await fetch("https://www.cloudflare.com/ips-v4/").then(resp => resp.text()).then((ips) => ips.trim().split("\n")),
        ...await fetch("https://www.cloudflare.com/ips-v6/").then(resp => resp.text()).then((ips) => ips.trim().split("\n"))
    ].map((ip) => [ip, "cf-connecting-ip"])

    return Object.fromEntries(ips)
}
```

davidmytton: My preference is to use the existing `proxies` property for this because it provides a single place for similar config:

- Listing the individual proxy IPs or CIDRs
- Listing known services with IP allowlists (individual or CIDR) and the associated header we should look at

Where possible we'll auto-detect the platform and get the right header, but this will allow users to specify the headers to check where requests are coming from specific IPs/CIDRs.

The problem with loading IPs at runtime like https://github.com/arcjet/arcjet-js/pull/5398/files#r2504848944 in the performance overhead of the fetch. This is probably acceptable on a server startup, but not in serverless environments. It's better to bundle the list of IPs in at build-time. We could provide them in a package, but we'd need to keep that package up to date in an automated fashion.

wooorm-arcjet: > I wonder if we can change the API to something like my example to make this easy to configure different services that all point to 1 application:

I don’t know what is easier about that example. With services is an array, your example using 3 services, would look like:

```js
services: [
    {
        ips: [
            "103.21.244.0/22",
            // …
            "2c0f:f248::/32",
        ],
        platform: "cloudflare",
    },
    // some google range
    {ips: ["192.104.160.0/23"], platform: "gcp"},
    // some range I manage
    {ips: ["192.168.0.1"], platform: "my-own-platform"}
]
```

> The problem with the above that it relies exclusively on supported platforms and doesn't allow me to configure my custom proxy. If these platforms pointed to headers, we'd be able to support custom configurations:

Using objects for different services also allows for different properties:

```js
    // …
    // some google range
    {
        headers: ["whatever-gcp-uses"],
        ips: ["192.104.160.0/23"]
    },
    // some range I manage
    {
        headers: ["my-custom-header"],
        ips: ["192.168.0.1"]
    }
```

Alternatively, a `platform` is a name for one or more headers, and that mapping can also be made explicit and configurable:

```
{platform: "MyCustomPlatform", ips: ["192.168.0.1"]} -> {headers: ["my-custom-header"], platform: "MyCustomPlatform"}
```

> This, in combination with GH-5381 would allow users to do:
>
> ```js
> services: {
>     ...cloudflare(),
>     // some google range
>     ...gcp(),
>     // some range I manage
>     "192.168.0.1": "my-custom-header"
> }
> ```

With services as an array, the example looks like:

```js
services: [
    cloudflare,
    gcp,
    {headers: ["my-custom-header"], ips: ["192.168.0.1"]}
]
```

However, I want to stress that this case is not tested: Is this on Vercel? What order are these tools in? Or is this hosted on 3 domains, each with a different service, if so why? Are there infinite loops? Should GCP and the custom proxy not be configured to pass the correct IP to Cloudflare? Is this really something to support?

> But that begs the question of why we name this services instead of just expanding the type definition of proxies to allow this:

I wonder about ordering of things and infinite loops.
Though this worry also exists with `services: Array<Service>`.

***

To take a step back, I believe the root problem is that Vercel is not a good web citizen and strips `x-forwarded-for`, whereas other tools are forwarding who connects to them correctly. For these other tools, `proxies` (and adding support for platforms, such as for firebase) are enough.

I think the practical case of this feature is only one entry: “yes, I am on Vercel, and even though Vercel hides it, I use cloudflare, and so `cloudflare` can be trusted (in certain cases)”
If users also put other (custom) custom proxies in front of that (first: why?), they should pass correct IPs to Cloudflare.
If users host their app 3 times, each with different proxies, mostly also: why?

Your point about needing to verify whether Cloudflare can be trusted, can also be met by users doing:

```js
import {matches} from "@arcjet/ip"
import {cloudflare} from "@arcjet/proxies"

if (matches(request.ip, cloudflare)) {
    request.ip = request.headers.get("cf-connecting-ip")
}
```

As for users not verifing and omitting that if-condition, humans can also do `"0.0.0.0/0": "my-custom-header"` (a CIDR matching any IPv4) or `request.headers.set("x-real-ip", request.headers.get("cf-connecting-ip"))`.

wooorm-arcjet: A problem with letting users pass header names, is that there are different header values, and a proper solution should likely model that:

- some specify IPv4, others IPv6, (probably) others both
- IPs sometimes contain ports (w/ `:`)
- Sometimes keywords such as `unknown` or `hidden` appears
- some are a single IP, others are comma-separated lists. But there are also more complex values (such as `forwarded`) which have key/value pairs, and quoted values, in which case it becomes more precarious to simply split on commas.

Edit: GH-5404 and GH-5409 help with this

ip/index.ts
```diff
@@ -812,7 +907,7 @@ export interface Options {
   /**
    * Trusted proxies.
    */
-  proxies?: ReadonlyArray<string | Cidr> | null | undefined;
+  proxies?: ReadonlyArray<Cidr | Service | string> | null | undefined;
```
blaine-arcjet: This API is still too complex. Nothing about the platform should leak into this API because a proxy is not the platform the application is *running* on.

In addition to `ReadonlyArray<Cidr | string>`,  `null`, and `undefined`, we should allow `Record<string, string>` that is a mapping of IPs or CIDRs to the header. To improve performance of the SDKs, we could also accept a pre-parsed Map as `Map<Cidr | string, string>`
```suggestion
  proxies?: ReadonlyArray<Cidr | string> | Record<string, string> | Map<Cidr | string, string> | null | undefined;
```

All other concepts for services can be pushed into other building blocks, such as my previous suggestion to map all Cloudflare ranges to the `cf-connecting-ip` header.

wooorm-arcjet:
ok I don’t see this landing any time soon, closing

Issue #5152, opened 2025-09-10
Create an adapter for Netlify  edge functions


qw-in:
Based on the discussion around https://github.com/arcjet/arcjet-js/issues/5135 & a meeting between @blaine-arcjet and I it is becoming clear that if we want to support [Netlify's edge functions](https://docs.netlify.com/build/edge-functions/overview/) they'll need their own adapter.

Some of the known challenges:
- Edge functions run on a Deno 1.X runtime
- Edge functions have quite strict compute limits (they timeout calling `.protect` today even on the personal plan)
- Deno 1.X doesn't support `npm:@arcjet/*` currently (have to use a cdn like `https://esm.sh/@arcjet/deno@1.0.0-beta.11`)
- We'd need to pull the `ip` off of the [Netlify global](https://docs.netlify.com/build/edge-functions/api/#netlify-global-object)
- Netlify has some [framework-specific integrations of edge functions](https://docs.netlify.com/build/edge-functions/overview/#use-cases) we'd want to keep in mind

Something to evaluate before proceeding too far is Netlify's view on the future of their edge functions & Deno. The sense I get is that they may be evaluating other options rather than updating to Deno 2 but that's just my gut feeling.

PR #5404, opened 2025-11-10
refactor(ip): simplify which headers are used when

wooorm-arcjet:
Given that more platforms are added, more headers are added, and potentially users can configure platforms/headers too, and potentially different formats are added, this moves 200 lines of if statements into configuration, to improve and prepare for these changes.

### DIFF
--- a/ip/index.ts
+++ b/ip/index.ts
@@ -955,7 +955,7 @@ const candidatesByPlatform: Record<Platform, ReadonlyArray<Candidate>> = {
 /**
  * Generic candidates.
  */
-const genericCanidates: ReadonlyArray<Candidate> = [
+const genericCandidates: ReadonlyArray<Candidate> = [
   // Standard headers used by Amazon EC2, Heroku, and others.
   { format: "ip", header: "x-client-ip" },
   // Load-balancers (AWS ELB) or proxies.
@@ -1000,7 +1000,7 @@ function findIpInHeaders(
   const candidates =
     platform && platform in candidatesByPlatform
       ? candidatesByPlatform[platform]
-      : genericCanidates;
+      : genericCandidates;
 
   if (headers && typeof headers === "object") {
     for (const { format, header } of candidates) {


ip/index.ts
```diff
@@ -896,213 +877,158 @@ export function findIp(
...
+      const value = getHeader(headers, header);
+
+      if (value) {
+        if (format === "ips") {
+          // As per MDN X-Forwarded-For Headers documentation at
+          // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For
+          // The `x-forwarded-for` header may return one or more IP addresses as
+          // "client IP, proxy 1 IP, proxy 2 IP", so we want to split by the comma and
+          // trim each item.
+          for (const item of value.split(",").reverse()) {
```
blaine-arcjet, line 1016: We must not have enough tests because you changed this behavior and no tests changed.

wooorm-arcjet: what changed?

blaine-arcjet: I thought the `reverse()` changed but that was handled in outside of the parse function.

wooorm-arcjet: indeed! 😅

For `@arcjet/ip`, the test coverage is very good!

blaine-arcjet:
Thinking out loud: What if instead of a `format`, we provided a function to execute? The important ones would be parsing `x-forwarded-for` and `forwarded` (after #5409) but this would also allow flexibility on platforms that put the IP in an environment variable or something.

wooorm-arcjet:
Yes, that is possible. I see it as a next step on top.
The reason is that `ips` / `rfc7239` are more tough to get right for users. But those are the most likely candidates of this.
Example case: a user is on a vercel-like platform, which strips `forwarded`, and they use some proxy. The user does control their proxy, and adds a rule `req.headers.my-x-forwarded-for = req.header.x-forwarded-for; req.headers.my-forwarded = req.header.forwarded`.
One alternative to that is to expose these parsers as functions too, but the downside is that it becomes part of the API surface. With keywords, there is more control.

Instead of turning `format` into a function, it is also possible to allow `options.proxies` (or `options.proxies[number]`) as function. This is what Fastify/Express allow too. It is a bit different though, so may not be enough: those functions would get found IPs, and return whether they are accepted or not.

It depends on how you see your case “this would also allow flexibility on platforms that put the IP in an environment variable or something.”
I imagine such platforms to put their dedicated IP in there? In which case `proxies: [process.env['dedicated-ip']]` is static, or `proxies(ip) { return ip !== process.env['dedicated-ip'] }` is dynamic

wooorm-arcjet:
> There's probably another design that can retain flexibility while not using if-statements but is it really going to be better than just writing the if-statement and custom code inside it?

A reason for using an array of objects is to support the `Service` API correctly, as it maps to further headers, those headers may contain lists of candidate IPs, which may be further ignored or map to yet further headers, and so on and on. So, these conditions and handling are generated dynamically.

wooorm-arcjet:
Here’s some pseudocode to show more complex needs of `Service`s:

```js
      findIp(
        {
          headers: {
            "x-forwarded-for": "6.6.6.6, 2.2.2.2, 1.1.1.1",
            "x-my-forwarded-for": "6.6.6.6, 4.4.4.4, 3.3.3.3",
          },
        },
        {
          platform: "vercel",
          proxies: ["1.1.1.1", { format: "ips", header: "x-my-forwarded-for", ips: ["2.2.2.2"] }, "3.3.3.3"],
        },
      )
```

1. `6.6.6.6` is evil, a malicious user set that
2. `platform: vercel` means the rightmost `x-forwarded-for` is trusted
3. that IP is the first of `proxies`, which means it is a trusted appender, so `N - 1` can be used
4. that IP `2.2.2.2` is in the service, so it is trusted to set header M correctly
5. `format: ips` dictates that the header is another right-to-left list
6. that IP `3.3.3.3` is also in `proxies`, a trusted appender, so `N - 1` can be used
7. `4.4.4.4` is the client IP; and it is not listed as a trusted proxy

blaine-arcjet:
> I imagine such platforms to put their dedicated IP in there? 

Unfortunately no. Netlify actually provides the conneting IP on the `Netlify` global at `Netlify.context.ip` and we'll want to support Netlify through [an adapter](https://github.com/arcjet/arcjet-js/issues/5152) in the future. And while the Netlify IP could be plucked and passed in by the adapter, that would make the result of `findIP()` different between direct usage and the adapter (which may or may not be a big deal).

> A reason for using an array of objects is to support the Service API correctly

You are talking about the `Service` API as though it is already completed, but I think we want to simplify it down to just a mapping.

> Here’s some pseudocode to show more complex needs of Services:

I understand your desire to map an "ip list" header to a trusted proxy, but I think that is making the API too complex. I think it is fine to restrict the mapped header to be exactly *one* IP. If a service allows remapping of headers, they are likely going to support extracting one IP from a list too.

blaine-arcjet:
I'm going to draft this PR because of my concerns above. If this is informed by how the proxy mapping API is designed, then we need to work backwards from the simplest form of that.

PR #5410, opened 2025-11-11
feat(proxies): add `cloudflare` and `google` IP ranges

wooorm-arcjet:
Closes GH-5381.

### DIFF
--- a/proxies/eslint.config.js
+++ b/proxies/eslint.config.js
@@ -9,4 +9,4 @@ import base from "@arcjet/eslint-config";
  *
  * @type {Array<Linter.Config>}
  */
-export default [...base, { ignores: ["test/*.js", "index.js"] }];
+export default [...base, { ignores: ["test/*.js", "build.js", "index.js"] }];


davidmytton:
How do we keep this up to date and how do we know when the IPs change?

wooorm-arcjet:
> How do we keep this up to date

Similar to generated wasm: run scripts once in a while.
For the wasm, that’s from the other repo; for this, `node build.ts` / `npm run generate`.

> how do we know when the IPs change?

The generated source code is checked into Git, so changes are visible.

***

Cloudflare is infrequent; for others (such as `goog.json` which is added here now, but also future candidate sources) it’s good to wait before landing to verify that the list is indeed somewhat static.

If useful lists are rather dynamic, it makes more sense to expose functions instead of lists. Users can then do `const ips = await google()` or so.

If servers are somewhat long-lived (several days) and candidate lists are very dynamic (several hours), the `proxies` option could be made dynamic (accept functions).

That being said, 10 steps back, the `proxies` option only works if there are *several* potential client IPs, and those are only available on headers. As headers are sent to the server, it may be possible to stop trying to find the “true” client IP locally in SDKs/integrations, and instead do that on the server. On the server, everything around such lists can be controlled, and IP lookups are possible.

wooorm-arcjet:
> a) knowing when the IPs change;

The `last-modified` header on CF point to today, so that means it cannot be trusted.
For Google, there is `createDate` and `syncToken`, but those also point to today, so that seems useless too.
So, to answer this, a hash of the input/output could be used. Or a `git` command to check if the working tree is dirty.

> b) re-generating and releasing the packages.

this is more involved and can go in various directions

> We could do both - provide a static list, which is the default recommended way to avoid loading every time. But also expose a function to allow them to pull in the list themselves.

An option is to put these functions copy/pasteable on the website instead (this is like [blaines comment here](https://github.com/arcjet/arcjet-js/pull/5398#discussion_r2504848944)).
Is that an acceptable alternative solution to this issue/PR to you?
Or should this issue/PR still happen, just with a) and b)?

davidmytton:
We can do b) manually, but we need to know when there's a change to the lists so we can then do a release shortly afterwards. 

> An option is to put these functions copy/pasteable on the website instead (this is like https://github.com/arcjet/arcjet-js/pull/5398#discussion_r2504848944).
> Is that an acceptable alternative solution to this issue/PR to you?

Yeah we can do that for now. Then if we can do a) then we can release this PR as well.

wooorm-arcjet:
They’re in the docs PR now.

blaine-arcjet:
Drafting until we have time to implement the automatic updates.

PR #5450, opened 2025-11-24
fix!: Require quotes for `'script'` in `require-trusted-types-for` directive

hckhanh:
This pull request makes a minor adjustment to the `CspDirectives` interface in `nosecone/index.ts`, specifically updating the type for the `requireTrustedTypesFor` property to ensure it matches the expected string literal format. 

* Type definition update:
  * Changed the type of `requireTrustedTypesFor` from `ReadonlyArray<"script">` to `ReadonlyArray<"'script'">` in the `CspDirectives` interface to correctly reflect the required string literal format.

### DIFF
--- a/nosecone/index.ts
+++ b/nosecone/index.ts
@@ -92,7 +92,7 @@ export interface CspDirectives {
   navigateTo?: StaticOrDynamic<Source | ActionSource> | undefined;
   reportUri?: string[] | undefined;
   reportTo?: string[] | undefined;
-  requireTrustedTypesFor?: ReadonlyArray<"script"> | undefined;
+  requireTrustedTypesFor?: ReadonlyArray<"'script'"> | undefined;
   trustedTypes?:
     | ReadonlyArray<"none" | "allow-duplicates" | "*" | string>
     | undefined;


nosecone/index.ts
```diff
@@ -92,7 +92,7 @@ export interface CspDirectives {
   navigateTo?: StaticOrDynamic<Source | ActionSource> | undefined;
   reportUri?: string[] | undefined;
   reportTo?: string[] | undefined;
-  requireTrustedTypesFor?: ReadonlyArray<"script"> | undefined;
+  requireTrustedTypesFor?: ReadonlyArray<"'script'"> | undefined;
```
blaine-arcjet, line 95: According to https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/require-trusted-types-for, the `'script'` value is indeed supposed to be quoted. However, I believe this also needs to be added to the `QUOTED` constants used for validation and a test needs to be added.

hckhanh: Hi @blaine-arcjet 
I tried the non-quote value and it threw me an error and said that it is not correct
So I started to think that this is a type miss match

Can you tell me any other places that need to update?

davidmytton: You'd need to add this into https://github.com/arcjet/arcjet-js/blob/db7d2a6f776f57b0904f61c3d431116f94473227/nosecone/index.ts#L515 and write a test to cover this case.

davidmytton:
Closing as this has become stale, but feel free to reopen with the requested changes.

Issue #5453, opened 2025-11-25
env: remove unused `apiKey` utility function


wooorm-arcjet:
This function is unused.
All SDKs require users to retrieve API keys themselves and pass them explicitly.
The only exception is the Arcjet Astro SDK, which does not use this function but retrieves the key the regular Astro way.
As this function is unused and undocumented (on the site, in examples), and probably also not recommended for use, I’d recommend removing it.

PR #5454, opened 2025-11-25
chore(env)!: remove unused `apiKey` utility function

wooorm-arcjet:
This function is unused.
All SDKs require users to retrieve API keys themselves and pass them explicitly.
The only exception is the Arcjet Astro SDK,
which does not use this function but retrieves the key the regular Astro way.
As this function is unused and undocumented (on the site, in examples),
and probably also not recommended for use,
I’d recommend removing it.

Closes GH-5453.

### DIFF
--- a/env/README.md
+++ b/env/README.md
@@ -58,13 +58,7 @@ npm install @arcjet/env
 
 ```ts
 import process from "node:process";
-import {
-  apiKey,
-  baseUrl,
-  isDevelopment,
-  logLevel,
-  platform,
-} from "@arcjet/env";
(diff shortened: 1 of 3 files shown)


qw-in (approved):
Awesome! The less we can touch globals the better IMO - especially with all the different runtimes we're trying to support.

I did a sanity check and didn't find this being written to anywhere. Probably worth a release note just in case when this gets published tho

davidmytton (approved):

wooorm-arcjet:
yay!

Issue #5457, opened 2025-11-25
Fix reading from `process.env` in Bun, Deno


wooorm-arcjet:
The `@arcjet/bun` and  `@arcjet/deno` packages almost read from `{ env } from "bun"` / `Deno.env` correctly, except for each a sole lonely `if (isDevelopment(process.env)) {` check.

PR #5458, opened 2025-11-25
fix(bun, deno): do not read from `process.env`

wooorm-arcjet:
The `@arcjet/bun` and  `@arcjet/deno` packages almost read from `{ env } from "bun"` / `Deno.env` correctly,
except for each a sole lonely `if (isDevelopment(process.env)) {` check.

Closes GH-5457.


davidmytton:
Merging manually due to Trunk stalling.

PR #5467, opened 2025-11-27
ci: fix trunk hanging

wooorm-arcjet:
Related-to: arcjet/arcjet#6259.


davidmytton:
Forcing the merge because Trunk is broken.

PR #5484, opened 2025-12-02
deps: update `astro`

wooorm-arcjet:


examples/astro-integration/package-lock.json
blaine-arcjet, line 1: This looks like a full package-lock rebuild, as things like esbuild increased. We should scope the updates to just the upgraded packages.

package-lock.json
```diff
@@ -96,14 +96,379 @@
...
-        "astro": "5.16.0",
+        "astro": "5.16.3",
         "eslint": "9.39.1",
         "typescript": "5.9.3"
       },
       "peerDependencies": {
         "astro": "^5.9.3"
       }
     },
+    "arcjet-astro/node_modules/astro": {
```
blaine-arcjet: It is weird that this got hoisted. I need to dig in deeper to understand why it would be.

examples/astro-integration/package-lock.json
wooorm-arcjet: `npm update astro` also updates the things inside it, I think dependabot does the same

package-lock.json
wooorm-arcjet: It moves the packages `cookie@1.1.1`, `estree-walker@3.0.3`, `fdir@6.5.0`, `magicast@0.5.1`, `p-limit@6.2.0`, `picomatch@4.0.3`, `unstorage@1.17.3`, `vite@6.4.1`, `yocto-queue@1.2.2`, `zod-to-json-schema@3.25` inside it.
Of those, `cookie`, `estree-walker`, `magicast`, `p-limit`, `picomatch`, `yocto-queue` are *older* in the root, newer in the astro integration.
The reverse is true for `vite`, newer in root, older in astro.

examples/astro-integration/package-lock.json
blaine-arcjet: I don't trust the `npm update` command. I always change the version and use `npm install --package-lock-only` to apply the minimum set of changes to the lockfile. In the case of this large astro increase, there are still a lot of changes.

examples/astro-integration/package.json
```diff
@@ -12,7 +12,7 @@
   "dependencies": {
     "@arcjet/astro": "file:../../arcjet-astro",
     "@astrojs/node": "^9.4.3",
```
blaine-arcjet: This needed to be updated in tandem to avoid duplicated dependencies in the lockfile.

package-lock.json
blaine-arcjet: I do not believe this characterization is correct. I deleted these lines and ran `npm i --package-lock-only` and just the versions that needed upgrades were updated. However, I'm not sure what puts npm in the inconsistent state.

blaine-arcjet (approved):

PR #5488, opened 2025-12-02
deps(examples): update `node-forge`

wooorm-arcjet:


blaine-arcjet:
> not possible in newman as it locks things and hasn’t updated.

Has the API changed? If not, you can specify an override and update the lockfile.

arcjet-rei:
Yes, you want to use [`"overrides"`](https://docs.npmjs.com/cli/v8/configuring-npm/package-json#overrides) in the example's `package.json` for this:

```json
  "overrides": {
    "newman": {
      "node-forge": "1.3.2"
    }
  }
```

PR #5456, opened 2025-11-25
feat: sync lists of env variables

wooorm-arcjet:
* add `FIREBASE_CONFIG`, `RENDER` to `@arcjet/astro`
* add `FIREBASE_CONFIG` to `@arcjet/env`, `@arcjet/node`
* add `FIREBASE_CONFIG` to `turbo.json`
* remove duplicate code for `VERCEL_GIT_COMMIT_SHA` in `@arcjet/next`
* update env readme

It looks like `NODE_ENV` is also intentionally not supported on Astro. That may be intentional but is an exception, as all other SDKs do read from it, so it could be added too.

Closes GH-5455.

PR #5490, opened 2025-12-02
feat: send extra `vercel-git-commit-sha` field to cloud api

wooorm-arcjet:
In `@arcjet/next`, some headers and an env variable were sent in `extra` on Vercel. As `arcjet` does not have access to the environment, such behavior must be in integrations. This PR changes each integration to send such information on Vercel.

As headers are already sent in headers, I removed that, but that can be added.

As an alternative to this PR, this behavior could also be removed from `@arcjet/next`.

Related-to: GH-5456.

### DIFF
--- a/arcjet-astro/internal.ts
+++ b/arcjet-astro/internal.ts
@@ -292,8 +292,19 @@ export function createArcjetClient<
       }
     }
 
+    const extra: Record<PropertyKey, string> = {};
+
+    // Add extra info from `env` on Vercel.
+    if (platform(env) === "vercel") {
+      // Vercel git commit SHA.
+      // <https://vercel.com/docs/environment-variables/system-environment-variables#VERCEL_GIT_COMMIT_SHA>
+      // @ts-expect-error: types in GH-5456.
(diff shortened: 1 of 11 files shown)


blaine-arcjet:
Closing this, as the conversation in Slack ended with:

> we could include it, but we don't use these values for anything right now so I wouldn't add more until we have a use for them

We'll remove these and open an issue to add them if we need them later.

wooorm-arcjet:
I think @davidmytton’s last comment was about adding more fields, specifically `VERCEL_GIT_COMMIT_REF`

blaine-arcjet:
It was also stated that:

> we don't use it right now, but the idea was to be able to provide more details in future

If we don't use it, we should not keep it. We can open an issue to gather feedback if any user needs it.

davidmytton:
To clarify what I meant:

1. We collect these, but don't use them.
2. We only collect them with the Next.js SDK adapter. To be consistent we could collect them from every SDK adapter, but it's not a high priority because we don't use them.
3. We shouldn't collect any new fields until we decide we need them.

PR #5504, opened 2025-12-04
deps(examples): remove old `eslint`, `eslint-config-next`

wooorm-arcjet:
This example intentionally uses older packages.
This bumps past vulnerabilities.


examples/nextjs-14-nextauth-4/package.json
```diff
@@ -23,7 +23,7 @@
     "@types/react-dom": "^18",
     "autoprefixer": "^10",
     "eslint": "^8",
-    "eslint-config-next": "^14",
```
blaine-arcjet, line 26: Let's just drop this dependency since they aren't in sync now.

Issue #5518, opened 2025-12-10
`@arcjet/node`, others: do not read body if `Content-Length` says it’s too large


wooorm-arcjet:
(others are `@arcjet/nest` and `@arcjet/nuxt`)

These packages are limited to 1mb, and through `@arcjet/body` verify that the actual read body matches `Content-Length` when it is defined, and they throw if that doesn’t match.

That sounds like it’s possible to exit early when `Content-Length` is defined and it is larger than the imposed limit.

Related-to: GH-4562.

Issue #5517, opened 2025-12-10
`@arcjet/bun`, `@arcjet/deno`, others: implement limits when reading bodies


wooorm-arcjet:
Currently, WHATWG `Request` based SDKs call `request.text()`. This can read big files.

Stream-based SDKs (`@arcjet/nest`, `@arcjet/node`, and `@arcjet/nuxt`) read up to 1mb.

It would be nice to have similar behavior across SDKs.

It may be possible to support streams for web requests too, it looks like `request.body` is a `ReadableStream`: https://developer.mozilla.org/en-US/docs/Web/API/Request/body.
Alternatively, it may be possible to check `Content-Length`.

Related-to: GH-4562.

Issue #5516, opened 2025-12-10
`@arcjet/node`: hangs when body is already read


wooorm-arcjet:
When a user reads the body after passing `request` into `@arcjet/node`, it’ll hang.
The `request.readableEnded: boolean` field is turned to `true`.
It several other SDKs, requests are cloned inside the SDK.
~This should probably work for Node too.~ it doesn’t, because node streams are not cloneable like that

wooorm-arcjet:
Closing, this is inherent to streams, which is why it’ll be deprecated in GH-5679.
Users that use tools such as `body-parser` are not affected. Those patch `request.body` and `@arcjet/node` will take that without warning.
Users can read the body and pass it with `sensitiveInfoValue` in GH-5678.

Issue #5515, opened 2025-12-10
`@arcjet/bun`: silently does not read body when already read


wooorm-arcjet:
When a user reads the body before passing `request` into `@arcjet/bun`, the SDK will receive a `""` body back.

In Deno (and probably others), reading a used stream again will throw, which will cause SDKs to log an error and return an error decision.

There is a `request.bodyUsed: boolean` field which could be used to tell users that `@arcjet/bun` cannot do work.

PR #5519, opened 2025-12-10
test: add some tests for reading the request body

wooorm-arcjet:
To check if GH-5505 would work, and figure out what wouldn’t, I added some tests for `@arcjet/bun`, `@arcjet/deno`, `@arcjet/fastify`, and `@arcjet/node`. These are the integrations with simple servers that can easily be tested, for others it’ll likely mean pulling into React and JSX and maybe even playwright. There are various bugs with reading the body, I opened GH-5515, GH-5516, GH-5517, GH-5518. And there are existing bugs, GH-4562, GH-5351, GH-1399.

Note that current tests and CI and container is based on Node.js. So I did not make Bun and Deno run on `npm test`. I guess that also doesn’t really make sense. Their corresponding examples, in `reusable-examples.yml`, do install those runtimes. So a solution would probably be something like that but in `reusable-test.yml`?

Related-to: GH-5505.

### DIFF
--- a/.github/workflows/reusable-test.yml
+++ b/.github/workflows/reusable-test.yml
@@ -11,7 +11,7 @@ env:
 
 jobs:
   bun-test:
-    name: "Run tests: Bun"
+    name: Bun
     permissions:
       contents: read
     runs-on: ubuntu-latest
@@ -33,7 +33,7 @@ jobs:
       - run: bun test
         working-directory: arcjet-bun
   deno-test:
-    name: "Run tests: Deno"
+    name: Deno
     permissions:
       contents: read
     runs-on: ubuntu-latest
--- a/.trunk/trunk.yaml
+++ b/.trunk/trunk.yaml
@@ -57,11 +57,11 @@ actions:
     - trunk-upgrade-available
 merge:
   required_statuses:
+    - Run tests / Bun
+    - Run tests / Deno
     - "Run tests / Run tests (OS: ubuntu-latest, Node: 20)"
     - "Run tests / Run tests (OS: ubuntu-latest, Node: 22)"
     - "Run tests / Run tests (OS: ubuntu-latest, Node: 24)"
-    - "Run tests: Bun"
-    - "Run tests: Deno"
     - Build examples / astro-integration
     - Build examples / bun-hono-rate-limit
     - Build examples / bun-rate-limit


qw-in:
Very cool to see! I'll be able to drop manual clone calls from some of the examples soon!

arcjet-bun/test/index.test.ts
```diff
@@ -0,0 +1,346 @@
+// Note: this monorepo is tested with Node.js, but this package targets Bun.
+// These tests don’t work with Node.js, only with the Bun test runner.
+// Use `bun test` to run them.
```
arcjet-rei: To be able to run this command, `bun` needs to be added as a [feature](https://github.com/michidk/devcontainers-features/tree/main/src/bun) to `.devcontainer/devcontainer.json`.

arcjet-deno/test/index.test.ts
```diff
@@ -0,0 +1,337 @@
+// Note: this monorepo is tested with Node.js, but this package targets Deno.
+// These tests don’t work with Node.js, only with the Deno test runner.
+// Use `deno test test/index.test.ts --allow-env --allow-net --no-check` to run them.
```
arcjet-rei: To do run this command, Deno needs to be added as a [feature](https://containers.dev/features) (search for Deno) to `.devcontainer/devcontainer.json`.

arcjet-rei (changes requested):
Don't forget that, per our discussion a couple months ago, all development needs to be done within the devcontainers. You should also add a README to `arcjet-bun` and `arcjet-deno` with instruction on how to run the tests, because most people aren'g going to be looking for comments in the files.

wooorm-arcjet:
@arcjet-rei how about this?
Most Deno devcontainer features do include more stuff, and then things start breaking in this monorepo that is Node focussed.

qw-in:
@wooorm-arcjet if it's helpful I have a deno devcontainer over in arcjet/examples that I keep entirely separate (in a branch I have a very similar bun version as well)

wooorm-arcjet:
That sounds pretty great?!
Though I do worry that https://github.com/arcjet/examples/blob/c334c0f552b8919200f602ae4d5cdc72fa39897d/.devcontainer/deno-2/devcontainer.json#L7C32-L7C38 and such will interfere 🤔 That’s what I saw with trying some of the ones listed over on MSs website

wooorm-arcjet:
Thanks for the review!

PR #5533, opened 2025-12-12
test: add body tests for Vite-based frameworks

wooorm-arcjet:
This is like GH-5519, but for 4 Vite based frameworks. This essentially checks small apps into Git, serves those apps with Vite, and posts into them.
This uses `vite dev` to run the server.
I investigated building for production,
but that sometimes requires more dependencies and generates many artifacts.

This comes at a cost of introducing more dev-dependencies which is a pity, though those are already in the examples.
In the end I do think it’s worth it though, to have tests for SDKs.


wooorm-arcjet:
wip for now until I get back to solve CI

PR #5534, opened 2025-12-12
fix(astro): swallow astro error on missing ip

wooorm-arcjet:
Astro itself sometimes throws an error when getting the client ip:

https://github.com/withastro/astro/blob/16f3994fdb83d1b3421491c00bfd5ac9f7e37a5c/packages/astro/src/core/middleware/index.ts#L103-L110

This prevents `@arcjet/astro`s own error from being thrown:

https://github.com/arcjet/arcjet-js/blob/ca5a64a30fe4293db9715785aa69899efd7abcc0/arcjet-astro/internal.ts#L267-L270

I extracted this code, which was needed for GH-5533 to work, so that it can go in its own commit.

### DIFF
--- a/arcjet-astro/middleware.ts
+++ b/arcjet-astro/middleware.ts
@@ -6,8 +6,16 @@ import { defineMiddleware } from "astro/middleware";
 const ipSymbol = Symbol.for("arcjet.ip");
 
 export const onRequest = defineMiddleware((ctx, next) => {
+  let ip: string | undefined;
+
+  try {
+    ip = ctx.clientAddress;
+  } catch {
+    // Swallow error.
+  }
+
   if (!ctx.isPrerendered) {
-    Reflect.set(ctx.request, ipSymbol, ctx.clientAddress);
+    Reflect.set(ctx.request, ipSymbol, ip);
   }
 
   return next();


arcjet-astro/middleware.ts
```diff
@@ -6,8 +6,16 @@ import { defineMiddleware } from "astro/middleware";
 const ipSymbol = Symbol.for("arcjet.ip");
 
 export const onRequest = defineMiddleware((ctx, next) => {
+  let ip: string | undefined;
+
+  try {
+    ip = ctx.clientAddress;
+  } catch {
+    // Swallow error.
+  }
```
blaine-arcjet, line 15: If this is throwing again, then they've re-introduced a bug for `clientAddress` not always being available in 
`!isPrerendered` context. @qw-in actually fixed this in the past for the Astro team.

What are the scenarios where this failed? We probably need to report (and possibly fix) upstream

wooorm-arcjet: Failing test case is GH-5533. It may be specific to local Vite, given GH-5535.

Maybe upstream would be better yes.

blaine-arcjet: When you say "Vite-based" in the PR title, does that mean you are compiling to static files? The Arcjet SDK doesn't support static-generated sites since there is nothing to protect.

wooorm: it sounds like that but no, it’s testing the server build (node adapters), using `vite dev`. I’ll do more digging when I am back

blaine-arcjet: Thanks 🙏 I'm going to convert to draft to reduce discussions on PR review.

PR #5535, opened 2025-12-12
fix(sveltekit): swallow sveltekit error on missing ip

wooorm-arcjet:
SvelteKit itself sometimes throws an error when getting the client ip:

https://github.com/sveltejs/kit/blob/2140546af29c1fae747e1c4f54c09ef0fed139ad/packages/kit/src/exports/vite/dev/index.js#L563-L567

This prevents `@arcjet/sveltekit`s own default handling (such as header lookups, `127.0.0.1` in dev, warning in prod) from working:

https://github.com/arcjet/arcjet-js/blob/ca5a64a30fe4293db9715785aa69899efd7abcc0/arcjet-sveltekit/index.ts#L317-L327

I extracted this code, which was needed for GH-5533 to work, so that it can go in its own commit.

### DIFF
--- a/arcjet-sveltekit/index.ts
+++ b/arcjet-sveltekit/index.ts
@@ -305,15 +305,18 @@ export default function arcjet<
     const xArcjetIp = isDevelopment(env)
       ? headers.get("x-arcjet-ip")
       : undefined;
+
+    let maybeIp: string | undefined;
+
+    try {
+      maybeIp = event.getClientAddress();
+    } catch {
+      // Swallow error.
+    }
+
     let ip =
       xArcjetIp ||
-      findIp(
-        {
-          ip: event.getClientAddress(),
-          headers,
-        },
-        { platform: platform(env), proxies },
-      );
+      findIp({ ip: maybeIp, headers }, { platform: platform(env), proxies });
     if (ip === "") {
       // If the `ip` is empty but we're in development mode, we default the IP
       // so the request doesn't fail.


arcjet-sveltekit/index.ts
```diff
@@ -305,15 +305,18 @@ export default function arcjet<
...
       ? headers.get("x-arcjet-ip")
       : undefined;
+
+    let maybeIp: string | undefined;
+
+    try {
+      maybeIp = event.getClientAddress();
+    } catch {
+      // Swallow error.
+    }
```
qw-in, line 315: Is it possible to be a bit more defensive/explicit here?


```suggestion
    try {
      maybeIp = event.getClientAddress();
    } catch (cause) {
    
      if (!isTheErrorWeWantToSwallow(cause)) {
        throw cause;
      }
      // Swallow error because...
    }
```

wooorm-arcjet: the problem is that this error is adapter based, and there are many adapters 🤔

qw-in: > the problem is that this error is adapter based, and there are many adapters

Oh no! Hmm 🤔

So I guess the tradeoff is (read: pick one):

1. We leave this unchanged and occasionally throw when we shouldn't
2. We swallow the error and could potentially hide errors depending on adapter

If we went with 2, is it reasonable to assume astro would raise a similar error/warning anyway elsewhere?

wooorm-arcjet: 1\. was pretty annoying for Vite locally, it just failed. Which is extra annoying because Arcjet has all this code for all these cases.

blaine-arcjet: What triggers this? Is it static generation? If so, that is unsupported by Arcjet and we may want to go with @qw-in's (1) here.

wooorm: The code linked in OP throws the error: https://github.com/sveltejs/kit/blob/2140546af29c1fae747e1c4f54c09ef0fed139ad/packages/kit/src/exports/vite/dev/index.js#L563-L567. I’ll do more digging when I am back, but it appeared to be `vite dev` on a sveltekit server with adapter-node

blaine-arcjet: Interesting. Excited to find out what changed in SvelteKit. I'm going to convert to a draft to reduce discussion in PR review.

Issue #29, opened 2023-12-13
Add release workflows


blaine-arcjet:
Need:
- [x] [Version bumping](https://github.com/arcjet/arcjet-js/pull/74)
- [ ] Publishes
- [ ] Provenance

blaine-arcjet:
We might want to use GitHub Actions attestations and then use the `--provenance-file` flag to the npm command: https://github.com/npm/cli/pull/6490

blaine-arcjet:
Some research links for build provenance:
* https://docs.npmjs.com/cli/v10/commands/npm-publish#provenance
* https://github.com/secure-systems-lab/dsse?tab=readme-ov-file
* https://github.com/npm/cli/tree/latest/workspaces/libnpmpublish
* https://github.com/slsa-framework/slsa-github-generator
* https://github.com/actions/attest-build-provenance?tab=readme-ov-file#outputs
* https://docs.github.com/en/actions/security-for-github-actions/using-artifact-attestations/using-artifact-attestations-to-establish-provenance-for-builds

blaine-arcjet:
We've attempted to sign up for the OIDC/"trusted publishers" beta for npm so we could get closer to securely publishing from CI.

Ref: https://github.com/orgs/community/discussions/161015#discussioncomment-13325080

wooorm-arcjet:
that landed now, right?

blaine-arcjet:
> that landed now, right?

We have access to the trusted publishers preview.

PR #5578, opened 2026-01-05
chore: add publish workflow

wooorm-arcjet:
This adds `publish.yml` workflow based on
<https://docs.npmjs.com/trusted-publishers#github-actions-configuration>. Together with configuration of each package on `npmjs.com`, this sets up both automatic publishing (after a PR such as GH-5461 is merged) and automatic provenance (reference: https://docs.npmjs.com/trusted-publishers#automatic-provenance-generation).

There are a couple things to discuss, such as:

1. which node to publish with: I’d say 24
2. whether this will work: this uses `--dry-run` for now to prevent accidents
3. what trigger to use: this uses tags matching `v[0-9]+.*` now

I believe that the correct corresponding config on `npmjs.com` for all packages is:

* organization: `arcjet`
* repository: `arcjet-js`
* filename: `publish.yml`
* environment:`npm-publish`

Screenshot 2026-01-05 at 11 10 13 AM

Closes GH-29.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,3 +1,12 @@
+name: Publish
+on:
+  push:
+    tags:
+      - v*
+  workflow_dispatch: {}
+permissions:
+  contents: read
+  id-token: write
 jobs:
   publish:
     environment: npm-publish
@@ -17,12 +26,3 @@ jobs:
       # UNKNOWN(2): this should work when everything is set up correctly, but `--dry-run` for now.
       - name: Publish packages
         run: npm publish --dry-run --workspaces
-name: Publish
-on:
-  push:
-    tags:
-      - v*
-  workflow_dispatch: {}
-permissions:
-  contents: read
-  id-token: write


davidmytton:
We've avoided doing automatic publishing (https://github.com/arcjet/arcjet-js/issues/29) so far because of the potential for compromise. Running the process manually at least includes a verification step.

However, I think we can do this with the new GitHub rulesets.

1. Trigger the publish workflow on tag creation e.g. 'v1.0.0' like you have in here.
2. We set up a ruleset to block direct tag creation except by the publishers team.
3. Create the tag in the release please flow then when the PR is merged, the tag is created and the publish workflow is triggered.

blaine-arcjet:
We also want Github "Environments" and to block the deployment on approval of the workflow in that environment using a ruleset.

wooorm-arcjet:
The “Environment name” I remembered is in the screenshot above, so that means that goes onto npm’s website

blaine-arcjet:
Yeah, that'll be the same Environment as protects the workflow itself. 👍

davidmytton:
I've created the environment called "NPM Publish" and a ruleset for the `v*` tag.

**Environment**

image

**Ruleset**

image

Let me know how that works.

wooorm-arcjet:
In my OP and in the code there are 3 questions to discuss. Particularly the last, whether `v*` is too broad, maybe applies to this rule too?

The environment config *looks* okay to me, but this is the first time I see environments 😅

blaine-arcjet:
We don't want to apply this to any tags or branches. We are only using it for https://docs.github.com/en/actions/how-tos/deploy/configure-and-manage-deployments/control-deployments#using-environments

FWIW, this probably should also be lowercase and hyphenated to make it easier to enter into the workflow.

davidmytton:
I've removed the tag ruleset and set up the environment called `npm-publish`.

wooorm-arcjet:
used `npm-publish` as an env here.

@blaine-arcjet are you saying that the `on.push.tags` should be removed here? Or was that about the GH ruleset interface

blaine-arcjet:
> Or was that about the GH ruleset interface

Just the ruleset 👍

.github/workflows/publish.yml
```diff
@@ -1,18 +1,25 @@
+jobs:
+  publish:
+    environment: npm-publish
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
```
blaine-arcjet: Let's name our steps.

.github/workflows/publish.yml
```diff
@@ -1,18 +1,25 @@
...
+      - uses: actions/setup-node@395ad3262231945c25e8478fd5baf05154b1d79f # v6.1.0
+        with:
+          # UNKNOWN(1): Which Node to use? More recent seems better to me?
```
blaine-arcjet: We need a version that contains the npm that supports trusted publishers so we need latest stable node. 👍

.github/workflows/publish.yml
```diff
@@ -1,18 +1,25 @@
...
 name: Publish
-
 on:
-  workflow_dispatch: {}
   push:
     tags:
-      - v*
-
+      # UNKNOWN(3): GH (<https://docs.npmjs.com/trusted-publishers#github-actions-configuration>) uses simply `v*`, that seems too loose? Syntax docs at <https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#filter-pattern-cheat-sheet>.
+      - v[0-9]+.*
```
blaine-arcjet: Why are you saying it is too loose? I think any tag starting with `v` is probably fine. Honestly, it should just be *any* tag 🤷

wooorm-arcjet: ok done

blaine-arcjet (approved):
This LGTM!

wooorm-arcjet:
Configured `arcjet`, `nosecone`, all `@arcjet/*` (except `@arcjet/slack`) packages, and all `@nosecone/*` packages with the info in the OP on `npmjs.com`!

PR #5579, opened 2026-01-05
fix(bun): fail explicitly when body already read

wooorm-arcjet:
Instead of failing silently, this errors when the body cannot be read, when it was read already.

Closes GH-5515.

### DIFF
--- a/arcjet-bun/test/index.test.ts
+++ b/arcjet-bun/test/index.test.ts
@@ -67,55 +67,46 @@ test("should support `sensitiveInfo`", async function () {
   assert.equal(response.status, 200);
 });
 
-// TODO(GH-5515): this currently silently reads the body as `""`.
-// Should emit an error log instead.
-// Along the lines of:
-// ```js
-// if (request.bodyUsed) {
-//   log.error("Body already read. See <https://docs.arcjet.com/troubleshooting#error-body-already-read-or-body-is-unusable>");
-//   return;
-// }
-// ```
-// test("should emit an error log when the body is read before `sensitiveInfo`", async function () {
-//   const restore = capture();
-//   let body: string | undefined;
-//   let parameters: Array<unknown> | undefined;
+test("should emit an error log when the body is read before `sensitiveInfo`", async function () {
+  const restore = capture();
+  let body: string | undefined;
+  let parameters: Array<unknown> | undefined;
 
-//   const arcjet = arcjetBun({
-//     key: exampleKey,
-//     rules: [sensitiveInfo({ deny: ["EMAIL"], mode: "LIVE" })],
-//     log: {
-//       debug() {},
-//       error(...values) {
-//         parameters = values;
-//       },
-//       info() {},
-//       warn() {},
-//     },
-//   });
+  const arcjet = arcjetBun({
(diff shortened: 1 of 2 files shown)


arcjet-bun/test/index.test.ts
```diff
@@ -67,55 +67,46 @@ test("should support `sensitiveInfo`", async function () {
...
+
+  await server.stop();
+  restore();
+
+  assert.equal(body, "My email is alice@arcjet.com");
+  assert.equal(response.status, 200);
+  assert.deepEqual(parameters, [
+    "Body already read. See <https://docs.arcjet.com/troubleshooting#error-body-already-read-or-body-is-unusable>",
+  ]);
+});
```
qw-in, line 109: This appears to be asserting on a log message. Would it not make sense to return an error reason from the rule itself? Then we could assert on that instead

wooorm-arcjet: That also happens: https://github.com/arcjet/arcjet-js/blob/8d4c7ef33dff5e032d77be4ab6a1de4392fff054/arcjet/index.ts#L1835-L1847

wooorm-arcjet: I guess, these tests are covering entire servers. Talking over HTTP.
So it’s a bit simpler to have a basic same server, to start up/wind down for each test case. That’s at the bottom of this file.
And then normally there would not be log messages. But when there are, they are checked.

qw-in: I pulled it down and it doesn't look like it is an error response that we get. Is that correct? https://github.com/arcjet/arcjet-js/pull/5590

wooorm-arcjet: indeed, it is not: the error decision goes into to a 200.
GH-5590 is possible. I am okay with that!
But I didn’t, because typically it is recommended to “fail open” right? Not to 500 if Arcjet is down?

qw-in: > typically it is recommended to “fail open” right
 
 I think we usually suggest that as a starting point, but we offer the flexibility to choose based on the use case
 
Here (in the tests) we probably want to differentiate between an `ERROR` & `ALLOW` I'd imagine? Could be useful to catch changes

wooorm-arcjet: okay, good to me!

blaine-arcjet:
The title of this PR is "fail explicitly" but the implementation seems to just log. Do we want to reword the PR title?

wooorm-arcjet:
There is later handling that does:

```js
      const body = await context.getBody();
      if (typeof body === "undefined") {
        return new ArcjetRuleResult({
          state: "NOT_RUN",
          conclusion: "ERROR",
          reason: new ArcjetErrorReason(
            "Couldn't read the body of the request to perform sensitive info identification.",
          ),
          ...
```

So I think that the title is still correct.

It is probably good to propagate errors there.
To stop using `undefined` but instead catch?

I’ll merge this, and if you think throwing/catching is a good idea, we can do that in a future issue/PR?

wooorm-arcjet:
Thanks for the review!

PR #5580, opened 2026-01-05
feat(node): expose body at `request.body` after reading

wooorm-arcjet:
If the Arcjet Node SDK is used, and it needs access to the body, users cannot read the body too. Neither before nor after. This is because Node stream can be read once.

It is possible to use a tool like `body-parser` to read the body *before* Arcjet. That works because after reading it stashes what it found as `request.body`. Arcjet then takes that.

It is possible to recommend that users do that. Use `body-parser` or when they read the stream manually, to stash the body there.

But with other Arcjet SDKs, only the inverse order works: call `.protect` first, then (optionally) read the body later. This ordering seems more natural.

In the case of Arcjet reading the streaming body, it can also stash the body at `request.body`. Then it works like other tools, which can be documented.

Closes GH-5516.

### DIFF
--- a/arcjet-node/index.ts
+++ b/arcjet-node/index.ts
@@ -461,7 +461,7 @@ export default function arcjet<
         const getBody = async () => {
           try {
             // If request.body is present then the body was likely read by a package like express' `body-parser`.
-            // If it's not present then we attempt to read the bytes from the IncomingMessage ourselves.
+            // If it's not present then we read the bytes from the IncomingMessage ourselves.
             if (typeof request.body === "string") {
               return request.body;
             } else if (
@@ -494,6 +494,7 @@ export default function arcjet<
                 limit: 1048576,
                 expectedLength,
               });
+              request.body = body;
               return body;
             }
 
--- a/arcjet-node/test/index.test.ts
+++ b/arcjet-node/test/index.test.ts
@@ -114,43 +114,38 @@ test("`@arcjet/node`", async function (t) {
     },
   );
 
-  // TODO(GH-5516): this hangs indefinitely.
-  // Fix bug, probably by calling `request.clone()`?
-  // await t.test("should support reading body after `sensitiveInfo`", async function () {
-  //   const restore = capture();
-  //   let body = "";
+  await t.test(
+    "should support accessing body after `sensitiveInfo`",
+    async function () {
+      const restore = capture();
+      let body = "";
 
-  //   const arcjet = arcjetNode({
-  //     key: exampleKey,
-  //     rules: [sensitiveInfo({ deny: ["EMAIL"], mode: "LIVE" })],
-  //   });
+      const arcjet = arcjetNode({
+        key: exampleKey,
+        rules: [sensitiveInfo({ deny: ["EMAIL"], mode: "LIVE" })],
+      });
 
-  //   const { server, url } = await createSimpleServer({
-  //     async after(request) {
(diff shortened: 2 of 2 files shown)


arcjet-node/index.ts
```diff
@@ -494,6 +494,7 @@ export default function arcjet<
                 limit: 1048576,
                 expectedLength,
               });
+              request.body = body;
```
blaine-arcjet, line 497: I'm concerned about modifying the user's environment. This is an express/fastify-ism that might not work with other frameworks.

How could this be done in a different way to avoid modifying a user-object? Maybe we could cache it in the context or closure?

wooorm-arcjet: And nuxt/h3 and friends (e.g., https://github.com/oscartbeaumont/h3/blob/845eecaaaeffdb415321f76744e31d55074f37b8/src/utils/body.ts#L40-L46, https://github.com/arcjet/arcjet-js/pull/5305).

I don’t think there is a better alternative.

Other then that users themselves make `request` into a https://github.com/mcollina/cloneable-readable and pass that to arcjet and also read from that themselves.

Or make `body` an extra parameter that folks have to pass themselves.

Or, pass `body` back after `.protect` in a decision

blaine-arcjet: > I don’t think there is a better alternative.

What options did you consider?

wooorm-arcjet: The issue is first that *users* cannot read the stream in combination with Arcjet.

I think the above alternatives fall into the categories:

1. do not read request bodies, let users do it
   1. require users to use `body-parser` and the like which set `request.body`
   2. require users to read the body and pass it as a parameter to `protect`
2. read request bodies, pass it back
   1. this pr, like other tools, set `request.body` when reading
   2. pass it back from `protect` as `decision.body`
3. require users to pass a cloneable stream (https://github.com/mcollina/cloneable-readable)

My preference is to work like other tools and, after reading, patch `body` (ref: https://github.com/arcjet/arcjet-js/pull/5305)

The follow up problem, which is I think what you mean with your point “Maybe we could cache it in the context or closure?” helps more with adding `body` to filters (GH-5505), as in: 2 rules that read the body.
This is indeed what I’d want: read streams once. Ref https://github.com/arcjet/arcjet/pull/6412#issuecomment-3641545142.
Using the `request.body` field in Node-stream-based SDKs solves that too though.
Though the Web-stream-based SDKs don’t have that (yet).

blaine-arcjet: I think those capture the alternatives. I don't think we should modify anything (objects, env) on the user's behalf (this likely needs to be added to the SDK philosophy).

We already have a concept of additional properties passed to the SDK so we could achieve (1) very easily and it would work hand-in-hand with most Node.js frameworks that parse bodies.

wooorm-arcjet: The existing fields-passed-by-users-to-protect have the following problems:

* those are passed to the server, there would need to be careful splitting of what is passed and what not
  * there would need to be some handling globally (I mean `arcjet.protect`, in opposition of `rule.protect`) to never send a list (currently only `body`) with `decide` or `report`
* those fields are currently all *required*, this given body would be optional, which does not work with all the `Properties extends Empty ? [] : [Properties]` stuff

qw-in:
@wooorm-arcjet "you can close it"

PR #5581, opened 2026-01-05
build: only test built JavaScript files

wooorm-arcjet:
This prevents running test suites twice on Node 24 and later, which otherwise read TypeScript files too.

Closes GH-4477.

### DIFF
--- a/analyze-wasm/package.json
+++ b/analyze-wasm/package.json
@@ -56,8 +56,8 @@
     "build": "npm run build:jco && npm run build:rollup",
     "lint": "eslint .",
     "prepublishOnly": "npm run build",
-    "test-api": "node --test",
-    "test-coverage": "node --experimental-test-coverage --test",
+    "test-api": "node --test -- test/*.test.js",
+    "test-coverage": "node --experimental-test-coverage --test -- test/*.test.js",
     "test": "npm run build && npm run lint && npm run test-coverage"
   },
   "dependencies": {},
--- a/analyze/package.json
+++ b/analyze/package.json
@@ -40,8 +40,8 @@
     "build": "rollup --config rollup.config.js",
     "lint": "eslint .",
     "prepublishOnly": "npm run build",
-    "test-api": "node --test",
-    "test-coverage": "node --experimental-test-coverage --test",
+    "test-api": "node --test -- test/*.test.js",
+    "test-coverage": "node --experimental-test-coverage --test -- test/*.test.js",
     "test": "npm run build && npm run lint && npm run test-coverage"
   },
   "dependencies": {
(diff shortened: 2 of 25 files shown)


qw-in (approved):
Makes sense to me 👍

analyze-wasm/package.json
```diff
@@ -56,8 +56,8 @@
     "build": "npm run build:jco && npm run build:rollup",
     "lint": "eslint .",
     "prepublishOnly": "npm run build",
-    "test-api": "node --test",
-    "test-coverage": "node --experimental-test-coverage --test",
+    "test-api": "node --test -- test/*.test.js",
+    "test-coverage": "node --experimental-test-coverage --test -- test/*.test.js",
```
blaine-arcjet, line 60: I just ran into this on my open source project. This relies on shell expansion, but you can rely on node.js expansion by avoiding `--` and wrapping in quotes `"test/*.test.js"` (of course it also needs escapes).

See the bottom part of https://nodejs.org/docs/latest-v24.x/api/test.html#running-tests-from-the-command-line

wooorm-arcjet: - I use the `--` more to be clear that there won’t be more flags afterwards. That’s my preference. Fine to do without it though
- I considered and tried quoting them, to indeed let Node do it, not the shell. I have a slight preference for no escaped quotes in quotes. Either works and I am fine with either!

blaine-arcjet: Generally, I prefer not escaping if possible but I worry about different shells having different expansion rules. Specifically, I think the `test/*.test.js` pattern has problems if run on Windows.

wooorm-arcjet: Wellll I did try it again, and sadly, Node glob expansion does not work on Node 20, which is in the devcontainer / one of our test targets 😓

Fails:
* `"node --test \"test/*.test.js\""`
* `"node --test -- \"test/*.test.js\""`

Works:
* `"node --test test/*.test.js"`
* `"node --test -- test/*.test.js"`

The docs you linked to on v20 are also different: https://nodejs.org/docs/latest-v20.x/api/test.html#running-tests-from-the-command-line.

blaine-arcjet: Yeah, it looks like the globbing was added in Node 22

wooorm-arcjet: OK, so I think it is not possible to depend on globbing-in-quotes yet and have to depend the shell to expand.
When the lowest supported Node is 22, we could start doing that.
But at that point, I think we can switch to reading the TS files directly instead. And maybe not even building the test TS files.

wooorm-arcjet:
Thanks for the reviews!

PR #5582, opened 2026-01-05
build: type check all TypeScript files

wooorm-arcjet:
Previously, explicit `include` patterns were used. These did not include all existing TypeScript files, affecting mostly test files. This meant that TypeScript syntax was not allowed
(Rollup only strips TS syntax in included files, so there was an error). And some files were not checked, such as `env/test`, `sprintf/test`

The default for `include` is all TS files, except those in `exclude`, which in `tsconfig.base.json` is `node_modules/`. Removing this field when not needed means all TS files are checked and TS syntax is allowed in them. It is needed for the 4 packages where `.d.ts` files are manually authored.

### DIFF
--- a/.github/workflows/reusable-examples.yml
+++ b/.github/workflows/reusable-examples.yml
@@ -16,7 +16,7 @@ jobs:
             api.github.com:443
             github.com:443
             objects.githubusercontent.com:443
-            release-assets.githubusercontent:443
+            registry.npmjs.org:443
           disable-sudo-and-containers: true
           egress-policy: block
       - uses: actions/checkout@8e8c483db84b4bee98b60c0593521ed34d9990e8 # v6.0.1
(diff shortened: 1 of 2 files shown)


wooorm-arcjet:
It may be this: There are some differences between how TS itself reads `tsconfig.json`s, and how Rollup reads the tsconfig:

- https://github.com/arcjet/arcjet-js/blob/8d4c7ef33dff5e032d77be4ab6a1de4392fff054/tsconfig.base.json#L9
- https://github.com/arcjet/arcjet-js/blob/8d4c7ef33dff5e032d77be4ab6a1de4392fff054/rollup-config/index.js#L210-L215

I did this on a clean copy of the repo, and it indeed looks like everything works fine

wooorm-arcjet:
Thanks for the review!

Issue #5058, opened 2025-09-03
Improve types around custom entities for sensitive info rule


wooorm-arcjet:
Same as GH-4883 / GH-4982, but applied to the sensitive info rule in `arcjet` core.

Especially of interest: https://github.com/arcjet/arcjet-js/pull/4883#issuecomment-3248406814.

That would close GH-1455.

PR #5584, opened 2026-01-05
refactor(arcjet)!: improve types around custom entities

wooorm-arcjet:
This applies the TypeScript changes of GH-4883 into Arcjet core, for the `sensitiveInfo` rule. Which is slightly different, having no `entities` but an `allow` or `deny` field, and not having a `replace` method. But otherwise largely the same.

Closes GH-1455.
Related-to: GH-4883.
Closes GH-5058.

### DIFF
--- a/arcjet/index.ts
+++ b/arcjet/index.ts
@@ -853,7 +853,6 @@ export type EmailOptions =
  *
  * @template DetectedEntities
  *   Custom entity names that are returned from `detect` and optionally listed in `allow`.
- *   in `allow` or `deny`.
  * @template ListedEntities
  *   Entity names that can be listed in the `allow` field.
  */


arcjet/index.ts
```diff
@@ -847,46 +847,24 @@ export type EmailOptions =
...
-
 /**
  * Configuration for the sensitive info detection rule to allow certain
  * sensitive info types and deny others.
  *
- * @template Detect
- *   Custom detection function to identify sensitive information.
+ * @template DetectedEntities
+ *   Custom entity names that are returned from `detect` and optionally listed in `allow`.
+ *   in `allow` or `deny`.
```
blaine-arcjet:
```suggestion
```

blaine-arcjet (approved):
Changes look good (same as #4883). All the types propagate as I expect in my IDE 👍 

One typo to remove in the docs, but otherwise good to go.

wooorm-arcjet:
Thanks for the review!

PR #5588, opened 2026-01-06
build: update `allowed-endpoints` for step security

wooorm-arcjet:
This makes CI work again, by listing all connections that are needed.

For some of these, things still work if StepSecurity blocks them, except that StepSecurity then fails. For example, `binaries.prisma.sh` or `unpkg.com` or fonts. I did not find a way for StepSecurity to block those, but not treat that block as a failure.


blaine-arcjet (approved):

wooorm-arcjet:
Thanks for the review!

Issue #4562, opened 2025-07-04
Configurable reasonable max-size of body


wooorm-arcjet:
> I think this requires a long-form discussion so you should probably open an issue.
> — @blaine-arcjet in https://github.com/arcjet/arcjet-js/pull/4538#discussion_r2183902505

`arcjet-nest` and `arcjet-node` contain the value `1048576` as a max-length of a body to read.
Bodies bigger than that, crash.
There could be cases where someone might want to crash earlier. Or later.

blaine-arcjet:
> Bodies bigger than that, crash.

To clarify, a larger body will produce an Errored Rule Result.

wooorm-arcjet:
Crosslinking https://svelte.dev/docs/kit/adapter-node#Environment-variables-BODY_SIZE_LIMIT.

Also, this becomes more of a thing with GH-5505.

wooorm-arcjet:
I think this can be closed as not planned with GH-5678 and GH-5679.
Users that want certain limits can use `@arcjet/body` to read Node streams and web streams and limit them:

```js
const value = readBodyWeb(request, { limit: 1024 })
const decision = await arcjet.protect(request, { sensitiveInfoValue: value })
```

PR #5589, opened 2026-01-06
feat!: add support for limits to web streams

wooorm-arcjet:
Arcjet SDKs using Node.js streams support limiting, timeouts, and length conformance checks.
SDKs using web streams had no such protections.
This PR adds a `readBodyWeb` function from `@arcjet/body` that works similar to `readBody`. The web stream based SDKs use this function.

Closes GH-5517.

### DIFF
--- a/arcjet-bun/test/index.test.ts
+++ b/arcjet-bun/test/index.test.ts
@@ -308,7 +308,7 @@ test("should not support `sensitiveInfo` 5 megabytes of data", async function ()
   assert.equal(response.status, 200);
   assert.deepEqual(parameters, [
     "failed to get request body: %s",
-    "request entity too large",
+    "Cannot read stream whose expected length exceeds limit",
   ]);
 });
 
--- a/arcjet-deno/test/index.test.ts
+++ b/arcjet-deno/test/index.test.ts
@@ -310,7 +310,7 @@ test("should not support `sensitiveInfo` 5 megabytes of data", async function ()
   assert.equal(response.status, 200);
   assert.deepEqual(parameters, [
     "failed to get request body: %s",
-    "request entity too large",
+    "Cannot read stream whose expected length exceeds limit",
   ]);
 });
 


arcjet-deno/index.ts
```diff
@@ -335,11 +336,23 @@ export default function arcjet<
...
-            // Awaited to throw if it rejects and we'll just return undefined
-            const body = await clonedRequest.text();
+            let expectedLength: number | undefined;
+            const expectedLengthString = request.headers.get("content-length");
+            if (typeof expectedLengthString === "string") {
+              try {
+                expectedLength = parseInt(expectedLengthString, 10);
+              } catch {
+                // Empty.
+              }
```
qw-in: Does `parseInt` throw? I thought it returned `NaN` for a parse error

blaine-arcjet: I think you are right. Not something I usually think about!

arcjet-astro/internal.ts
```diff
@@ -333,11 +334,23 @@ export function createArcjetClient<
...
+            const expectedLengthString = request.headers.get("content-length");
+            if (typeof expectedLengthString === "string") {
+              try {
+                expectedLength = parseInt(expectedLengthString, 10);
+              } catch {
+                // Empty.
+              }
+            }
+
+            const body = await readBodyWeb(clonedRequest.body!, {
```
blaine-arcjet: Why `!`?

body/index.ts
```diff
@@ -22,14 +22,78 @@ type EventHandlerLike = (
...
+      if (length !== undefined && received !== length) {
+        reject(new Error("request size did not match content length"));
+        return;
+      }
+
+      // Need to call it one final time to flush any remaining chars.
+      buffer += decoder.decode();
+      resolve(buffer);
+    })().then(undefined, reject);
+  });
```
blaine-arcjet: Why did you choose this implementation instead of [pipeTo](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/pipeTo) a [WritableStream](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream) with an AbortSignal for timeout?

wooorm-arcjet: Also very possible, I tried various things, with `stream.getReader` and such too. I settled on the async iterable. But I just pushed writable stream / pipeTo / signal!

arcjet-astro/internal.ts
wooorm-arcjet: Body is `null` for `HEAD`/`GET`. Which causes this to error, which is then caught.

Making the stream optional in `readBodyWeb` is different from `readBody`, so I don’t like that.

Extra handling in each adapter may be good, to have better errors for HEAD/GET, but then it would be good to have that in the Node-stream-based ones too?

arcjet-deno/index.ts
wooorm-arcjet: removed them, also in the node-stream-based versions

wooorm-arcjet: See GH-5599

arcjet-astro/internal.ts
blaine-arcjet: I'd rather we check for existence and return early, which would allow for a different error message.

body/README.md
```diff
@@ -1,3 +1,6 @@
+<!-- trunk-ignore-all(markdownlint/MD001) -->
+<!-- trunk-ignore-all(markdownlint/MD024) -->
+
```
blaine-arcjet, line 3: What linting is failing now?

body/index.ts
```diff
@@ -22,14 +22,89 @@ type EventHandlerLike = (
...
+ * @param stream
+ *   Stream.
+ * @param options
+ *   Configuration (required).
+ * @returns
+ *   Promise that resolves to a concatenated body.
+ */
+export async function readBodyWeb(
+  stream: ReadableStream,
+  // TODO(@wooorm-arcjet): make optional.
```
blaine-arcjet: Optional or defaulted?

body/index.ts
```diff
@@ -22,14 +22,89 @@ type EventHandlerLike = (
...
+
+  const controller = new AbortController();
+
+  // Ensure that we do not wait forever if the stream is incorrectly configured.
+  let timeout: ReturnType<typeof setTimeout> | undefined = setTimeout(
+    function () {
+      controller.abort(new Error("received no body chunks after 100ms"));
+    },
+    100,
+  );
```
blaine-arcjet: Is https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/timeout_static available everywhere we need?

body/index.ts
```diff
@@ -22,14 +22,89 @@ type EventHandlerLike = (
...
+
+  const decoder = new TextDecoder("utf-8");
+  let received = 0;
+  let buffer = "";
+
+  await stream.pipeTo(
```
blaine-arcjet, line 95: Do we need to check the result to know if it succeeded?

body/index.ts
```diff
@@ -22,14 +22,89 @@ type EventHandlerLike = (
...
+    new WritableStream({
+      write(chunk: Uint8Array) {
+        if (timeout) {
+          clearTimeout(timeout);
+          timeout = undefined;
+        }
```
blaine-arcjet: Not needed if we can use `AbortSignal.timeout()`

body/index.ts
```diff
@@ -22,14 +22,89 @@ type EventHandlerLike = (
...
+          throw new Error("request entity too large");
+        }
+
+        buffer += decoder.decode(chunk, { stream: true });
+      },
+    }),
+    {
+      // Needed for Bun not to hang on a thrown error in `write`.
+      // Works OK in Deno and Node without.
+      preventCancel: true,
```
blaine-arcjet: Do we want to pass this through options so it can only be set for bun? Perhaps pass `runtime` through? Or does it not make a difference on Deno and Node?

blaine-arcjet (changes requested):
The stream implementation looks great. I had some follow-up comments.

body/index.ts
wooorm-arcjet: Possibilities:

- could not check `limit` if missing
- could default `limit` to `Infinity`, which is effectively the same
- could default to a particular size, such as 1mb, which is now what Arcjet uses everywhere

I think users of `@arcjet/body` may be OK defaulting to `Infinity`.
For use through Arcjet, it makes more sense to default to 1mb, probably even 128kb (as this is text and that can fit a book), if configurable.

A PR for GH-4562 could solve this one too.

body/index.ts
wooorm-arcjet: That code appears cleaner, but I would say not. Available from April 2024. It’s not that essential code, I’m fine with using a regular timeout.

Also: the timeout needs to be cancelled: it’s not just a 100ms timeout, it’s 100ms for the *first* chunk.

body/index.ts
wooorm-arcjet: Apparently returns `Promise<undefined>`, so no

body/index.ts
wooorm-arcjet: I think it is, this is about time-to-first-chunk, not time-total?

body/index.ts
wooorm-arcjet: See https://github.com/arcjet/arcjet-js/pull/5601 to discuss this

body/index.ts
wooorm-arcjet: it now works without! All of a) `preventCancel: true`, b) `preventCancel: false`, and c) omitting work.
I opted for c).
Not entirely sure what made it work

body/README.md
wooorm-arcjet: Fails for same headings (`Parameters` and the like), and for going from `h3` to `h6`.

I think a `h4` for `Parameters` / `Returns` / `Type` / `Fields` etc is too much.
And a `p` with a `strong` in it, or so, is not better.

arcjet-nest/index.ts
```diff
@@ -479,7 +479,11 @@ function arcjet<
               const headers = new ArcjetHeaders(request.headers);
               const expectedLengthStr = headers.get("content-length");
               if (typeof expectedLengthStr === "string") {
-                expectedLength = parseInt(expectedLengthStr, 10);
+                try {
+                  expectedLength = parseInt(expectedLengthStr, 10);
```
blaine-arcjet: Fixed in the other PR, right?

arcjet-node/index.ts
```diff
@@ -482,7 +482,11 @@ export default function arcjet<
               const headers = new ArcjetHeaders(request.headers);
               const expectedLengthStr = headers.get("content-length");
               if (typeof expectedLengthStr === "string") {
-                expectedLength = parseInt(expectedLengthStr, 10);
+                try {
+                  expectedLength = parseInt(expectedLengthStr, 10);
```
blaine-arcjet: Fixed in the other PR?

arcjet-nuxt/internal.ts
```diff
@@ -338,7 +338,11 @@ function createGetBody(state: State, event: ArcjetH3Event) {
         const headers = new ArcjetHeaders(event.node.req.headers);
         const expectedLengthStr = headers.get("content-length");
         if (typeof expectedLengthStr === "string") {
-          expectedLength = parseInt(expectedLengthStr, 10);
+          try {
+            expectedLength = parseInt(expectedLengthStr, 10);
```
blaine-arcjet: Same as others

body/index.ts
```diff
@@ -22,14 +22,86 @@ type EventHandlerLike = (
...
+ * @param stream
+ *   Stream.
+ * @param options
+ *   Configuration (required).
+ * @returns
+ *   Promise that resolves to a concatenated body.
+ */
+export async function readBodyWeb(
+  stream: ReadableStream,
+  // TODO(@wooorm-arcjet): make optional.
```
blaine-arcjet: The other PR makes the node variant options optional so we probably want to mirror that here then

arcjet-nest/index.ts
wooorm-arcjet: yes, solved with rebase + applying those changes!

arcjet-node/index.ts
wooorm-arcjet: solved with rebase + applying those changes!

body/index.ts
wooorm-arcjet: solved with rebase + applying those changes!

wooorm-arcjet:
Applied the other things to this PR, from other PRs that already landed!

PR #5994, opened 2026-04-14
chore(proto): sync generated proto

qw-in:
Sync protocol.


qw-in:
@davidmytton not sure if intentional but I don't think @arcjet-review has ~write / review access~ codeowner

davidmytton (approved):

PR #6014, opened 2026-04-24
fix: update @bufbuild/protobuf to 2.12.0 and add root override to fix Bun build

qw-in:
When Bun ignores the npm lockfile and resolves dependencies independently, @connectrpc/connect's peer dep ^2.7.0 resolves to @bufbuild/protobuf@2.12.0 at root while workspaces pinned 2.11.0. Version 2.12.0 added utf8Validation which 2.11.0 lacks, causing TypeScript type mismatch errors during build.

- Update @bufbuild/protobuf from 2.11.0 to 2.12.0 in transport and protocol
- Add @bufbuild/protobuf override to root package.json to force consistent resolution across all workspaces under Bun

Agent-Logs-Url: https://github.com/arcjet/arcjet-js/sessions/e80853e0-598e-47b7-842f-8790866a010f

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -298,7 +298,7 @@
         "@arcjet/eslint-config": "1.4.0",
         "@arcjet/rollup-config": "1.4.0",
         "@rollup/wasm-node": "4.59.0",
-        "astro": "^6.1.6",
+        "astro": "6.1.6",
         "eslint": "9.39.3",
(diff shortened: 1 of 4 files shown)


qw-in:
@arcjet/js-sdk-team looks like I need a codeowner review

davidmytton:
Config changed to allow the auto merge.

PR #6017, opened 2026-04-29
docs: refresh root, next, and guard READMEs for guards release

qw-in:
Refresh readmes

### DIFF
--- a/arcjet-guard/README.md
+++ b/arcjet-guard/README.md
@@ -60,17 +60,17 @@ like `@arcjet/next` or `@arcjet/node` instead.
 Guards share some features with the request SDKs but are designed for
 non-HTTP contexts. Here's what's available where:
 
-| Feature | Request SDKs | `@arcjet/guard` |
-| --- | :---: | :---: |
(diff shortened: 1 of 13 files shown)


arcjet-guard/README.md
```diff
@@ -36,6 +57,21 @@ This is the [Arcjet][arcjet] Guards SDK.
 
 ## Features
 
+Guards share some features with the request SDKs but are designed for
+non-HTTP contexts. Here's what's available where:
+
+| Feature | Request SDKs | `@arcjet/guard` |
+| --- | :---: | :---: |
+| Rate Limiting | ✅ | ✅ |
```
arcjet-rei: Is there a reason this table is formatted differently (not justified) than the others?

arcjet-rei (approved):
One nit (which should be caught by prettier, or at least would be on the monorepo), otherwise looks good.

arcjet-guard/README.md
qw-in: Probably trunk, one moment

PR #6059, opened 2026-06-08
feat(guard): add experimental_moderateContent rule

qw-in:
Mirror detectPromptInjection as a new experimental content moderation rule, exported as experimental_moderateContent. Adds the rule factory, the RuleResultModerateContent / config / with-config / with-input types, proto conversion (both directions + reason maps), exports, and the regenerated decide v2 proto bindings.

No moderation model is wired up server-side yet, so the rule currently returns an error result (fail open).


qw-in (commented):
Addressed Arcjet Review feedback:

1. **`detected` field**: Added `detected: boolean` to `RuleResultModerateContent` type and wired it through `resultFromProto`. Tests updated to assert it.

2. **128 KiB input limit**: Not adding a client-side check — the server already enforces this (error code AJ1141) and `detectPromptInjection` follows the same pattern (no client-side length check). Adding one only for moderation would be inconsistent.

3. **Fail-open behavior**: This is intentional and documented via the `experimental_` prefix and JSDoc. The ADR explicitly addresses this: the rule is non-functional until a model is wired up, callers see ERROR, and the prefix signals instability.

arcjet-rei (approved):

PR #6090, opened 2026-06-18
feat: require Node.js >=22.21.0 and drop EOL Node.js 20

qw-in:
## Summary

Drops support for the now end-of-life Node.js 20 and raises the minimum supported Node.js to `>=22.21.0 <23 || >=24.5.0` across all packages. This lands ahead of the proxy-support change (#6089) and deliberately pulls in the exact engine range that work requires, so the engines don't have to change twice.

The exact range is driven by `@arcjet/transport`: its proxy support relies on the built-in proxy support of the Node.js HTTP agent, which is only available on Node.js `>=22.21.0` and, on the 24 line, `>=24.5.0`. Node.js 23 is not supported. Because every package depends (directly or transitively) on `@arcjet/transport`, the requirement is applied to all of them rather than only those that import it directly. Anyone tracking an active LTS release is unaffected.

This is intentionally just the Node-version policy slice — no proxy/transport implementation (that's #6089).

## Changes

- **Engines** → `>=22.21.0 <23 || >=24.5.0` on every workspace package (incl. `@arcjet/guard`). Marked `feat:` so release-please surfaces the new requirement in the SDK release notes.
- **CI matrices** now run Node.js **22, 24, 26** (dropping 20); standalone Deno/examples setup steps bumped 20 → 22.
- **`@types/node`** pinned to the latest 22.x (`22.19.21`) so type checking reflects the minimum supported runtime; renovate keeps it on the 22.x line via a dedicated `<23` rule while the node/actions group tracks up to `<27`.
- **Coverage thresholds** enabled now that the required `--test-coverage-*` flags are available on supported Node.js versions (`@arcjet/ip` at 100%; others ratcheted to current).
- **Docs**: `@arcjet/guard` README runtime table + footnote and CONTRIBUTING, a `@arcjet/transport` README section, the v22 `process.version` doc link, example README cleanups (removed the obsolete `--env-file` caveats), and `examples/remix-express` engines → `>=22.0.0`.

The framework adapters (`@arcjet/astro`, `bun`, `deno`, `nuxt`, `remix`, `react-router`) carry no `engines`/`@types/node` fields, matching existing repo convention.

## Verification

- `npm run build` (33 packages) ✅
- `npm run lint` (35 packages) ✅
- `@arcjet/guard`: `oxlint` ✅, `tsgo` typecheck against 22.x types ✅
- Coverage gates verified passing on Node 22


davidmytton (approved):

arcjet-rei (approved):
This makes more sense than bundling an unrelated semver-minor change into the proxy support, and it sets us up for 26 to go LTS in the fall. Nice! After this merges, I'll update #6089 once this merges to main.

PR #6096, opened 2026-06-23
feat(guard): error/warning decision model with hasFailedOpen() gate

qw-in:
Replace the single conflated `hasError()` boolean on a guard decision with a two-axis model — scope (decision vs. rule) × severity (error vs. warning) — consistent with the Python and Go SDKs. Severity is carried by channel, not inferred from placement: an error is a result (a rule or the decision could not be processed); a warning is a `warnings` entry (processed correctly, but fix it).

Surfaces:
- decision.warnings surfaces the GuardResponse.errors payload as decision-level warnings instead of reading it only to flip a boolean and dropping it. Validated at the SDK boundary (non-string code/message from a malformed Connect-JSON response fall back to defaults).
- result.warnings adds the per-rule warning channel, empty until the Decide service emits per-rule diagnostics.
- decision.errorResults() returns the results that could not be processed, including the synthetic decision-level error result. Replaces the "did any rule fail" boolean with the data.
- decision.hasFailedOpen() is the fail-closed gate: true when the conclusion is ALLOW only because a rule or the decision could not be processed.
- decision.hasError() is deprecated (still works; triggers on warnings or errors) for one release cycle.

guard() never throws for runtime degradation: transport, parse, and missing-decision failures all return a fail-open ALLOW carrying a single synthetic error result, so hasFailedOpen()/errorResults() see the failure.

errorResults() is computed once and shared by hasFailedOpen()/hasError() rather than re-filtering on each call.

### DIFF
--- a/README.md
+++ b/README.md
@@ -847,8 +847,10 @@ const decision = await arcjet.guard({ label: "tools.weather", rules: [...] });
 decision.conclusion;   // "ALLOW" or "DENY"
 decision.reason;       // "RATE_LIMIT", "PROMPT_INJECTION", "SENSITIVE_INFO", "CUSTOM", "ERROR", etc.
 
-// Layer 2: error detection
-decision.hasError();   // true if any rule errored or the server reported diagnostics
+// Layer 2: error/warning detection
+decision.hasFailedOpen(); // true if ALLOW only because a rule/decision could not be processed (fail-closed gate)
+decision.errorResults();  // results that errored (rules or the decision that could not be processed)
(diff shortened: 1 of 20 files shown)


davidmytton (approved):
Makes sense! Should we also include the README updates here as well [as suggested on Slack](https://arcjethq.slack.com/archives/C0ARWVAJHDG/p1782252964510569?thread_ts=1782245836.198639&cid=C0ARWVAJHDG)?

qw-in:
> Should we also include the README updates here

Updated

PR #6097, opened 2026-06-25
feat(arcjet-guard): export experimental_moderateContent from node and fetch entrypoints

estellajaysong:
## Summary

Export `experimental_moderateContent` (and its `RuleResultModerateContent` / `ExperimentalModerateContentConfig` types) from both the `node` and `fetch` entrypoints, so the content-moderation rule is reachable from the public API on every runtime.

> A follow-up refactor that centralized the shared re-export list into an internal `src/public.ts` was reverted; each entrypoint keeps its own curated re-export list as before.

## Verification

- The emitted `node.d.ts` and `fetch.d.ts` public export lists are **byte-identical to each other** and both now include `experimental_moderateContent` and its associated types.
- Unit tests: 263/263 pass.
- Fetch runtime suite: 26/26 pass. Node runtime suite: all in-memory + h2c cases pass. (The one TLS-self-signed-cert runtime test fails locally with `ERR_SSL_CA_MD_TOO_WEAK` — a pre-existing fixture issue where the mock server mints a SHA-1 CA cert without `-sha256`; unrelated to this change and not touched by it.)


arcjet-guard/src/public.ts
```diff
@@ -0,0 +1,66 @@
...
+  experimental_moderateContent,
+  localDetectSensitiveInfo,
+  defineCustomRule,
+
+  // Transport-agnostic factory
+  launchArcjetWithTransport,
+
+  // Internal
+  _launchWithTransportFactory,
+} from "./index.ts";
```
qw-in: I'd prefer to just export them from the two entrypoints. Its not expensive

qw-in (approved):
Thanks!

PR #6103, opened 2026-06-26
deps(protocol): vendor TypeID generation, drop typeid-js

qw-in:
Replace the external typeid-js dependency with a small, dependency-free inline TypeID generator in @arcjet/protocol. This removes typeid-js and its transitive uuid dependency, resolving the uuid <11.1.1 advisory in the dependency tree (root npm audit: 8 → 6 findings, uuid gone).

We only ever mint new local request IDs with a fixed prefix, so the vendored module covers generation only: a UUIDv7 (RFC 9562) encoded as a 26-character Crockford base32 suffix, matching the TypeID spec (https://github.com/jetify-com/typeid). It mirrors the implementation already vendored and shipping in the Arcjet Python SDK, so both SDKs produce identical IDs.

Tests are ported from the Python SDK's test_typeid.py (format, UUIDv7 version/variant, first-char bound, uniqueness, time-sortability, zero/max timestamp, low/max entropy, alphabet, and fuzz/round-trip invariants). Also drops the now-unused typeid-js alias from the deno example.

Validated: 13 ported tests pass, protocol lint passes, typeid-js + uuid removed from the root and deno lockfiles.

---

@arcjet-rei would appreciate a review since you added it in arcjet-py!

### DIFF
--- a/examples/deno-sensitive-info/deno.json
+++ b/examples/deno-sensitive-info/deno.json
@@ -24,7 +24,6 @@
     "@bufbuild/protobuf": "npm:@bufbuild/protobuf@2.11.0",
     "@connectrpc/connect": "npm:@connectrpc/connect@2.1.1",
     "@connectrpc/connect-node": "npm:@connectrpc/connect-node@2.1.1",
-    "arcjet": "../../arcjet/index.js",
-    "typeid-js": "npm:typeid-js@1.2.0"
+    "arcjet": "../../arcjet/index.js"
   }
 }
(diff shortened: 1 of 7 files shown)


qw-in:
~Looks like I missed some instances of `typeid-js`. Will pick those up as well!~

Actually, just in examples or arcjet-guard picking it up from the published version

arcjet-rei (approved):
This is a good idea for the same reason that I decided to do something similar in Python – there's a bunch of additional stuff under the TypeID umbrella that we don't use or care about. Assuming the implementation is the same (and, as much as I can tell from eyeballing it, it is), this simplifies maintenance of the SDK and gets rid of some ongoing toil, so I'm all in favor of it. Good idea!

davidmytton (approved):
Great to drop an external dep!

PR #6104, opened 2026-06-26
feat: expose correlationId on protect() and guard()

qw-in:
Expose an optional, caller-supplied `correlationId` on the public entry points and map it to the protocol fields added in the proto regen:

- `protect()` (core `arcjet`): a new `correlationId` request field, mapped to `RequestDetails.correlation_id` (v1alpha1) and sent on both the decide and report paths.
- `guard()` (`@arcjet/guard`): a new `correlationId` option, mapped to `GuardRequest.correlation_id` (v2).

`correlationId` is treated as a known request field (so it is not funneled into the `extra` map) and is deliberately excluded from the fingerprint — and therefore the decision cache key — so two calls that differ only by correlation ID still share rate-limit state.

Server-side sanitization (max 256 bytes, printable ASCII; invalid values dropped, not truncated) remains authoritative; SDK-side validation can follow.

### DIFF
--- a/arcjet-bun/index.ts
+++ b/arcjet-bun/index.ts
@@ -179,7 +179,7 @@ export interface ArcjetBun<Props extends PlainObject> {
    */
   protect(
     request: Request,
-    ...properties: MaybeProperties<Props>
+    ...properties: MaybeProperties<Props & { correlationId?: string }>
   ): Promise<ArcjetDecision>;
 
   /**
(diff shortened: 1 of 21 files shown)


davidmytton (approved):

qw-in (commented):
Guard is looking correct but I don't think the Request version is. This is the expected api:

```
aj.protect(req, {
  correlationId: "foo"
})
```

PR #6108, opened 2026-06-29
test(arcjet): force test runner to exit; add CI test-job timeouts

qw-in:
## What

On Node 26 the `arcjet` package's `node --test` run **completed** (all tests passed and the coverage report printed) but the process never exited, hanging the `Run tests (Node: 26)` job until GitHub's 6h default. Because the run finishes before hanging, a stray ref'd handle is keeping the event loop alive *after* the runner is done — only surfaced in the CI sandbox (Node 22 limped through in ~9.5m; Node 26 hung and was cancelled).

## Changes

- Add `--test-force-exit` to the `arcjet` package's `test-coverage` / `test-api` scripts so the runner exits promptly once tests and reporters finish. This does not mask failures — pass/fail is determined before exit.
- Add `timeout-minutes: 15` to the reusable test jobs (`node-test`, `bun-test`, `deno-test`, `transport-runtime`) as a safety net so any future hang fails fast instead of running for hours.

## Verification

- Locally on Node 26.4.0 the `arcjet` coverage run passes and exits cleanly with the flag.
- `actionlint` and `zizmor` pass on the workflow.
- The Node 26 `Run tests` job on this PR is the real confirmation that the hang is gone.

### DIFF
--- a/.github/workflows/reusable-test.yml
+++ b/.github/workflows/reusable-test.yml
@@ -11,6 +11,9 @@ env:
 jobs:
   bun-test:
     name: Bun
+    # Tests normally finish in a few minutes; cap well above that so a hung
+    # process fails fast instead of running to the 6h GitHub default.
+    timeout-minutes: 15
     permissions:
       contents: read
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


davidmytton (approved):

qw-in:
This didn't solve it but at least limits the damage to 15m

PR #6109, opened 2026-06-30
feat(nosecone-next): add nonce API to get the CSP nonce

### LINKED ISSUE #3133: Add an API to Nosecone to get the nonce
This would help when developers need to provide the nonce to scripts which do not use something like [the Next.js Script component](https://nextjs.org/docs/app/api-reference/components/script).

For example, PostHog requires you to [pass the nonce into the init code](https://posthog.com/docs/advanced/content-security-policy#supporting-nonce-directives).

davidmytton:
## What

Adds a `nonce` export to `@nosecone/next` that returns the Content-Security-Policy nonce for the current request, so developers can pass it to third-party scripts that don't integrate with the Next.js `<Script>` component (e.g. PostHog, which requires the nonce in its `init` call).

```tsx
import { nonce } from "@nosecone/next";

export default async function Page() {
  const cspNonce = await nonce();
  return <script nonce={cspNonce}>{`/* ... */`}</script>;
}
```

## How

Nosecone's middleware sets a `Content-Security-Policy` response header containing a per-request nonce. Next.js copies that header onto the request headers and reads the nonce back out to apply it to its own rendered scripts. `nonce` reads the same request CSP header (via `next/headers`) and extracts the nonce using the same parsing Next.js uses — guaranteeing the returned value matches the nonce Next.js applies to its scripts. No middleware changes were needed.

Notable decisions:
- `next/headers` is imported dynamically inside `nonce()` so the server-only module isn't pulled into the Edge middleware bundle when `createMiddleware` is imported from the same entry point.
- Returns the bare nonce value (not `'nonce-...'`), matching what `nonce=""` attributes and libraries like PostHog expect, and mirroring Next.js's own extraction.
- The private nonce generator was renamed `nonce` → `createNonce` to free up the public name.

This is scoped to `@nosecone/next`. The base `nosecone` package generates no nonce by default and has no request context to read from; the SvelteKit adapter delegates nonce handling to SvelteKit's native CSP (`%sveltekit.nonce%`).

## Testing

- `npm run build`, `npm run lint`, and `npm test` pass for `nosecone-next`.
- Updated the public-API surface test to include `nonce`.

Closes #3133

### DIFF
--- a/nosecone-next/README.md
+++ b/nosecone-next/README.md
@@ -80,6 +80,33 @@ export default createMiddleware();
  }
 ```
 
+### Accessing the nonce
+
+When the `Content-Security-Policy` is enabled, Nosecone generates a unique
+nonce for every request. Next.js applies it to the scripts it renders
+automatically, but third-party scripts that don't use the Next.js `<Script>`
+component need the nonce passed in manually. For example,
+[PostHog](https://posthog.com/docs/libraries/next-js) requires the nonce in its
+`init` call.
+
+Use the `nonce` function to read the nonce for the current request from a Server
+Component, Route Handler, or anywhere else `next/headers` is available:
+
+```tsx
+import { nonce } from "@nosecone/next";
+
+export default async function Page() {
+  const cspNonce = await nonce();
+
+  return <script nonce={cspNonce}>{`/* ... */`}</script>;
+}
+```
+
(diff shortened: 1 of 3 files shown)


qw-in:
I clicked update branch to try and get the checks to pass (hope that's ok). Reviewing now

qw-in (approved):

PR #6111, opened 2026-06-30
fix(ci): force node --test to exit for all coverage packages (Node 26 hang)

qw-in:
## What

Follow-up to #6108 (which fixed only `arcjet` core + added job timeouts). After that merged, the Node 26 job still hung — turbo just stalled at the *next* coverage package.

## Root cause

`@arcjet/transport`'s `createHttp2Transport` does an optimistic pre-connect (`sessionManager.connect()`), leaving a ref'd connecting socket. Under CI's harden-runner `egress-policy: block`, packets to non-allowlisted hosts are **dropped**, so that socket stays in the `connecting` state indefinitely and keeps the event loop alive — the test run completes (all tests pass, coverage prints) but the process never exits. Node 22 idle-closed it after ~340s (its ~9.5m runs); Node 26 doesn't.

Reproduced locally by pointing the transport at a blackholed RFC5737 address (`192.0.2.1`): the run finishes, then the process hangs on a `ConnectWrap`/`TCPSocketWrap` until killed. Verified `--test-force-exit` makes the same scenario exit `0`.

The affected packages are exactly those whose coverage tests exercise the transport (`transport`, `guard`, and the `node`/`fastify`/`astro`/`nuxt`/`react-router` adapters), so the flag is applied uniformly to all `node --test --experimental-test-coverage` scripts.

## Change

Add `--test-force-exit` to all remaining coverage test scripts. It exits the runner once tests and reporters finish; it does not mask failures (pass/fail is decided before exit).

### DIFF
--- a/analyze-wasm/package.json
+++ b/analyze-wasm/package.json
@@ -56,7 +56,7 @@
     "build": "npm run build:jco && npm run build:rollup",
     "lint": "eslint .",
     "test-api": "node --test -- test/*.test.js",
-    "test-coverage": "node --experimental-test-coverage --test -- test/*.test.js",
+    "test-coverage": "node --experimental-test-coverage --test --test-force-exit -- test/*.test.js",
     "test": "npm run build && npm run lint && npm run test-coverage"
   },
   "dependencies": {},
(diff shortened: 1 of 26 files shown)


qw-in:
Superseded by the targeted fix in the new PR: skip the two Node 26-broken @arcjet/transport http2-proxy tests instead of force-exiting (force-exit dropped coverage below thresholds).

PR #6114, opened 2026-06-30
fix(arcjet-guard): accept object input on string-input rules

qw-in:
Previously it was impossible to set metadata on certain rules a "invocation" time:

```js
const pi = detectPromptInjection();
const bucket = tokenBucket();

await arcjet.guard({
  rules: [
    bucket({ requested: 12, metadata: { /* this */ } }),
    pi("string") // cannot set metadata
    // now we support this form too
    pi({ inputText: "string", metadata: {} })
  ]
})
```

Prompt injection, content moderation, and sensitive info rules took a bare string (and, for moderate content, a separate options bag for metadata) — diverging from the rate-limit rules, which take a single input object with metadata inside it. As a result per-request metadata was unavailable on prompt injection and sensitive info, and the SDK had two different metadata-passing shapes.

Unify on the object-input pattern. Each string-input rule now accepts `string | { inputText, metadata? }`, where a bare string is shorthand for `{ inputText }`, and the bound `RuleWithInput*.input` is that object. `ruleMetadataToProto` collapses to a single config+input merge (every rule's input is now an object), dropping the string/sibling special cases. Exports `DetectPromptInjectionInput` and
`LocalDetectSensitiveInfoInput`; reshapes `ExperimentalModerateContentInput` to carry `inputText`.

The bound `.input` for these three rules is now an object rather than a string. This also opens the path to richer inputs (e.g. a future conversation/messages variant) without another positional-arg migration.


qw-in:
> This PR changes the public rule input contract for prompt injection, content moderation, and sensitive-info rules, and adds tests for the new object-based input form

I consider the current form a bug. In practice this will not disrupt anyone given current usage and it is cleaner to fix it now. It will be mentioned in the changelog

estellajaysong (approved):

arcjet-rei (approved):
This looks good to me. I thought about a few alternative shapes (since we're in here redefining APIs already), but anything more general I could come up with that was an improvement also used generics, which falls afoul of the repo's style guidelines.

At some point, we may want to re-evaluate the elements of that style, but what we have works fine for now and revamping the API as a whole probably should wait until there's a deeper need / a more significant problem to solve.

qw-in:
> generics

The "request" api has a lot of fun with generics 😅. Titus and I spent many hours trying to untangle them. They also don't port very nicely to python, go, etc. TypeScript is special 😆

PR #6117, opened 2026-06-30
fix(security): update lockfiles to address npm security advisories

qw-in:
Resolve more security notices. Some unfixable:

Can't fix (no upstream release):

postcss — next@16.2.9 pins it at exactly 8.4.31; waiting on Next.js to loosen that pin
cookie — @sveltejs/kit@2.68.0 is both the latest and still in the vulnerable range; waiting on a kit release
Won't fix:

uuid in arcjet-guard/ — typeid-js pins uuid <11; will resolve after next release
uuid in examples/express-newman/ — would need newman major bump
React Router DoS in examples/remix-express/ — remix eol?
estree-util prototype pollution in examples/remix-express/ — remix eol?
Multer DoS ×2 in examples/nestjs/ — nest still pins multer below safe range
esbuild file read in react-router, astro, nuxt — blocked by parent package peer dep constraints; Windows-only dev server issue, low risk

### DIFF
--- a/arcjet-guard/package-lock.json
+++ b/arcjet-guard/package-lock.json
@@ -31,8 +31,8 @@
     },
     "node_modules/@arcjet/analyze": {
       "version": "1.5.0",
-      "resolved": "https://registry.npmjs.org/@arcjet/analyze/-/analyze-1.3.1.tgz",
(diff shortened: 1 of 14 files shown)


qw-in:
Upon examination Sonnet did a bad job. This doesn't fix anything real

PR #6093, opened 2026-06-20
chore: modernize build tooling

qw-in:
1. Unify arcjet-guard into shared workspace
2. Port from rollup to tsdown
3. Port from typescript 5 to 6 & tsgo
4. Port from eslint to oxlint & oxfmt

Advice to reviewers:

Focus on package.json, tsconfig, and tsdown config changes. The rest of the diff is mostly just noise. Specifically:

1. The GitHub actions changes
2. The wasm-specific handling (from the old rollup config). For instance in `analyze-wasm/wasm-plugins.js`
3. The `tsdown` config (all very similar for instance `analyze/tsdown.config.ts`)
4. The package.json changes (sorted by oxlint but specifically the `exports` & `main` fields - for instance in `analyze/package.json`)

The individual commits (before the last few) are fairly clean and may help!

Fixes #6122

PR #6130, opened 2026-07-06
ci(publish): preflight visibility and validation for the publish workflow

### LINKED ISSUE #6129: tighten down npm publish dist-tag selection
Today we (I) accidentally published `1.7.0-rc.1` tagged as `latest` instead of the intended `rc`. We resolved the issue quickly by re-publishing `1.6.1` as `1.7.0`.

However, before publishing another pre-release I'd like to find more guardrails. Some ideas:

1. Can we improve the workflow approval view so reviewers can easily see which selection was made? In this case `latest` vs `rc`? The base GitHub UI doesn't seem to surface that anywhere
2. Add a check to the publish workflow so when `latest` is selected we abort on any tags that don't match a reasonable regex (`^v\d+\.\d+\.\d+$`)

qw-in:
Publishing is gated by the `npm-publish` environment, but the approval UI doesn't show `workflow_dispatch` inputs — the reviewer approves blind, which is how `1.7.0-rc.1` briefly landed on the `latest` dist-tag. This restructures the publish workflow so every request is visible and validated before a human is asked to approve.

**Visibility**

- `run-name` carries the ref, dist-tag, and dry-run flag, so the run list and the approval notification say exactly what's being published: `Publish v1.7.0 to npm dist-tag latest`
- An ungated `preflight` job writes a step summary the reviewer can read before approving: dist-tag, ref/SHA, and every resolved package version (workspace + `@arcjet/guard`)

**Validation (all in preflight, before the approval gate)**

- The dist-tag must be a known channel — the input no longer defaults to `latest`, and since API dispatches bypass the choice list (a missing input arrives as `""`, and an empty `npm_config_tag` falls back to npm's default: `latest`), unknown/empty values fail closed
- The workflow must be dispatched on a release tag (`vX.Y.Z` or `vX.Y.Z-<pre-release>`), never a branch
- Every package version must equal the tag being published
- Versions aimed at `latest` must be stable releases (issue idea 2)

**Mechanics**

- The `publish` job depends on preflight and keeps the environment gate; `id-token: write` is scoped to it rather than the whole workflow
- A `dry-run` input exercises the entire flow — approval included — with `npm publish --dry-run`; both it and the dist-tag reach npm as `npm_config_*` environment config, scoped so guard's install/build/test always run for real (`npm ci` honors `dry-run`)
- A concurrency group queues concurrent publishes so dependency levels from two runs can't interleave
- The `tag` input is renamed to `dist-tag` to match the npm CLI

Every run-script was executed locally, including failure paths: the assertions were tested against valid/invalid tags, branches, empty and unknown dist-tags; the full publish step dry-ran all 35 packages tagged `rc` (on throwaway versions — npm dry-run still rejects already-published versions, a bonus guard); guard's step ran its real install/build/303 tests with only the publish dry. actionlint and zizmor (auditor persona) are clean.

Closes #6129

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -45,40 +45,27 @@ jobs:
         uses: actions/setup-node@6044e13b5dc448c55e2357c09f80417699197238 # v6.2.0
         with:
           node-version: 24
-      - name: Summarize the publish request
+      - name: Assert valid workflow inputs
         env:
           DIST_TAG: ${{ inputs['dist-tag'] }}
           DRY_RUN: ${{ inputs.dry-run }}
         run: |
-          # The jq and node programs are single-quoted on purpose.
-          # shellcheck disable=SC2016
-          versions="$(npm pkg get version --workspaces | jq -r 'to_entries[] | "\(.key)@\(.value)"')"
-          guard="@arcjet/guard@$(node --print 'require("./arcjet-guard/package.json").version')"
-          {
-            echo "## Publish request"
-            echo "- npm dist-tag: \`$DIST_TAG\`"
-            echo "- ref: \`$GITHUB_REF_NAME\` (\`$GITHUB_SHA\`)"
-            echo "- dry run: \`$DRY_RUN\`"
-            echo
-            echo "<details><summary>Versions</summary>"
-            echo
-            echo '```'
-            echo "$versions"
-            echo "$guard"
-            echo '```'
-            echo "</details>"
-          } >> "$GITHUB_STEP_SUMMARY"
-      - name: Assert a known dist-tag was provided
-        env:
-          DIST_TAG: ${{ inputs['dist-tag'] }}
-        run: |
           # The web form enforces the choice list, but API dispatches do not:
           # a missing input arrives as an empty string, and an empty
           # npm_config_tag would fall back to npm's default tag - latest.
+          # Validating both inputs here also means the later steps (including
+          # the approval-facing summary) only ever see known values.
           case "$DIST_TAG" in
             latest|rc) ;;
             *)
               echo "::error::Unknown npm dist-tag '$DIST_TAG'; expected 'latest' or 'rc'."
+              exit 1
+              ;;
+          esac
+          case "$DRY_RUN" in
+            true|false) ;;
+            *)
+              echo "::error::Invalid dry-run value; expected 'true' or 'false'."
               exit 1
               ;;
           esac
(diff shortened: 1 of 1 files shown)


.github/workflows/publish.yml
```diff
@@ -1,23 +1,145 @@
...
+# approval notification, since the approval UI does not display
+# workflow_dispatch inputs.
+run-name: "Publish ${{ github.ref_name }} to npm dist-tag ${{ inputs['dist-tag'] }}${{ inputs.dry-run && ' (dry run)' || '' }}"
 on:
   workflow_dispatch:
     inputs:
-      tag:
-        default: latest
-        description: The npm tag (distribution channel) to publish to.
+      dist-tag:
```
arcjet-rei, line 9: Thank you for renaming this. It is much less confusing this way.

.github/workflows/publish.yml
```diff
@@ -1,23 +1,145 @@
 name: Publish
+# Surface the inputs everywhere the run is shown, including the environment
+# approval notification, since the approval UI does not display
+# workflow_dispatch inputs.
+run-name: "Publish ${{ github.ref_name }} to npm dist-tag ${{ inputs['dist-tag'] }}${{ inputs.dry-run && ' (dry run)' || '' }}"
```
arcjet-rei, line 5: Clever!

.github/workflows/publish.yml
```diff
@@ -34,7 +156,10 @@ jobs:
         run: npm test
       - name: Publish packages
         env:
-          TAG: ${{ github.event.inputs.tag }}
+          # npm reads config from the environment; this step only runs
+          # publish commands, so these apply to all of them.
+          npm_config_tag: ${{ inputs['dist-tag'] }}
```
arcjet-rei, line 168: It's been so long since I last dealt with these variables that I forgot they existed. This is a good use for them.

.github/workflows/publish.yml
```diff
@@ -103,10 +228,13 @@ jobs:
       # TypeScript Go, rolldown) and it can rejoin the main workspaces.
       - name: Publish @arcjet/guard
         env:
-          TAG: ${{ github.event.inputs.tag }}
+          npm_config_tag: ${{ inputs['dist-tag'] }}
+          DRY_RUN: ${{ inputs.dry-run }}
```
arcjet-rei, line 239: Why not use `npm_config_dry_run` here as well?

arcjet-rei: Or, put another way, when do you decide to use `DRY_RUN` and when to use `npm_config_dry_run`?

arcjet-rei (approved):
A couple questions but nothing that blocks landing this. I guess the best test we can do is run it and pay close attention to the process.

.github/workflows/publish.yml
qw-in: This is to allow `npm ci` to run in live (non `dry_run`) mode. I've added comments and this can get dropped once #6093 lands (guard no longer needs a special case)

PR #6133, opened 2026-07-07
ci(publish): pass dist-tag via --tag and drop publishConfig.tag

arcjet-rei:
## What

- Pass the dist-tag with `--tag "$DIST_TAG"` on each `npm publish` in the publish workflow, instead of the `npm_config_tag` environment variable.
- Remove the redundant `publishConfig.tag: "latest"` from every published package (34 packages, incl. `@arcjet/guard`). `access: "public"` is kept.
- Add a post-publish `Verify dist-tags` step that asserts each package's requested dist-tag actually resolved to the published version, failing the job otherwise.

## Why

On the `1.8.0-rc.0` publish, the workflow requested the `rc` dist-tag via the `npm_config_tag` env var, and npm even logged `Publishing … with tag rc` — yet `1.8.0-rc.0` landed on `latest` and the `rc` tag was never advanced (it stayed at `1.7.0-rc.0`). Every package installing `@arcjet/*` at `latest` would have pulled a prerelease.

Root cause: every package hardcodes `publishConfig.tag: "latest"`. A **command-line `--tag`** outranks `publishConfig.tag`, but an **environment variable** (`npm_config_tag`) does not — so `publishConfig.tag: "latest"` silently won and routed every publish to `latest`. The previous workflow used `npm publish --tag "$TAG"` and worked correctly; the recent rewrite (#6130) switched to `npm_config_tag`, which exposed the trap.

`publishConfig.tag: "latest"` provided no benefit to remove:
- It only restated npm's default tag.
- It created the precedence trap above.
- It suppressed [npm 11's own guard](https://github.com/npm/cli/releases/tag/v11.0.0) that refuses to publish a prerelease under the default `latest` tag (an explicitly-set `latest` satisfies the guard).

## Fix from both ends

`--tag` on the command line makes the requested tag authoritative regardless of manifest config, and dropping `publishConfig.tag` removes the trap entirely. Belt and suspenders.

Note: `npm_config_dry_run` is intentionally left as an env var — it's a boolean with no `publishConfig` equivalent, so it has no analogous precedence issue.

## Safety net: verify dist-tags after publishing

The failure above shipped **green** — `npm publish` reports the requested tag but can silently write a different one while exiting 0. The new `Verify dist-tags` step closes that gap:

- Derives the published package list from `npm pkg get name --workspaces` (the same source preflight uses — no hardcoded list to drift), then reads `dist-tags.<requested-tag>` from the registry for each and asserts it equals the just-published version.
- Fails the job with `::error::` annotations naming any package whose tag didn't resolve.
- Skipped on dry runs (`if: ${{ !inputs.dry-run }}`).
- Retries a few times with backoff per package to absorb brief registry propagation lag (avoids flaky false failures).

Validated against live registry data: it passes on the current correct state (`latest=1.8.0`) and **reproduces a failure on the incident signature** (`rc` stuck at `1.7.0-rc.0` while expecting the published version) — i.e. had it existed, the bad publish would have failed loudly instead of shipping.

## Verification

- All 34 `package.json` files still parse and retain `access: "public"`; only the `tag` key was removed (`+1/-2` per file).
- Workflow YAML validated; 5 publish levels all carry `--tag "$DIST_TAG"`; no live `npm_config_tag` remains (only an explanatory comment). The `Verify dist-tags` run script is shellcheck-clean.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -51,10 +51,10 @@ jobs:
           DRY_RUN: ${{ inputs.dry-run }}
         run: |
           # The web form enforces the choice list, but API dispatches do not:
-          # a missing input arrives as an empty string, and an empty
-          # npm_config_tag would fall back to npm's default tag - latest.
-          # Validating both inputs here also means the later steps (including
-          # the approval-facing summary) only ever see known values.
+          # a missing input arrives as an empty string, and an empty dist-tag
+          # would publish `--tag ""`. Validating both inputs here also means
+          # the later steps (including the approval-facing summary) only ever
+          # see known values.
           case "$DIST_TAG" in
             latest|rc) ;;
             *)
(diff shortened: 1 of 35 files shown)


qw-in (approved):
Thank you!

qw-in:
Any idea why socket is picking up arcjet@node 1.1.0?

qw-in (approved):

arcjet-rei:
@qw-in It picked it up from `examples/nodejs-rate-limit/package-lock.json`, which embeds `../../arcjet-node` (so it's a stale version reference). Not a big deal, but I guess we should regenerate the lockfile at some point.

Issue #1478, opened 2024-09-02
Reduce HTTP2 session idle timeout


davidmytton:
We introduced the HTTP2 session manager in https://github.com/arcjet/arcjet-js/pull/1201 with an idle timeout of 600 seconds. 

We've moved our edge infra to AWS Global Accelerator which has a [maximum idle timeout of 340 seconds](https://docs.aws.amazon.com/global-accelerator/latest/dg/introduction-how-it-works.html#about-idle-timeout), so we need to adjust our default.

PR #6137, opened 2026-07-14
fix(arcjet-guard): detect and recycle silently dropped HTTP/2 connections

arcjet-rei:
## Problem

The guard direct HTTP/2 transport holds one long-lived session with **no keep-alive probing**. If an intermediary (NAT gateway, L4 load balancer, conntrack) silently drops the connection state during an idle period, the SDK holds a session that looks open but black-holes every write: **every guard call times out and fails open, indefinitely, until the process restarts**. This is the leading hypothesis for a production incident (traffic to the Arcjet API stopping entirely after a burst of timeouts, while the app keeps running).

PINGs were originally disabled citing AWS Global Accelerator's idle-timeout docs. That was a misreading: GA ignores **dataless TCP keepalive packets**, but HTTP/2 PING frames are stream data, which *does* reset its idle timer ([ref](https://docs.aws.amazon.com/global-accelerator/latest/dg/introduction-how-it-works.html#about-idle-timeout)).

## Changes

Two layers of defense in `createHttp2Transport`:

1. **PING keep-alive** (`pingIntervalMs: 55s`, `pingTimeoutMs: 5s`, `pingIdleConnection: true`):
   - verifies a connection with a PING before reusing it after inactivity, transparently re-dialing if it's dead;
   - tears down a dead connection within 5s once a PING goes unanswered (the 15s default is slow next to 1–2s guard timeouts);
   - keeps NAT/conntrack entries alive (AWS NAT gateways expire idle flows at 350s, GA at 340s; 55s clears both plus more aggressive middleboxes).
   - connect-es `unref()`s idle connections and all ping timers, so this does not keep a quiescent process from exiting (verified in the connect-node source).

2. **Connection recycling backstop** (`transport-recycle.ts`): after 3 consecutive `DeadlineExceeded` failures with no success in between, abort the session and optimistically re-dial. Only deadline failures count — caller cancellation (`Canceled`) and connection-level errors (which already put the session manager in its error state) don't. A recycle logs a warning when `ARCJET_LOG_LEVEL` requests it, following the `detect-proxy.ts` edge-safe logging pattern.

`createHttp2Transport` now returns `{ transport, sessionManager }` (internal API) so callers and tests can tear connections down deterministically.

## Not in scope (follow-ups from the same investigation)

- Defaulting the proxy path to HTTP/2-over-CONNECT (`proxyHttpVersion: "2"`)
- Runtime feature-detection of the Node agent `proxyEnv` option (silently ignored on Node < 22.x backport)
- Mirroring these options in `@arcjet/transport` for the main SDK

## Testing

- 7 new unit tests for the recycling wrapper via the in-memory Connect router transport (threshold, reset-on-success, non-deadline errors neither count nor reset, caller cancellation excluded, run restarts after recycle, pass-through, log gating).
- New runtime test driving `createHttp2Transport` itself against a real HTTP/2 server (previous runtime tests hand-assembled their transports and never exercised the factory).
- Full guard suite: 310 unit + 28 Node runtime + 26 Bun runtime tests pass; build, oxlint, tsgo, oxfmt clean.

### DIFF
--- a/arcjet-guard/src/transport-bun.ts
+++ b/arcjet-guard/src/transport-bun.ts
@@ -31,7 +31,7 @@ export function createTransport(baseUrl: string): Transport {
 
   // No proxy: connect directly over HTTP/2.
   if (proxyUrl === undefined) {
-    return createHttp2Transport(baseUrl);
+    return createHttp2Transport(baseUrl).transport;
   }
 
   // Proxy: Bun's native `fetch` honors the proxy environment variables. The
--- a/arcjet-guard/src/transport-http2.ts
+++ b/arcjet-guard/src/transport-http2.ts
@@ -14,20 +14,56 @@
 import type { Transport } from "@connectrpc/connect";
 import { createConnectTransport, Http2SessionManager } from "@connectrpc/connect-node";
 
+import { withConnectionRecycling } from "./transport-recycle.ts";
+
+/**
+ * A direct HTTP/2 transport plus the session manager that owns its connection.
+ *
+ * The session manager is exposed so callers (and tests) can tear the
+ * connection down deterministically.
+ */
+export interface Http2TransportHandle {
+  transport: Transport;
+  sessionManager: Http2SessionManager;
+}
+
 /**
  * Create a direct HTTP/2 Connect transport, optimistically pre-connecting.
  *
  * The session is pre-connected so the first `.guard()` call doesn't pay the
- * full TCP + TLS setup cost.
+ * full TCP + TLS setup cost. PING keep-alive and deadline-based connection
+ * recycling detect a silently dropped connection (an intermediary expiring an
+ * idle flow without notifying either end) and replace it, instead of letting a
+ * dead session fail every call until the process restarts.
  *
  * @param baseUrl Base URL for the Arcjet API.
- * @returns A Connect transport that talks HTTP/2 directly to `baseUrl`.
+ * @returns The transport and its session manager.
  */
-export function createHttp2Transport(baseUrl: string): Transport {
+export function createHttp2Transport(baseUrl: string): Http2TransportHandle {
   const sessionManager = new Http2SessionManager(baseUrl, {
-    // AWS Global Accelerator doesn't support PING so we use a very high idle
-    // timeout. Ref:
-    // https://docs.aws.amazon.com/global-accelerator/latest/dg/introduction-how-it-works.html#about-idle-timeout
+    // Detect and survive silently dropped connections:
+    //
+    // - `pingIntervalMs` sends PING frames on connections with in-flight
+    //   streams and, crucially, verifies a connection with a PING before
+    //   reusing it after `pingIntervalMs` of inactivity — transparently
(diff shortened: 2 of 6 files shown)


qw-in (approved):
I think this makes sense. Is it something that we should port to `@arcjet/transport` as well or is there some difference in implementation there?

I'm a little surprised we're only encountering this now given how long a very similar pattern has been in use. Do you know if it's changes in Node, AWS, or some dependency?

arcjet-rei:
@qw-in 
> I'm a little surprised we're only encountering this now given how long a very similar pattern has been in use. Do you know if it's changes in Node, AWS, or some dependency?

Well, from the PR description:

> PINGs were originally disabled citing AWS Global Accelerator's idle-timeout docs. That was a misreading: GA ignores **dataless TCP keepalive packets**, but HTTP/2 PING frames are stream data, which *does* reset its idle timer ([ref](https://docs.aws.amazon.com/global-accelerator/latest/dg/introduction-how-it-works.html#about-idle-timeout)).

So the old implementation was intended to keep us out of trouble with AWS GA, and either things have changed, or we were always misreading the GA docs, or Claude is being confidently wrong here.

> Is it something that we should port to @arcjet/transport as well or is there some difference in implementation there?

We can port it over (and I'll put up a second PR with a similar change to @arcjet/transport), but the main SDK-Decide API connection works differently enough from guards and prompt injection detection that it's not as useful or necessary there.

davidmytton:
#1201 introduced the custom HTTP/2 session manager in July 2024 with a 600-second idle timeout. Its comment said AWS ALB did not support PING for resetting idle connections, which is correct [per the docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/application/edit-load-balancer-attributes.html):

> Application Load Balancers do not support HTTP/2 PING frames. These do not reset the connection idle timeout.

We did the infra config in https://github.com/arcjet/arcjet/pull/2766

When we moved edge traffic to Global Accelerator, #1478 and #1479 reduced the client timeout from 600 to GA’s 340-second maximum. In that change, the source comment changed from effectively “ALB doesn’t support PING” to “Global Accelerator doesn’t support PING.” That was a misreading of the docs:

> To prevent connection timeout, Global Accelerator requires that you send a packet with a minimum of one byte of data, in the ingress or egress direction, within the TCP connection timeout window. You cannot use TCP keep-alive packets to maintain an open connection.

The change in this PR only applies to the client to GA leg, not the GA to ALB leg or the ALB to our backend. We won't have HTTP2 all the way through to the origin until https://github.com/arcjet/arcjet/pull/8496 lands.

One concern with the recycling logic: the deadline counter is shared across all requests and isn't tied to a particular session. If many requests time out concurrently, the first three can trigger a reconnect, then later completions from that old session can cross the threshold again and abort the newly established session. Could we make recycling single-flight and associate failures with a session generation, so each suspect session can trigger at most one recycle?

arcjet-rei:
> Could we make recycling single-flight and associate failures with a session generation, so each suspect session can trigger at most one recycle?

This makes sense. I'll put together a follow-up PR with something like this in it.

PR #6148, opened 2026-07-16
chore: Guard experimental_capture

qw-in:


davidmytton (approved):

qw-in:
This will be re-opened or re-created based on the amended adr.

PR #6163, opened 2026-07-23
chore(proto): sync generated bindings with arcjet monorepo main

qw-in:
Regenerates the vendored guard v2 bindings from the arcjet monorepo `main` branch. The only change is the Capture RPC message types (`CaptureEvent`, `CaptureRequest`, `CaptureResponse`) introduced in arcjet/arcjet#8528.

Generated code only; no behavioral change. Establishes a current proto baseline ahead of the metadata work.


qw-in:
The goal here is just to keep the proto in sync with the server.

davidmytton (approved):

PR #6164, opened 2026-07-23
feat(guard): add provider-namespaced AI SDK helpers as @arcjet/guard subpaths

arcjet-rei:
## Summary

Adds provider-namespaced AI SDK helpers to `@arcjet/guard` as a **subpath export**. No new package is published — the helpers ship inside the existing `@arcjet/guard` release.

This replaces the original approach on this branch, which added a separate `@arcjet/ai` package. That package has been removed; all of its behaviour now lives under `@arcjet/guard/vercel-ai/v7`.

## Two import paths

```ts
// Core guard — unchanged, no AI SDK involved
import { launchArcjet, slidingWindow, tokenBucket } from "@arcjet/guard";

// Every agent helper, SDK-coupled or not — one path
import {
  guardTool, aiToolsContext,
  createAgentContext, guardAction, captureAction, securityMetadata,
} from "@arcjet/guard/vercel-ai/v7";
```

An earlier revision of this PR also published the SDK-agnostic helpers as `@arcjet/guard/agents`. That path is gone, per @qw-in ([here](https://github.com/arcjet/arcjet-js/pull/6164#discussion_r3675568860) and [on the ADR](https://github.com/arcjet/arcjet/pull/8658#discussion_r3675738196)): *"I'd prefer to keep it all in vercel-ai until the abstraction proves itself over multiple different providers… I'm not sure users want/need to think about these layers other than 'I'm using the Vercel AI SDK'."*

`src/agents/` still exists and is still AI-SDK-free — a test walks its transitive import graph to keep it that way — but it has no export map entry and reaches users only re-exported from the vendor namespace. **The trade-off:** a non-AI codebase wanting only `guardAction` now installs the `ai` peer. That is deliberate. A public path is a compatibility commitment, and this one would be justified by a single integration; adding the entry later is additive, withdrawing it is breaking. The expected destination is the root export once a second SDK has proven the shape.

## The convention

Subpaths are `@arcjet/guard/<vendor-sdk>/v<major>` — vendor-prefixed and flat. `vercel-eve/v1` is the next planned namespace, so the pattern has a second instance by design rather than by extrapolation from one.

**Versions are explicit and there is no unversioned alias.** `@arcjet/guard/vercel-ai` does not resolve, deliberately: against a fast-moving SDK surface an alias would silently change meaning the moment a new major is supported. Verified against a packed tarball installed into a scratch project:

```
RESOLVES                               @arcjet/guard
RESOLVES                               @arcjet/guard/vercel-ai/v7
BLOCKED ERR_PACKAGE_PATH_NOT_EXPORTED  @arcjet/guard/agents
BLOCKED ERR_PACKAGE_PATH_NOT_EXPORTED  @arcjet/guard/vercel-ai
BLOCKED ERR_PACKAGE_PATH_NOT_EXPORTED  @arcjet/guard/agents/context

v7 exports: ArcjetDeniedError, ArcjetGuardUnavailableError, aiToolsContext,
            captureAction, createAgentContext, guardAction, guardTool,
            securityMetadata
```

Deep internal paths are encapsulated for free by the same mechanism.

## Existing guard users are unaffected

- **The root export surface is unchanged.** Nothing was added to or removed from `@arcjet/guard` itself.
- **`ai` and `@ai-sdk/provider-utils` are optional peer dependencies.** A project that installs `@arcjet/guard` for core guards alone gets no peer warning and no error.
- Note for pnpm users: pnpm does not reliably honour `peerDependenciesMeta.*.optional` (pnpm#5152, #8142), particularly under `--strict-peer-dependencies`. Documented in the README.

## Renames

Carried out across source, tests, docs, the packaged skill, and the example:

| Before | After |
|---|---|
| `createAiContext` | `createAgentContext` |
| `ArcjetAiContext` | `ArcjetAgentContext` |
| `protectTool` / `ProtectToolPolicy` | `guardTool` / `GuardToolPolicy` |
| `protectAction` / `ProtectActionPolicy` | `guardAction` / `GuardActionPolicy` |

The verb rename follows @qw-in's point that these run guard rules, so `guard*` is the honest name.

## `onGuardError` defaults to `"deny"`

Following @davidmytton's point that these wrap live enforcement events rather than page-view bot checks: if a policy cannot be evaluated, the call is **blocked** by default. `onGuardError: "allow"` is the opt-out for call sites where availability matters more than enforcement.

"Unavailable" means both the `guard()` call throwing *and* a decision whose `hasFailedOpen()` is `true` — the second is the common one. The two surfaces differ deliberately:

- `guardTool` returns `reason: "ERROR"` with `retryable: true` and `retryAfterSeconds: 5`, because a model only ever observes tool results and a second result type is harder to prompt against.
- `guardAction` throws `ArcjetGuardUnavailableError`, which is **not** `ArcjetDeniedError`, carrying `cause` (the call threw) or `decision` (a decision failed open). "A rule denied you" and "the policy could not be evaluated" are operationally different, and usually only the second is alert-worthy.
- The capture event records `outcome: "unavailable"`, not `"denied"`, so an unavailable guard can be queried separately from a rule's denial.

The layering reads as a contradiction otherwise, so stating it plainly: the `@arcjet/guard` client still fails open by construction and *reports* it via `hasFailedOpen()`; these helpers are what *decide* to block on it.

Also raises the client's default `guard()` deadline from 1s to 2s. Content moderation and prompt injection take materially longer than a rate-limit check, and because a deadline fails open, the 1s default was dropping those rules rather than evaluating them — callers got less protection than they configured, silently. That was true before anything here failed closed; it also reduces spurious blocks now that a deadline denies. This reaches every `guard()` caller, not only these helpers.

## Also in this PR

- **Capture now uses the client's `capture()`.** The agents layer was written against `@arcjet/ai`, a separate package that could be paired with an older `@arcjet/guard`, so it feature-detected an optional `experimental_capture()`. Rebasing onto main brought in the real capture layer, whose method is `capture()` — and because the old method was optional, this compiled while silently no-opping every capture. Fixed, with a type-level assertion that `ArcjetGuard` satisfies the helpers' client interface so a rename cannot go quiet again.
- **Example** (`examples/nextjs-ai-agent`) demonstrates rules derived from tool input (`rules: ({ orderNumber }) => [...]`) and the `onGuardError: "allow"` opt-out on a read-only lookup. The explicit-call alternative — `guardAction` inside a tool's `execute` instead of wrapping with `guardTool` — is documented in the README, with a pointer to it from the example.
- **Agent skill** moved to `arcjet-guard/skills/integrate-arcjet-guard-agents/`; ships in the npm package via `files: ["dist/", "skills/"]`.
- **README** documents the two paths, the convention, the no-alias rule, optional peers, `onGuardError`, what `correlationId` is for, and a recipe for adding the next namespace.
- **Release/publish wiring** for `@arcjet/ai` fully unwound — `git diff main -- .github/` now touches only `reusable-examples.yml`, which keeps the example's CI entry.

## Verification

| Check | Result |
|---|---|
| `build`, both typechecks, `lint` | green |
| `test-unit` | **462 pass / 0 fail** |
| Runtime suites: node / fetch / bun / cloudflare | 28 / 26 / 26 / 2 — all green, run locally |
| Runtime suite: deno | not installed locally — covered in CI by `Runtime (deno lts)` and `Runtime (deno latest)` |
| Example typecheck | green against the single import path |
| Live subpath resolution | both public paths resolve; `/agents`, `/vercel-ai`, and deep internals correctly blocked |
| Doc examples | 42 TS blocks extracted from the README, `SKILL.md`, and every JSDoc `@example`, compiled against the packed tarball's typings: **zero structural errors**. Residue is undeclared narrative placeholders (`userId`, `ctx`, `languageModel`) in continuation snippets |
| Doc imports | every named import across those 42 blocks verified to be a real export of the path it names — 8 from `/vercel-ai/v7`, 7 from the root |

The Turbopack subpath-exports bug (vercel/next.js#88540) affects transitive dependencies; the example depends on `@arcjet/guard` directly and builds cleanly, so no workaround was needed.

## Release impact

This lands as a **minor** bump for `@arcjet/guard` — nothing is removed or renamed in the published surface. `@arcjet/guard/agents` and `createAiContext` are named in the commit history as breaking, but neither was ever published, so no installed consumer can observe either.

Two changes **do** reach existing `guard()` callers who never touch these helpers, and are worth reading before upgrading:

- **The default request deadline moves from 1s to 2s.** Content moderation and prompt injection take materially longer than a rate-limit check, and because a deadline fails open, the 1s default was dropping those rules rather than evaluating them — callers got less protection than they configured, silently. A caller who does not set `timeoutSeconds` now waits up to a second longer before failing open.
- **`guard({ rules: [] })` now reaches the server** instead of returning a locally synthesized fail-open decision. It costs a round trip and a billed unit where it used to cost nothing, and it returns an ALLOW with a real decision id and an `AJ1002` warning rather than one whose `hasFailedOpen()` is true. `GuardOptions.rules` documented "at least one is required", so only callers already outside that contract are affected. Use `capture()` if you want a record without a decision.

## Open

- @davidmytton's [larger policy contract](https://github.com/arcjet/arcjet-js/pull/6164#discussion_r3692653682) — `action` / `actor` / `inputs`, with the policy defined remotely — is not implemented here. His question to @qw-in about that shape is still open, and it is a product decision rather than a refactor.
- `pilot-framework-helper.AC5.2` (example app live run against a dev site) and `AC6.1` (fresh-agent skill run) are manual and still outstanding; both procedures are recorded on ENG-987.


examples/nextjs-ai-agent/workflows/support-agent.ts
```diff
@@ -0,0 +1,113 @@
...
+      action: "order.looked-up",
+      rules: () => [lookupLimit({ key: "demo-user", requested: 1 })],
+      metadata: ({ orderNumber }) =>
+        securityMetadata({ resource: `order:${orderNumber}` }),
+    },
+  ),
+};
+
+export async function supportAgentWorkflow(input: SupportAgentInput) {
+  "use workflow";
```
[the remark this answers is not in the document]
arcjet-rei: These `"use workflow"` / `"use step"` directives are Vercel Workflow DevKit function-level directives (analogous to `"use server"`) — they are consumed by the workflow compiler wired in via `withWorkflow()` in this example's `next.config.ts`, and removing them would break the workflow. CodeQL doesn't recognize the directive names; this is intentional, working code (see https://vercel.com/docs/workflow). No change made.

examples/nextjs-ai-agent/workflows/support-agent.ts
```diff
@@ -0,0 +1,113 @@
...
+  const answer = await stepRunAgent(input);
+  await stepUpdateTicket(input, answer);
+  return { answer };
+}
+
+async function stepRunAgent(input: SupportAgentInput) {
+  "use step";
```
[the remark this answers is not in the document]
arcjet-rei: These `"use workflow"` / `"use step"` directives are Vercel Workflow DevKit function-level directives (analogous to `"use server"`) — they are consumed by the workflow compiler wired in via `withWorkflow()` in this example's `next.config.ts`, and removing them would break the workflow. CodeQL doesn't recognize the directive names; this is intentional, working code (see https://vercel.com/docs/workflow). No change made.

examples/nextjs-ai-agent/workflows/support-agent.ts
```diff
@@ -0,0 +1,113 @@
...
+    prompt: input.question,
+    tools,
+    toolsContext: aiToolsContext(input.arcjet, tools),
+    stopWhen: stepCountIs(3),
+  });
+  return result.text;
+}
+
+async function stepUpdateTicket(input: SupportAgentInput, answer: string) {
+  "use step";
```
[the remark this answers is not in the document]
arcjet-rei: These `"use workflow"` / `"use step"` directives are Vercel Workflow DevKit function-level directives (analogous to `"use server"`) — they are consumed by the workflow compiler wired in via `withWorkflow()` in this example's `next.config.ts`, and removing them would break the workflow. CodeQL doesn't recognize the directive names; this is intentional, working code (see https://vercel.com/docs/workflow). No change made.

arcjet-ai/test/generate-text.test.ts
```diff
@@ -0,0 +1,289 @@
...
+  assert.ok(toolResultPart, "first step should have a tool-result part");
+
+  const output = (toolResultPart as Record<string, unknown>).output as ArcjetDenialResult;
+  assert.strictEqual(output.arcjetDenied, true, "output should be an ArcjetDenialResult");
+  assert.equal(output.reason, "RATE_LIMIT", "denial reason should be RATE_LIMIT");
+
+  // The model should have received the denial result in its second call
+  // Check that the model was called twice (first with initial prompt, second with tool result)
+  assert.ok(
+    typeof mockModel === "object" && mockModel !== null && "doGenerateCalls" in mockModel,
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in 05fc6b83 — removed the impossible null comparison and the casts; the test now reads the typed `mockModel.doGenerateCalls` property directly.

arcjet-ai/test/protect-action.test.ts
```diff
@@ -0,0 +1,282 @@
...
+
+  assert.equal(fnCallCount, 0, "fn should never be called");
+  assert.equal(guardCalls.length, 1, "guard should be called once");
+});
+
+test("AC3.3: success path → one capture with metadata outcome: success and decisionId set", async () => {
+  const { client, captureCalls } = stubClient(decisionAllow());
+  const sentinel = { result: "success" };
+
+  const result = await protectAction(
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in 05fc6b83 — dropped the unused `result` binding; the call is now a bare `await protectAction(...)`.

arcjet-ai/test/generate-text.test.ts
```diff
@@ -0,0 +1,289 @@
...
+      "warning should mention missing context",
+    );
+  } finally {
+    console.warn = originalWarn;
+    delete process.env.ARCJET_LOG_LEVEL;
+  }
+});
+
+test("AC2.9: DENY decision → generateText completes, loop continues with denial result in first step", async () => {
+  const { client, guardCalls } = stubClient(
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in 05fc6b83 — removed the unused `guardCalls` binding from the AC2.9 destructuring.

arcjet-ai/test/protect-tool.test.ts
```diff
@@ -0,0 +1,618 @@
...
+        return sentinel;
+      },
+    }),
+    executeCalls,
+    sentinel,
+  };
+}
+
+test("AC2.1: ALLOW decision → original execute called, result returned unchanged", async () => {
+  const { client, guardCalls, captureCalls } = stubClient(decisionAllow());
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in 05fc6b83 — removed the unused `captureCalls` binding (kept `guardCalls`, which the test asserts on).

arcjet-ai/test/protect-tool.test.ts
```diff
@@ -0,0 +1,618 @@
...
+  assert.equal(captureCalls.length, 1, "capture should be called once on denial");
+  const captureCall = captureCalls[0] as Record<string, unknown>;
+  assert.equal(captureCall.decisionId, "gdec_deny1");
+  const metadata = captureCall.metadata as Record<string, string>;
+  assert.equal(metadata.outcome, "denied");
+});
+
+test("AC2.6: guard throws → execute runs, warning emitted", async () => {
+  const guardError = new Error("guard API error");
+  const { client, guardCalls, captureCalls } = stubClient(guardError);
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in 05fc6b83 — removed the unused `guardCalls` binding from the AC2.6 destructuring.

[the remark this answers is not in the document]
arcjet-rei: Addressed in 05fc6b83 — removed the unused `captureCalls` binding from the AC2.6 destructuring.

arcjet-ai/test/protect-tool.test.ts
```diff
@@ -0,0 +1,618 @@
...
+      "warning should mention guard error",
+    );
+  } finally {
+    console.warn = originalWarn;
+    delete process.env.ARCJET_LOG_LEVEL;
+  }
+});
+
+test("AC2.6: guard resolves fail-open ALLOW → execute runs, fail-open warning", async () => {
+  const { client, captureCalls } = stubClient(decisionFailOpenAllow());
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in 05fc6b83 — removed the unused `captureCalls` binding from the fail-open ALLOW test.

arcjet-ai/test/protect-tool.test.ts
```diff
@@ -0,0 +1,618 @@
...
+    assert.strictEqual(e, testError, "same error should propagate");
+  }
+
+  assert.equal(captureCalls.length, 1, "capture should fire once with error");
+  const captureCall = captureCalls[0] as Record<string, unknown>;
+  const metadata = captureCall.metadata as Record<string, string>;
+  assert.equal(metadata.outcome, "error");
+});
+
+test("Untested branch: non-RATE_LIMIT DENY (PROMPT_INJECTION) → retryable=false, no retryAfterSeconds, correct message", async () => {
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in 05fc6b83 — renamed both titles (lines 331 and 360) to describe the behavior under test instead of the coverage gap they originally closed.

arcjet-ai/test/protect-tool.test.ts
```diff
@@ -0,0 +1,618 @@
...
+    messages: [],
+  } as never);
+
+  assert.equal(guardCalls.length, 0, "guard should not be called with empty rules");
+  assert.equal(executeCalls.length, 1);
+  assert.strictEqual(result, sentinel);
+});
+
+test("Missing capture support: client without experimental_capture → warning only", async () => {
+  const { client, guardCalls } = stubClient(decisionAllow());
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in db1b349e7 — the test now destructures only `client`; the unused `guardCalls` binding is gone.

arcjet-ai/test/protect-tool.test.ts
```diff
@@ -0,0 +1,618 @@
...
+      warnCalls.some(
+        (call) =>
+          JSON.stringify(call).includes("guard check") && JSON.stringify(call).includes("errored"),
+      ),
+      "warning should mention guard error",
+    );
+  } finally {
+    console.warn = originalWarn;
+    delete process.env.ARCJET_LOG_LEVEL;
+  }
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in db1b349e7 — added a shared `setLogLevel()` test helper that records the previous `ARCJET_LOG_LEVEL` and restores it (or its absence) in `finally`. Applied to all four sites in this file, and also to the same pattern in protect-action, generate-text, and warn-missing-context tests.

arcjet-ai/test/protect-action.test.ts
```diff
@@ -0,0 +1,282 @@
...
+  const sentinel = { result: "success" };
+  let fnCallCount = 0;
+
+  const originalWarn = console.warn;
+  const warnCalls: unknown[] = [];
+  console.warn = (...args: unknown[]) => {
+    warnCalls.push(args);
+  };
+
+  process.env.ARCJET_LOG_LEVEL = "warn";
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in db1b349e7 — both sites now use the shared `setLogLevel()` helper, which saves the previous `ARCJET_LOG_LEVEL` and restores it in `finally` instead of unconditionally deleting it.

arcjet-ai/src/guarded.ts
```diff
@@ -0,0 +1,89 @@
...
+  if (rules !== undefined && rules.length > 0) {
+    let decision: Decision | undefined;
+    try {
+      decision = await client.guard({ label: action, rules, ...correlation, metadata });
+    } catch (error) {
+      // Defense in depth: the guard client itself converts transport failures
+      // into ALLOW decisions with hasFailedOpen() === true, so reaching here
+      // means something unexpected broke. Fail open.
+      if (shouldWarn()) {
+        console.warn(`@arcjet/ai: guard check for "${action}" errored; failing open:`, error);
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in db1b349e7 — the warning now uses a constant format string ('@arcjet/ai: guard check for "%s" errored; failing open:') with `action` passed as a format argument, so format specifiers in the label can no longer forge the log message.

arcjet-ai/package.json
```diff
@@ -0,0 +1,73 @@
...
+    "@arcjet/guard": "1.9.1",
+    "@types/node": "22.20.1",
+    "ai": "7.0.36",
+    "tsdown": "0.22.7",
+    "typescript": "7.0.2",
+    "zod": "4.4.3"
+  },
+  "peerDependencies": {
+    "@ai-sdk/provider-utils": ">=5 <6",
+    "@arcjet/guard": "1.9.1",
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in db1b349e7 — widened the peer dependency to `>=1.9.1 <2`, matching the range style of the other peers in this file, so consumers can pair @arcjet/ai with any newer 1.x @arcjet/guard.

examples/nextjs-ai-agent/.gitignore
```diff
@@ -0,0 +1 @@
+/.swc
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in db1b349e7 — this example now uses the standard Next.js example .gitignore shared by the other examples (`.vercel`, `.next/`, `node_modules`, env files, etc.), with `/.swc` kept.

arcjet-ai/src/context.ts
```diff
@@ -0,0 +1,138 @@
...
+export function createAiContext(init?: {
+  correlationId?: string;
+  metadata?: Record<string, string>;
+}): ArcjetAiContext {
+  let correlationId: string;
+
+  // Check if correlationId was explicitly provided
+  if (init?.correlationId !== undefined) {
+    correlationId = init.correlationId;
+    // Validate caller-supplied IDs (generated ULIDs are correct by construction)
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in db1b349e7 — `createAiContext()` now rejects a non-string `correlationId` with a `typeof` check before the regex (the error names the offending type), and a test covers the number case.

examples/nextjs-ai-agent/app/page.tsx
```diff
@@ -0,0 +1,113 @@
...
+        Ask a question about an order, e.g. "What's the status of order 42?"
+      </p>
+
+      <form onSubmit={handleSubmit}>
+        <input
+          type="text"
+          value={question}
+          onChange={(e) => setQuestion(e.target.value)}
+          placeholder="Ask about an order, e.g. what's the status of order 42?"
+          style={{
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in db1b349e7 — added `aria-label="Question for the support agent"` to the input.

arcjet-ai/src/protect-tool.ts
```diff
@@ -0,0 +1,220 @@
...
+          value !== null &&
+          typeof (value as ArcjetAiContext).correlationId === "string")
+      ) {
+        return { success: true, value: value as ArcjetAiContext | undefined };
+      }
+      return {
+        success: false,
+        error: new Error("@arcjet/ai: toolsContext entry is not an ArcjetAiContext"),
+      };
+    },
```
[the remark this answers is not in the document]
arcjet-rei: Addressed in db1b349e7 — `validate()` now also checks that `metadata`, when present, is a non-null object whose values are all strings, so a malformed `toolsContext` entry fails validation instead of being forwarded to guard/capture. Added tests covering the accepted and rejected shapes.

examples/nextjs-ai-agent/workflows/support-agent.ts
```diff
@@ -0,0 +1,112 @@
...
+    prompt: input.question,
+    tools,
+    toolsContext: aiToolsContext(input.ctx, tools),
+    stopWhen: stepCountIs(3),
+  });
+  return result.text;
+}
+
+async function stepUpdateTicket(input: SupportAgentInput, answer: string) {
+  "use step";
```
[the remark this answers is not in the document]
arcjet-rei: These `"use step"` directives are Vercel Workflow DevKit function-level directives (analogous to `"use server"`) — they are consumed by the workflow compiler wired in via `withWorkflow()` in this example's `next.config.ts`, and removing them would break the workflow. This is a re-post of the same finding that was already dismissed earlier on this PR, so closing it for the same reason. Won't fix.

arcjet-ai/skills/integrate-arcjet-ai/SKILL.md
davidmytton, line 1: Worth adding an explanation of what the `correlationId` is for, then the AI can decide which ID is best suited (or let us generate one).

arcjet-ai/skills/integrate-arcjet-ai/SKILL.md
```diff
@@ -0,0 +1,181 @@
...
+
+await protectAction(
+  arcjet,
+  ctx,
+  {
+    action: "review.submitted",
+    rules: [submitLimit({ key: repoId })],
+    metadata: securityMetadata({ destination: "github", reversibility: "compensable" }),
+  },
+  () => octokit.pulls.createReview(...),
```
davidmytton: This is the first and only time `octokit` appears. Worth commenting what it's representing, or use a generic example.

arcjet-ai/skills/integrate-arcjet-ai/SKILL.md
```diff
@@ -0,0 +1,181 @@
...
+(where effects go: `github`, `slack`, `internal`), `reversibility`
+(`reversible`/`compensable`/`irreversible`), `resource` (what's acted on,
+e.g. `repo:owner/name#123`). The `action` is not metadata — it is the guard
+label / capture action: `resource.verb` past tense, validated server-side as
+a slug (lowercase letters, digits, dash, and dot only — no underscores or
+uppercase). Use `order.looked-up`, not `order.looked_up`.
+
+Guard caps metadata server-side (max 20 pairs, key ≤64 bytes, value ≤512
+bytes), so keep maps small — merging `ctx.metadata` with per-call
+`securityMetadata()` can quietly exceed 20 pairs and the extras are dropped.
```
davidmytton: This is no longer accurate

arcjet-ai/README.md
```diff
@@ -0,0 +1,225 @@
...
+import { launchArcjet, tokenBucket } from "@arcjet/guard";
+import { tool, jsonSchema, generateText } from "ai";
+import {
+  createAiContext,
+  aiToolsContext,
+  protectTool,
+  protectAction,
+  captureAction,
+  securityMetadata,
+} from "@arcjet/ai";
```
qw-in: > protectTool

Should we call it "guardTool"? Since I think it uses "guard" rules?

arcjet-ai/skills/integrate-arcjet-ai/SKILL.md
```diff
@@ -0,0 +1,181 @@
...
+  }),
+};
+```
+
+- Omit `rules` for capture-only wrapping (audit without enforcement).
+- On DENY the tool's `execute` never runs; the model receives a structured
+  denial result (`arcjetDenied`, `reason`, `message`, `retryable`,
+  `retryAfterSeconds`) it can read and adapt to. Reshape it with `onDeny`.
+- Guard API failures fail open: the tool still runs, a warning is logged
+  when `ARCJET_LOG_LEVEL=warn`.
```
davidmytton: Worth revisiting this given the context - maybe our examples should fail closed given they are live enforcement events vs checking for bots on a website?

arcjet-ai/README.md
```diff
@@ -0,0 +1,225 @@
...
+
+// 1. Launch the guard client once (at module scope)
+const arcjet = launchArcjet({ key: process.env.ARCJET_KEY! });
+
+// 2. Create security context (at request entry point)
+const ctx = createAiContext({
+  correlationId: existingRunId, // omit to auto-generate
+  metadata: securityMetadata({ agent: "support", user: userId }),
+});
```
qw-in: I wonder if something like this will make sense (later) as a feature of the base sdk! Some nice way to link sequences together:

```ts
const sequence = createSequence()
```

arcjet-guard/src/agents/capture.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Fire-and-forget capture. Never throws. No-ops (with a gated warning) when
+ * the client predates `experimental_capture()`.
+ */
+export function captureEvent(client: ArcjetAiClient, opts: CaptureOptions): void {
+  if (typeof client.experimental_capture === "function") {
+    try {
+      client.experimental_capture(opts);
+    } catch {
+      // capture must never take the caller down.
+    }
```
qw-in: I see what you mean! Clever!

examples/nextjs-ai-agent/workflows/support-agent.ts
```diff
@@ -0,0 +1,112 @@
...
+        return lookupOrderRecord(orderNumber);
+      },
+    }),
+    {
+      action: "order.looked-up",
+      rules: () => [lookupLimit({ key: "demo-user", requested: 1 })],
+      metadata: ({ orderNumber }) =>
+        securityMetadata({ resource: `order:${orderNumber}` }),
+    },
+  ),
```
qw-in, line 61: I assume `protectTool` makes sense because it can automatically pull context out of the call.

The callbacks make sense (I presume I can do something like this):

```ts
rules: ({ inputParam }) => [moderateRule({ input: inputParam })],
```

My (personal) preference for guards would be to wire an explicit call in the `execute` block but I know many people would prefer something like this

qw-in (approved):
Have you had a look at the telemetry integrations for the ai sdk? It looks like it might be a quick way to handle the capture-side. Not sure if it can handle the `correlation_id` though (maybe if its available in context?)

https://ai-sdk.dev/docs/ai-sdk-core/telemetry#telemetry-integrations

---

Thanks for doing this! Its interesting to see how it comes together. Especially when thinking about the split between instrumentation (capture) and enforcement (guard).

---

Approving knowing that per the sync call you'll be trying to export this from the `@arcjet/guard` package directly something like `@arcjet/guard/vercel-ai` (not sure what casing is standard)

arcjet-ai/README.md
arcjet-rei: Yep. Incorporated into my plan for the move from `@arcjet/ai` -> `@arcet/guard/agents`, commits TBD.

arcjet-rei:
@qw-in 
> Have you had a look at the telemetry integrations for the ai sdk?

I had my good buddy Claude take a peek, and it turns that the answer is yes (we can do it on the capture side) and no (we can't do it on the guard side). `registerTelemetry()` does what we want, with the caveat that it ships the OTel data to the *user's* telemetry endpoint, not ours, so if we want to repurpose that for our own purposes we'll have to do a little behind the scenes surgery. That said, `enrichSpan()` can be made to use the `correlation_id`. On the guards side, we still need to get the various contexts in there, and that can't be done with the telemetry calls alone.

FYI, I'm backing the version into the paths and forcing callers to be explicit about versions, as part of my "never let AI assume anything ever" initiative: `@arcjet/guards/vercel-ai/v7`.

arcjet-ai/skills/integrate-arcjet-ai/SKILL.md
arcjet-rei: After the rework finishes, this will be controllable via policy: `"onGuardError": "allow"|"deny"`. The example will be updated to have it set to `"deny"` so we can show it being fail-close.

Here's a bit of authentic Claude gibberish for you to give me feedback on:
> One design call I'd like a second opinion on — how the fail-closed path surfaces:
> 
> - `guardTool` returns the *same* `ArcjetDenialResult` shape as a real denial, with
>   `reason: "ERROR"` and `retryable: true`. Rationale: a model only ever observes
>   tool results, and a second result type is harder to prompt against.
> - `guardAction` throws a *distinct* `ArcjetGuardUnavailableError` (carrying the
>   original error as `cause`), not `ArcjetDeniedError`. Rationale: "a rule denied
>   you" and "the policy couldn't be evaluated" are operationally different — usually
>   only the second is alert-worthy — and splitting them keeps
>  `ArcjetDeniedError.decision` non-optional.
> 
>  Reasonable? The asymmetry is deliberate but I could be talked out of it.

arcjet-ai/skills/integrate-arcjet-ai/SKILL.md
arcjet-rei: Here's the draft text of what will be added to the README:

> the correlation ID joins every guard decision and capture event from one
logical run into a single sequence in the console, so the best value is an ID the
app already has and can search by — request ID, job ID, ticket ID, review ID. Omit
it and we generate a ULID.

davidmytton:
> > Have you had a look at the telemetry integrations for the ai sdk?
> 
> I had my good buddy Claude take a peek, and it turns that the answer is yes (we can do it on the capture side) and no (we can't do it on the guard side). `registerTelemetry()` does what we want, with the caveat that it ships the OTel data to the _user's_ telemetry endpoint, not ours

Could we use [the custom integration hooks](https://ai-sdk.dev/docs/ai-sdk-core/telemetry#building-a-custom-integration)?

arcjet-ai/skills/integrate-arcjet-ai/SKILL.md
davidmytton: I suggest adding session as well:

> from one logical run or session

qw-in:
> I had my good buddy Claude take a peek, and it turns that the answer is yes (we can do it on the capture side) and no (we can't do it on the guard side).

Maybe just the `capture` side is enough to get us going? Or is the idea that this really should be testing the enforcement part as well? I'm not sure we've designed exactly how policies will enforce yet

> registerTelemetry() does what we want, with the caveat that it ships the OTel data to the user's telemetry endpoint, not ours, so if we want to repurpose that for our own purposes we'll have to do a little behind the scenes surgery. That said, enrichSpan() can be made to use the correlation_id. On the guards side, we still need to get the various contexts in there, and that can't be done with the telemetry calls alone.

```ts
import type { Telemetry } from 'ai';

class MyIntegration implements Telemetry {
  async onStart(event) {
    console.log('Generation started:', event.modelId);
  }
  // ...
```

[The custom integration](https://ai-sdk.dev/docs/ai-sdk-core/telemetry#building-a-custom-integration
) did look fairly compelling!

arcjet-ai/skills/integrate-arcjet-ai/SKILL.md
davidmytton: Agreed on splitting `ArcjetGuardUnavailableError` and `ArcjetDenialResult` because it gives visibility to decide what to do.

Will you make `onGuardError` default to `deny` ?

arcjet-rei:
## Change of approach: `@arcjet/ai` package → `@arcjet/guard` subpaths

Heads-up for anyone returning to an earlier review here: **the `@arcjet/ai` package this PR originally added no longer exists.** The `arcjet-ai/` workspace has been deleted and all of its behaviour now ships as subpath exports of the existing `@arcjet/guard` package. The PR title and description have been rewritten to match.

Why: a separate package meant a separate release cadence, a separate install, and a second place for the same concepts to drift. Subpaths give the same import ergonomics without any of that, and they make the vendor/version boundary explicit — `@arcjet/guard/<vendor-sdk>/v<major>` — so supporting a second SDK later is an additive change rather than a new package.

Practical consequence for reviewing: most inline comments are now marked outdated, because the files they were left on (`arcjet-ai/README.md`, `arcjet-ai/skills/integrate-arcjet-ai/SKILL.md`) are gone. The **feedback itself was not dropped** — here is where each thread landed.

### Addressed

- **@davidmytton — explain what `correlationId` is for.** Both the README and the packaged skill now carry the paragraph, using your "one logical run **or session**" wording: it joins every guard decision and capture event from one run or session into a single sequence in the console, so the best value is an ID the app already has and can search by (request ID, job ID, ticket ID, review ID). Omit it and a ULID is generated.
- **@davidmytton — the lone `octokit` reference.** Replaced with a generic external client, with a comment naming what it stands for.
- **@davidmytton — "this is no longer accurate" on the metadata caps.** Correct, and #6171 settled what replaces it. The stale "max 20 pairs / extras dropped" text is gone; README and skill now carry the same table — 128 top-level keys, 4 KiB serialized per value, nesting depth 10, key charset `letters, digits, - . _` — plus the points that change the guidance: metadata takes any JSON-serializable value including nested objects, drops surface on `decision.warnings` rather than silently, and metadata is untrusted and not redacted.
- **@davidmytton — "will you make `onGuardError` default to `deny`?"** Yes. It defaults to `"deny"`, on your reasoning that these wrap live enforcement events rather than page-view bot checks. `"allow"` is the opt-out, demonstrated in the example on a read-only order lookup where availability beats enforcement. The `ArcjetGuardUnavailableError` / `ArcjetDenialResult` split you agreed with is implemented as described, and "unavailable" covers both the `guard()` call throwing and a decision whose `hasFailedOpen()` is `true`.
- **@qw-in — "should we call it `guardTool`?"** Renamed throughout: `guardTool` / `guardAction`, with `GuardToolPolicy` / `GuardActionPolicy`. No `protect*` identifier remains anywhere.
- **@qw-in — is `rules: ({ orderNumber }) => [...]` supported, and can I wire the guard call explicitly instead?** Both yes. The `rules` callback does receive the tool's parsed input, and the example now shows it: `lookupOrder` derives its rate-limit key from `orderNumber` rather than discarding the argument as it did before. The explicit-call form is a first-class alternative, and it is written out in the README under "The explicit-call alternative", with the trade-off named — `guardTool` pulls the context out of the call automatically via the injected `contextSchema`, whereas an explicit `guardAction` inside `execute` keeps control flow visible but requires threading the context in by hand. The example points at it in a comment rather than duplicating it. Since you hit this reading the README the first time, the README is where it is spelled out.

### Carried forward, not in this PR

- **@qw-in — `createSequence()` as a base-SDK primitive for linking sequences.** Good idea and out of scope here; this PR keeps correlation as an explicit context object with no module state or `AsyncLocalStorage`. Worth its own issue.

### No action needed

- **@qw-in on `capture.ts`** — acknowledgement, nothing to change. That code moved to `arcjet-guard/src/agents/capture.ts`.

### Bot findings

- **CodeQL "unknown directive" on `"use step"` / `"use workflow"`.** Dismissed twice already and dismissed again: these are Workflow DevKit function-level directives consumed by `withWorkflow()` in `next.config.ts`. Removing them breaks the workflow.
- **Semgrep supply-chain on `sharp`/libvips in the example lockfile.** Pre-existing and not introduced by this branch — the same scan passed at `2326ba83` and began failing at `7dffe7ca`, a docs-only commit, which points at the advisory database updating rather than a dependency change here.

I have deliberately **not** resolved the threads themselves — that is @arcjet-rei's call once you have both had a chance to check the above against what actually shipped.

arcjet-guard/package.json
```diff
@@ -89,12 +98,26 @@
     "@connectrpc/connect-web": "2.1.2"
   },
   "devDependencies": {
+    "@ai-sdk/provider-utils": "5.0.12",
     "@types/node": "22.20.1",
+    "ai": "7.0.36",
```
qw-in, line 100: This is fine for now, in the future (assuming we keep following this pattern) we'll likely want to install different dependencies for different tests

arcjet-guard/README.md
```diff
@@ -506,6 +514,417 @@ Methods available on both `RuleWithConfig` and `RuleWithInput`:
...
+capture — `createAgentContext`, `guardAction`, `captureAction`, and
+`securityMetadata`:
+
+```ts
+import {
+  createAgentContext,
+  guardAction,
+  captureAction,
+  securityMetadata,
+} from "@arcjet/guard/agents";
```
qw-in: I don't think I'm quite understanding what this layer does that is distinct from the base `@arcjet/guard` api?

arcjet-guard/README.md
```diff
@@ -506,6 +514,417 @@ Methods available on both `RuleWithConfig` and `RuleWithInput`:
...
+const decision = await arcjet.guard({
+  label: "tools.chat",
+  rules: [
+    tokenBucket({ refillRate: 10, intervalSeconds: 60, maxTokens: 100 })({ key: userId, requested: 1 }),
+    detectPromptInjection()(userMessage),
+  ],
+});
+```
+
+### Agnostic helpers (`@arcjet/guard/agents`)
```
qw-in: I think I'd prefer to keep it all in vercel-ai until the abstraction proves itself over multiple different providers. At that point maybe we can just add it to the base SDK?

I think that could also help the docs maybe? I'm not sure users want/need to think about these layers other than "I'm using the Vercel AI SDK"

arcjet-guard/README.md
```diff
@@ -506,6 +514,417 @@ Methods available on both `RuleWithConfig` and `RuleWithInput`:
...
+  });
+  ```
+
+- **`@arcjet/guard/vercel-eve/v1`** (planned) — Vercel EVE integration
+
+### The namespace convention
+
+New vendor integrations follow the pattern `@arcjet/guard/<vendor-sdk>/v<major>`:
+
+- **Vendor-prefixed**: The namespace names the SDK being integrated
```
qw-in: I think some of this maybe belongs in `CONTRIBUTING.md` or `AGENTS.md`

arcjet-guard/README.md
```diff
@@ -506,6 +514,417 @@ Methods available on both `RuleWithConfig` and `RuleWithInput`:
...
+# or
+pnpm install --no-strict-peer-dependencies
+```
+
+### Why a separate `/agents` layer
+
+Importing `guardTool` loads the Vercel AI SDK as a dependency, so any consumer
+in a non-AI codebase that just wants `guardAction` and `captureAction` would be
+forced to install `ai` anyway. The `/agents` path provides those helpers and
+never reaches an AI SDK dependency, keeping your bundle clean.
```
qw-in: I think this is a fine tradeoff for the pilot personally

docs/test-plans/2026-07-23-pilot-framework-helper.md
qw-in, line 1: I this a new pattern that we are following?

examples/nextjs-ai-agent/workflows/support-agent.ts
```diff
@@ -0,0 +1,124 @@
+import {
+  aiToolsContext,
+  captureAction,
+  guardAction,
+  guardTool,
+  securityMetadata,
+} from "@arcjet/guard/vercel-ai/v7";
+import type { ArcjetAgentContext } from "@arcjet/guard/vercel-ai/v7";
```
qw-in:
```suggestion
import {
  aiToolsContext,
  captureAction,
  guardAction,
  guardTool,
  securityMetadata,
  type ArcjetAgentContext,
} from "@arcjet/guard/vercel-ai/v7";
```

qw-in (commented):
Thank you for making the changes! I'll go through the related ADRs which may answer some of my questions

examples/nextjs-ai-agent/workflows/support-agent.ts
```diff
@@ -0,0 +1,124 @@
...
+    prompt: input.question,
+    tools,
+    toolsContext: aiToolsContext(input.ctx, tools),
+    stopWhen: stepCountIs(3),
+  });
+  return result.text;
+}
+
+async function stepUpdateTicket(input: SupportAgentInput, answer: string) {
+  "use step";
```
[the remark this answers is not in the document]
arcjet-rei: This directive is used by the Vercel AI SDK, so this finding is incorrect.

arcjet-guard/package.json
arcjet-rei: The supported pattern is `peerDependencies`, which is to say the supported pattern sucks. Do you have a preference for how we handle this?

arcjet-guard/README.md
arcjet-rei: This is an artifact of the mechanical translation from `@arcjet/ai` -> `@arcjet/guard`. I'll take a look at moving them up to the `@arcjet/guard` namespace (but not making them part of the exported interface) and exposing them publicly via `@arcjet/guard/vercel-ai/v7`. Does that work for you?

arcjet-guard/README.md
arcjet-rei: Yes. Per previous comments, the approach I'm going to follow is moving the actual functionality into guard, but publicly expose it via the specific SDK. I think the plagform-agnostic machinery would make more sense if people were using multiple AI SDKs in a single application, but they're not doing that. Yet. I hope.

arcjet-guard/README.md
arcjet-rei: It probably belongs in some kind of design notes document. I'll think about this. You're right that wherever it goes, it doesn't belong in README.md.

docs/test-plans/2026-07-23-pilot-framework-helper.md
arcjet-rei: This is machinery leftover from the [skills](https://github.com/ed3dai/ed3d-plugins/blob/main/plugins/ed3d-plan-and-execute/README.md) I used to build out the code. It'll be removed before the PR lands.

arcjet-guard/README.md
arcjet-rei: Say more? Which part is the fine tradeoff?

arcjet-guard/src/agents/guarded.ts
```diff
@@ -0,0 +1,167 @@
...
+  } = params;
+  // Spread onto every guard/capture payload so `correlationId` is included
+  // when set and omitted otherwise (it is optional under
+  // `exactOptionalPropertyTypes`, so assigning `undefined` is a type error).
+  const correlation = correlationId === undefined ? {} : { correlationId };
+
+  const failClosed = onGuardError === "deny";
+
+  let decisionId: string | undefined;
+  if (rules !== undefined && rules.length > 0) {
```
davidmytton: Can we always call `guard()` here, using `rules ?? []`, rather than treating no rules as capture-only?

I'm thinking about how we include remote rules-like policies, so `guardTool()` and `guardAction()` can become injection points for remote rules. 

This probably also requires removing the empty-rules early return in `client.ts`, updating the `GuardActionPolicy`/`GuardToolPolicy` docs, plus tests proving an omitted or empty rules array still reaches the Guard RPC.

What do you think @qw-in ? We discussed this in [the Guards ADR](https://github.com/arcjet/arcjet/blob/main/docs/adrs/2026-03-18-guards-sdk-surface.md#L68-L73) and left it open in [the Capture ADR](https://github.com/arcjet/arcjet/blob/main/docs/adrs/2026-07-09-capture-trace-event-primitive.md#L127-L135).

The goal here is to not close this off so it has to be reversed.

qw-in: I think this gets into how we want to handle sequence/policy enforcement. Do we want a rule (or rules) for policies or remote rules? Something like this:

```tsx
import { policy } from "@arcjet/guard";

const p = policy("no-refund-if-suspicious")

await arcjet.guard({ rules: [p] });
```

---

For now, I'm happy to see this include a change that accepts an empty rule call to `guard` (we'd have to verify the server handles it gracefully but I expect it does)

arcjet-guard/README.md
qw-in: We've resolved this by dropping the `/agents` layer

arcjet-guard/package.json
qw-in: Yeah, peerDependencies (optional) makes sense. I'm thinking for the devDependencies for testing but we can cross that bridge when we come to it!

arcjet-guard/src/agents/guarded.ts
davidmytton: I'm thinking something like this to wrap the tool with a policy:

```ts
guardTool(arcjet, sendEmailTool, {
  action: "email.sent",
  actor: userId,
  inputs: ({ ... }) => ({
    ...
  }),
});
```

with a generic contract like:

```ts
interface GuardToolOptions<TArgs> {
  action: string;
  actor?: PolicyInput | (() => PolicyInput);
  inputs?: (args: TArgs) => Record<string, PolicyInput>;
}
```

The developer provides the inputs and actor, then the policy can be defined remotely. The challenge is giving the inputs to the policy because we can't impose policies remotely without them - this is why we can't do email verification rules remotely today for example. The goal is to split the policy content (owned by security, configured remotely) from the input/handling of the result (owned by engineering, implemented in code).

Is that how you were thinking about it?

examples/nextjs-ai-agent/package-lock.json
```diff
@@ -0,0 +1,8143 @@
...
+    "node_modules/negotiator": {
+      "version": "1.0.0",
+      "resolved": "https://registry.npmjs.org/negotiator/-/negotiator-1.0.0.tgz",
+      "integrity": "sha512-8Ofs/AUQh8MaEcrlq5xOX0CQ9ypTF5dl78mjlMNfOK08fzpgTHQRQPBxcPlEtIw0yRpws+Zo/3r+5WRby7u3Gg==",
+      "license": "MIT",
+      "engines": {
+        "node": ">= 0.6"
+      }
+    },
+    "node_modules/next": {
```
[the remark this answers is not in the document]
arcjet-rei: Upgraded `next` to 16.2.12 in 901f6b3b, above the 16.2.11 fix line.

The precondition you name (App Router with at least one Server Action) does not hold here — the example declares no `"use server"` anywhere. I upgraded rather than marking it a false positive: this is an example people copy, and "not currently reachable" is a weaker guarantee than "not vulnerable".

Typecheck and `next build` are green on 16.2.12.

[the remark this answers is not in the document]
arcjet-rei: Upgraded `next` to 16.2.12 in 901f6b3b, above the 16.2.11 fix line.

The precondition you name does not hold here — `examples/nextjs-ai-agent/next.config.ts` defines no `rewrites()` or `redirects()` at all:

```ts
const nextConfig: NextConfig = {
  reactStrictMode: true,
};

export default withWorkflow(nextConfig);
```

Upgraded anyway rather than marking it a false positive, for the same reason as the sibling `next` finding.

examples/nextjs-ai-agent/package-lock.json
```diff
@@ -0,0 +1,8143 @@
...
+        "url": "https://opencollective.com/express"
+      }
+    },
+    "node_modules/setprototypeof": {
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/setprototypeof/-/setprototypeof-1.2.0.tgz",
+      "integrity": "sha512-E5LDX7Wrp85Kil5bhZv46j8jOeboKq5JMmYM3gVGdGH8xFpPWXUMsNrlODCrkoxMEeNi/XZIwuRvY4XNwYMJpw==",
+      "license": "ISC"
+    },
+    "node_modules/sharp": {
```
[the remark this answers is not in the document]
arcjet-rei: Not fixable from this lockfile, and not reachable here.

`sharp` is an `optionalDependencies` entry of `next` itself, constrained to `^0.34.5`:

```
node_modules/next  [optionalDependencies]  sharp ^0.34.5
```

The fixed 0.35.0 is outside that range, so it cannot be resolved by upgrading anything this example declares — including the `next` bump to 16.2.12 in 901f6b3b, which still asks for `^0.34.5`. Forcing 0.35.x through an `overrides` block would pin a version Next has not tested against, in an example app. The real fix is upstream in Next.js.

On reachability: every CVE listed is in an image-decoding path (`vipsload`, `gifload`, `tiffload`, the EXIF directory decoder), reached by handing sharp an attacker-supplied image. This example never invokes image optimization — no `next/image`, no `<Image>`, no image route — so there is no call site that feeds it one. Worth re-checking if the example ever gains one.

examples/nextjs-ai-agent/workflows/support-agent.ts
```diff
@@ -0,0 +1,125 @@
...
```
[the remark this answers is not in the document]
arcjet-rei: False positive — `"use step"` is a real directive, and applying the suggested fix would break the file.

It comes from the Vercel Workflow SDK (`workflow@4.6.1`), wired up by `withWorkflow(nextConfig)` in `next.config.ts`. The directive is what marks a function as a durable, independently-retried step; it is not decorative. The SDK uses it in its own stdlib —

```js
// node_modules/workflow/dist/stdlib.js
export async function fetch(...args) {
    'use step';
    return globalThis.fetch(...args);
}
```

— and its runtime errors instruct users to add it: *"Move this call to a step function (`"use step"`) or call it outside the workflow context."*

Replacing line 87 with a comment, as suggested, would silently demote `stepUpdateTicket` from a durable step to an inline call. The sibling `"use workflow"` on line 65 is the same family of directive.

Leaving both as-is.

arcjet-guard/README.md
arcjet-rei: Moved to `CONTRIBUTING.md` in https://github.com/arcjet/arcjet-js/pull/6164/commits/5a5fa634ac2da29e5ee4bea23805616ae4570b74.

arcjet-ai/README.md
arcjet-rei: Captured in [ENG-1005](https://linear.app/arcjet/issue/ENG-1005/evaluate-promoting-correlation-to-a-base-sdk-sequence-primitive). This feels like a fairly simple first step on the road to multi-sequences.

arcjet-ai/skills/integrate-arcjet-ai/SKILL.md
arcjet-rei: Added [here](https://github.com/arcjet/arcjet-js/blob/901f6b3b/arcjet-guard/README.md#L780-L787) and [here](https://github.com/arcjet/arcjet-js/blob/901f6b3b/arcjet-guard/skills/integrate-arcjet-guard-agents/SKILL.md#L74-L77) as part of 901f6b3b – edited to "one logical run or session" in both places. (Rei wrote this. Rei still likes en dashes and refuses to give them up. Rei only feels a little weird writing about themself in the third person.)

arcjet-ai/skills/integrate-arcjet-ai/SKILL.md
arcjet-rei: Edited [here](https://github.com/arcjet/arcjet-js/blob/901f6b3b/arcjet-guard/skills/integrate-arcjet-guard-agents/SKILL.md#L183-L184) (in 901f6b3b) to replace `octokit` with a generic `externalServiceClient`.

arcjet-ai/skills/integrate-arcjet-ai/SKILL.md
arcjet-rei: Brought up to date (here)[https://github.com/arcjet/arcjet-js/blob/901f6b3b1984f30386e2708239efc397fab182ca/arcjet-guard/README.md?plain=1#L450-L459) and [here](https://github.com/arcjet/arcjet-js/blob/901f6b3b1984f30386e2708239efc397fab182ca/arcjet-guard/skills/integrate-arcjet-guard-agents/SKILL.md?plain=1#L238-L246). Now correct with respect to the implementation.

arcjet-ai/skills/integrate-arcjet-ai/SKILL.md
arcjet-rei: This is now documented in [README.md](https://github.com/arcjet/arcjet-js/blob/901f6b3b1984f30386e2708239efc397fab182ca/arcjet-guard/README.md?plain=1#L725-L753), the [shipped skill](https://github.com/arcjet/arcjet-js/blob/901f6b3b1984f30386e2708239efc397fab182ca/arcjet-guard/skills/integrate-arcjet-guard-agents/SKILL.md?plain=1#L121-L125), and the [sequences example](https://github.com/arcjet/arcjet-js/blob/901f6b3b1984f30386e2708239efc397fab182ca/arcjet-guard/skills/integrate-arcjet-guard-agents/SKILL.md?plain=1#L121-L125). All of these are written in light of the fail-open ADR.

arcjet-ai/README.md
arcjet-rei: There is now [a test case](https://github.com/arcjet/arcjet-js/blob/901f6b3b/arcjet-guard/src/vercel-ai/v7/index.test.ts#L262) that will fail on any use of the old names, in addition to having Claude do a sweep + rename. (Don't give me any credit for this, Claude decided to do it on its own.)

examples/nextjs-ai-agent/workflows/support-agent.ts
arcjet-rei: Yes. See [this call in the example](https://github.com/arcjet/arcjet-js/blob/901f6b3b/examples/nextjs-ai-agent/workflows/support-agent.ts#L55), documented [here](https://github.com/arcjet/arcjet-js/blob/901f6b3b1984f30386e2708239efc397fab182ca/arcjet-guard/README.md?plain=1#L790-L805) in `README.md`. Your explicit call pattern is also [documented in `README.md`](https://github.com/arcjet/arcjet-js/blob/901f6b3b1984f30386e2708239efc397fab182ca/arcjet-guard/README.md?plain=1#L755-L778).

arcjet-guard/src/agents/guarded.ts
arcjet-rei: David's requests at the top of the thread are implemented in fdf3cc34. A few things to know:
1. The back end already supported an empty rule list. It returns an empty ALLOW list along with response error AJ1002, and the SDK maps response-level errors to decision-level warnings. `hasFailedOpen()` stays false, and doesn't conflict with the fail-closed default because there isn't an error.
2. To make this work, the early return in `client.ts` [had to go as well](https://github.com/arcjet/arcjet-js/blob/901f6b3b/arcjet-guard/src/client.ts#L127-L135), because it would have returned `failOpen()`.
3. A no-rules call now gets a decision ID, so it's correlatable to a sequence.
4. All the tests @davidmytton asked for are now there, and in addition there's a couple extra test to verify that the various bits of the client-server dance work both in isolation and in combination.

arcjet-rei: As for connecting policies defined remotely to input / rule configuration in the code, I think we should be thinking very carefully about the trust boundaries as well as the social boundaries (i.e. making sure that security teams get the trust boundaries and mandatory enforcement they expect, and doing what we can to make sure that application developers don't have to deal with too much extra ceremony to implement the policies in code).  

This is something we want to push on when we talk to a security lead – are they fine with allowing app developers to have the ability to define the terms and inputs of a remotely-defined policy in application code? Or do they want them to be a black box where the application developers pass in the details of tool calls and then the back end / remote policy definition matches against the data it cares about? I can't speak about this definitively, but I have a hunch that a lot of security teams are going to prefer to retain as much control as possible when it comes to policy definition and enforcement.

arcjet-guard/src/agents/capture.ts
arcjet-rei: It sure was! Too bad it had to die (so that `capture()` could live)!

examples/nextjs-ai-agent/package-lock.json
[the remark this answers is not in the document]
arcjet-rei: /ar Present but unreachable here, and not fixable from this lockfile.

`sharp` is an `optionalDependencies` entry of `next` itself, pinned to `^0.34.5`:

```
node_modules/next  [optionalDependencies]  sharp ^0.34.5
```

The fixed 0.35.0 is outside that range, so nothing this example declares can resolve it — including the bump to `next@16.2.12` in 901f6b3b, which still asks for `^0.34.5`. Forcing 0.35.x through an `overrides` block would pin a version Next has not tested against, in an example app. The real fix is upstream in Next.js widening that range.

On reachability: every CVE listed is an image-decoding path (`vipsload`, `gifload`, `tiffload`, the EXIF directory decoder), and `sharp` is only invoked by Next's image optimizer. In `examples/nextjs-ai-agent` there is no path that reaches it — no `next/image` or `<Image>` usage, no `public/` directory, no `images` config (so `/_next/image` has no `remotePatterns` and rejects remote URLs), and no upload or file-handling route. The app has three routes, all returning JSON or HTML. Nothing attacker-controlled can become bytes handed to a libvips loader.

Scope: this is a demo app in the monorepo. It is not published to npm and is not part of any Arcjet SDK.

Re-triage if the example gains any of: `next/image` usage, an image under `public/`, `images.remotePatterns`, or a route that accepts uploads.

docs/test-plans/2026-07-23-pilot-framework-helper.md
arcjet-rei: Removed in 6cd817dd.

arcjet-guard/src/agents/guarded.ts
arcjet-rei: This PR is ready to merge, but I didn't want to push it in until this conversation has completed. Resolve the conversation if / when you're ready for the PR to land.

examples/nextjs-ai-agent/package-lock.json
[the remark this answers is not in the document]
arcjet-rei: Correcting my previous reply: this was fixable, so it is fixed rather than triaged. e382d4d0 overrides `sharp` to `^0.35.0`, above the 0.35.0 fix line.

I said an `overrides` block would pin a version Next has not tested against. That was overcautious — it is how this repo already patches transitive advisories (the root `package.json` carries eight, `express-newman` seven, `nestjs` three, and every Next example inherits `postcss` via `"$postcss"`). Verified after the override: no peer or `ERESOLVE` warnings, typecheck clean, `next build` emits the same six routes.

Please disregard the `/ar` above — the acceptable-risk argument rested on it being unfixable from this lockfile, which was wrong.

One thing worth separating out: eleven other example lockfiles and the root all carry the same `sharp@0.34.5` via `next`. This finding surfaces here only because the scan is limited to changed files and this PR adds a new lockfile. Fixing the rest means touching twelve lockfiles and should be its own security PR rather than a rider on this one.

arcjet-rei:
From your friend and mine, Claude:

> @davidmytton @qw-in — coming back to this, and correcting myself first: the
> "behind the scenes surgery" I described was wrong. You were both right that the
> custom integration hooks are the answer for the capture side.
> 
> ## Yes to the custom integration hooks
> 
> `TelemetryOptions.integrations` is per-call and takes precedence over anything
> registered globally, so we pass our own `Telemetry` object and the SDK invokes
> our callbacks directly. The user's OTel endpoint is never involved and nothing
> needs re-routing. Against the installed `ai@7.0.36`:
> 
>     // ai/dist/index.d.ts:982, JSDoc :977-980
>     /** Per-call telemetry integrations that receive lifecycle events during
>      *  generation. When provided, these integrations will take precedence over
>      *  the globally registered integrations for this call. */
>     integrations?: Arrayable<Telemetry>;
> 
> `onToolExecutionEnd` carries `toolExecutionMs` and a discriminated union on
> `success`, which maps onto the outcomes we already capture. And
> `includeToolsContext` (`:975`) surfaces tool-context properties into the
> events — which is exactly where `aiToolsContext` already puts the
> `correlationId`, so correlation needs no new plumbing.
> 
> The "different destination" objection I raised applies to the `@ai-sdk/otel`
> integration specifically, not to telemetry as a mechanism.
> 
> ## Enforcement is a different matter, and worth being precise about
> 
> @qw-in — to your question, capture alone would get us going. But it can't grow
> into enforcement later, so it's worth pinning down why now.
> 
> `onToolExecutionStart` looks like the seam: it's documented as firing "before
> the tool's `execute` function is invoked." Two things close it off. Every hook
> is a `Callback`, which has no return channel —
> 
>     // ai/dist/index.d.ts:903
>     type Callback<EVENT> = (event: EVENT) => PromiseLike<void> | void;
> 
> — and a throw doesn't help either, because `notify()` awaits each callback
> inside an empty `catch`:
> 
>     // ai/dist/index.js:2647-2655
>     async function notify(options) {
>       await Promise.all(asArray3(options.callbacks).map(async (callback) => {
>         try { await callback?.(options.event); } catch (e) {}
>       }));
>     }
> 
> So the hook can watch a tool call it's about to lose and has no way to stop it.
> 
> The one honest exception is `Telemetry.executeTool` (`:4288`), which receives
> `execute: () => PromiseLike<T>` and could simply not call it. But it's a
> tracing wrapper — its documented purpose is nesting spans when a tool calls
> `generateText` — and it's one hook for every tool in a call rather than
> per-tool, it can't inject the `contextSchema` `guardTool` depends on, and
> denying would mean synthesising a tool output through a hook whose contract is
> "run this and return its result." Wrapping `execute` stays the right mechanism.
> 
> That splits the work cleanly:
> 
> - **Capture** — telemetry integrations are a real option. ENG-992 has the full
>   detail, including the one blocker that does hold and the open question this
>   actually creates: `guardTool` already emits a capture event per outcome, so a
>   telemetry integration emitting tool-execution events would duplicate them.
>   Deciding which is canonical is the substance of that issue.
> - **Enforcement** — the SDK's real pre-execution hook is `toolApproval`, not
>   telemetry. ENG-993. Its approval functions do return a decision, and they see
>   `runtimeContext`, which `execute` does not.
> 
> ## On "I'm not sure we've designed exactly how policies will enforce yet"
> 
> Agreed, and still true. That's the live thread with @davidmytton about `guard()`
> becoming an injection point for remotely-defined policy — neither of those two
> issues settles it.

davidmytton:
Policy enforcement is implemented in https://github.com/arcjet/arcjet/pull/8676 and https://github.com/arcjet/arcjet-js/pull/6186 for JS.

davidmytton (approved):
I have implemented the SDK helpers with a policy enforcement wrapper example from https://github.com/arcjet/arcjet-js/pull/6186 in https://github.com/arcjet/arcjet-js/pull/6191.

arcjet-guard/src/agents/guarded.ts
arcjet-rei: Policy enforcement and remote policies are stacked on this PR, so this thread has been addressed in #6186 and #6191; resolving this thread and merging the PR.

PR #6176, opened 2026-07-27
feat(guard): batch and flush capture delivery

qw-in:


arcjet-guard/src/capture-delivery.ts
```diff
@@ -0,0 +1,279 @@
...
+  batchSize?: number;
+  /** Longest an event waits for a batch to fill. */
+  batchDelayMs?: number;
+}
+
+const DEFAULT_QUEUE_SIZE = 1000;
+const DEFAULT_BATCH_SIZE = 50;
+const DEFAULT_BATCH_DELAY_MS = 100;
+
+interface InFlightBatch {
```
qw-in: Prefer types unless it is an interface that will be implemented.

arcjet-guard/README.md
```diff
@@ -329,6 +329,58 @@ const decision = await arcjet.guard({
...
+a platform `waitUntil` hook, the SDK keeps a bounded in-memory queue, sends
+batches on size or delay, drops the newest event when the queue is full, and
+never retries a failed batch. On a discoverable `waitUntil` platform, it starts
+the send immediately and hands the promise to the platform.
+
+Call `flush()` during graceful shutdown to avoid losing the final batch:
+
+```ts
+await arcjet.flush(); // one-second deadline by default
+await arcjet.flush(250); // custom deadline in milliseconds
```
qw-in, line 403: Should we rename it to `persist`? Or `sync`?

arcjet-guard/src/diagnostics.ts
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Build the diagnostics channel for one client.
+ *
+ * A caller-provided logger receives every diagnostic. The console fallback
+ * logs once per code so repeated best-effort drops cannot flood application
+ * logs.
+ */
+export function createDiagnosticHandler(logger?: ArcjetDiagnosticLogger): DiagnosticHandler {
+  const logged = new Set<ArcjetDiagnostic["code"]>();
```
qw-in: Could we use `@arcjet/logger`?

arcjet-guard/src/diagnostics.ts
```diff
@@ -0,0 +1,56 @@
+/**
+ * A local SDK diagnostic that cannot be reported over the wire.
+ *
+ * Messages contain static text and, for metadata encoding warnings, escaped
+ * and length-bounded key names. They never include metadata values, capture
+ * actions, credentials, headers, or request bodies.
+ */
+export interface ArcjetDiagnostic {
+  /** Stable machine-readable code. */
```
qw-in: Prefer `type` unless it really is an interface

arcjet-guard/src/client.ts
```diff
@@ -44,6 +46,10 @@ export interface GuardClientOptions {
   transport: Transport;
   /** User-agent product token (e.g. `"arcjet-guard-js/0.1.0"`). */
   userAgent?: string;
+  /** Local diagnostics sink. */
+  logger?: ArcjetDiagnosticLogger;
```
qw-in: Could we just use `@arcjet/logger`?

arcjet-guard/src/capture-delivery.ts
```diff
@@ -0,0 +1,279 @@
...
+}
+
+/**
+ * Bounded, send-once delivery for best-effort capture events.
+ *
+ * Calls enqueue synchronously. A full queue drops instead of blocking, batches
+ * send on size or delay, failed sends are never retried, and the timer is
+ * unref'd where the runtime supports it.
+ */
+export class CaptureDelivery {
```
qw-in: This seems very complex. Do we really need all of this? Is there a reference design we can look at in some open source repo that is already battle tested?

qw-in (commented):
Seems complex

arcjet-guard/src/capture-delivery.ts
qw-in: (Claude, replying on Quinn's behalf.)

This file has no `interface` in it — `WaitUntil`, `CaptureDeliveryOptions`, `CaptureDelivery` and `PendingBatch` are all `type` aliases, and `git log -p` on the file shows there never was one. The thread has no line anchor, so I think it's pinned to a stale diff position.

Same for `diagnostics.ts` (the other thread): `ArcjetDiagnostic`, `DiagnosticLogger` and `DiagnosticHandler` are all `type`.

The `interface` declarations nearby are `LaunchOptions` and `ArcjetGuard` in `index.ts`. Those are pre-existing on `main` rather than introduced here — this PR only adds `capture()`/`flush()` members to `ArcjetGuard`. Happy to convert them in a follow-up if you want the rule applied there, but it'd be churn unrelated to capture.

arcjet-guard/README.md
qw-in: (Claude, replying on Quinn's behalf.)

I'd keep `flush`, for three reasons:

1. **It's the term of art.** OpenTelemetry uses `forceFlush()`, Sentry `flush()`, pino `flush()`, statsd clients `flush()`. Someone wiring up graceful shutdown greps for "flush".
2. **`persist` overpromises.** It implies the events are durably stored when the call returns. They are not: `CaptureResponse` is an empty ack meaning "received", and the server can still drop them afterwards. I confirmed that live — a capture request with a deliberately invalid key is accepted with an empty response, indistinguishable from a valid one. Naming it `persist` would claim a guarantee the protocol doesn't give.
3. **`sync` is ambiguous** — it reads as "synchronous", and this SDK already has a sync/async split elsewhere.

What the method actually does is "stop buffering and wait for what's queued, up to a deadline", which is what `flush` conveys to anyone who's used a telemetry SDK.

arcjet-guard/src/diagnostics.ts
qw-in: (Claude, replying on Quinn's behalf.)

Yes — done, and you were right that it was duplicated work. The fallback was `console.warn` with a hand-rolled `✦Aj WARN [CODE] message (N event(s))` string, which re-implemented the prefix `@arcjet/logger` already produces.

The default is now an `@arcjet/logger` instance. That deletes the local formatter, and `code`/`count` become structured fields on the log line instead of text interpolated into the message. `@arcjet/logger` was already a runtime dependency of this package, so it costs nothing.

Two details worth flagging:

- **Level is fixed at `"warn"`** rather than `logLevel(process.env)` like the framework adapters do. Every diagnostic is a warning, and using the env helper would mean adding `@arcjet/env` as a dependency plus reading `process.env` in a package that also runs on edge runtimes. A caller who wants different handling passes their own logger.
- **The logger is built on first use**, so a client that never reports a diagnostic — the expected case — never constructs one.

Behaviour is otherwise unchanged: a caller-supplied logger gets every diagnostic (they control filtering), and the default reports each code once so a persistent problem can't turn a best-effort drop into a logging incident.

arcjet-guard/src/diagnostics.ts
qw-in: (Claude, replying on Quinn's behalf.)

`diagnostics.ts` has no `interface` — `ArcjetDiagnostic`, `DiagnosticLogger` and `DiagnosticHandler` are all `type` aliases. I think this thread is pinned to a stale diff position (it has no line anchor). See the `capture-delivery.ts` thread for the same answer and where the real `interface` declarations are.

arcjet-guard/src/client.ts
qw-in: (Claude, replying on Quinn's behalf.)

It does now — see the `diagnostics.ts` thread for the detail. Short version: the default sink is an `@arcjet/logger` instance instead of a hand-rolled `console.warn` formatter.

On this line specifically: `logger?: DiagnosticLogger` is `Pick<Logger, "warn">` where `Logger` is imported from `@arcjet/logger`, so it already *is* that type, narrowed to the one method used. Any `@arcjet/logger` instance satisfies it, and narrowing keeps the option easy to satisfy with a test double. Happy to widen it to the full `Logger` if you'd rather the option be unambiguous about what to pass.

arcjet-guard/src/capture-delivery.ts
qw-in: (Claude, replying on Quinn's behalf.)

Fair challenge. There is a reference design, and this is it:

**OpenTelemetry's `BatchSpanProcessor`** (and `BatchLogRecordProcessor`), which is the same shape with the same four knobs — `maxQueueSize` (bounded, drops when full), `maxExportBatchSize`, `scheduledDelayMillis`, and `forceFlush()`. It's implemented across every OTel language SDK and is about as battle-tested as this pattern gets. Sentry's transport and most statsd clients are the same idea with fewer knobs.

Ours is that design **minus** retry and **plus** `waitUntil`.

What's non-negotiable is the four behaviours, because they're the capture ADR's amendments 3 and 4: bounded queue, drop when full, batch, never retry, and `flush()` drains. Take any one out and it stops matching the decision.

Where I agree there's genuine extra weight is `waitUntil`, which OTel has no equivalent of. It exists because on Cloudflare and Vercel, background work after the response is killed, so without it a queued event is simply lost. **If you'd rather require callers to pass `waitUntil` explicitly, the ~40 lines of discovery (the Vercel symbol lookup plus its two type predicates) delete cleanly.** Say the word and I'll cut it.

One thing I tried and backed out of, since it's the obvious simplification: collapsing those two type predicates into the inline `typeof` checks that `lookupWaitUntil` in the `arcjet` package uses. It doesn't work here — inline narrowing leaves `waitUntil` typed as `Function`, and this package runs the type-aware lint rules, so calling it trips `no-unsafe-call`. The predicates are load-bearing, not ceremony; the `arcjet` copy only gets away without them because root lint skips type-aware rules for that package. I've noted that in the code so the next person doesn't retry it.

PR #6178, opened 2026-07-27
feat: add optional Arcjet client registration

qw-in:


arcjet/README.md
```diff
@@ -526,6 +526,33 @@ server.listen(8000);
...
+  unregisterArcjet,
+} from "arcjet";
+
+const client = launchArcjet({ key: process.env.ARCJET_KEY! });
+registerArcjet(client);
+
+capture({ action: "refund.issued", correlationId: "workflow_123" });
+await flush();
+
+unregisterArcjet(client);
```
qw-in: We should not need a reference to the client to unregister it.

arcjet/src/guard/registry.ts
```diff
@@ -0,0 +1,105 @@
...
+  createDiagnosticHandler,
+  symbolArcjetDiagnostics,
+  type ArcjetDiagnostic,
+  type DiagnosticHandler,
+} from "./diagnostics.ts";
+import type { ArcjetClient } from "./index.ts";
+import type { CaptureOptions, Decision, GuardOptions } from "./types.ts";
+
+/** One slot shared by duplicate package copies in the same JavaScript realm. */
+const CLIENT_SLOT = Symbol.for("arcjet.client");
```
qw-in, line 12: Shouldn't this live in the `arcjet` package?

qw-in (commented):
Mostly makes sense but will have to check if adding `@arcjet/guard` as a dependency makes sense?

Instead I'd kind of expect a lot of the code to get pulled into `arcjet` and `@arcjet/guard` to become a pretty shallow wrapper

PR #6179, opened 2026-07-27
test: add registered Arcjet test client

qw-in:


qw-in (commented):
Like the previous pr I'd expect it to work like this:

```ts
import { launchTestClient } from "arcjet/testing" // actual impl
import { launchTestClient } from "@arcjet/guard/testing" // re-export

```

PR #6206, opened 2026-08-04
feat(guard): add optional client registration and a test client

qw-in:
See ADR: https://github.com/arcjet/arcjet/blob/main/docs/adrs/2026-07-24-unified-arcjet-client-registration-and-lifecycle.md

Note: It is only added to Guard here as having a single client is happening elsewhere

```ts
import { launchArcjet, registerArcjet } from "@arcjet/guard";

const arcjet = launchArcjet(/* ... */);

// Stored on globalThis
registerArcjet(arcjet);
```

```ts
import { capture, guard } from "@arcjet/guard";

// ... later

cature(/* ... */) // Pulls client off globalThis
```

```ts
import { registerTestClient } from "@arcjet/guard/testing";
import { refund } from "./refund.ts";

test("refund captures an event", async () => {
  using arcjet = registerTestClient();

  await refund("inv_1");

  assert.equal(arcjet.captures[0]?.action, "refund.issued");
});
```

### DIFF
--- a/arcjet-guard/README.md
+++ b/arcjet-guard/README.md
@@ -474,14 +474,24 @@ if (decision.hasFailedOpen()) {
 }
 ```
 
-`capture()` drops the event, and `flush()` resolves immediately. Both report
-`AJ3005` locally. This is the one failure the SDK cannot report well: with no
-client there is no configured logger either, so it can only reach the default
-sink.
+`capture()` drops the event silently, and `flush()` resolves immediately.
+Nothing is logged: the client that would have carried a logger is the thing
+that is missing, so the only available sink would be an unconfigurable console
+warning on a request path — noise an application cannot turn off. The decision
+returned by `guard()` is the observable signal, and making the `capture()` case
+observable is planned as an opt-in on the call itself.
 
 ### Registering twice, and unregistering
 
-Registration is guarded. A second client does not displace the first — the
+Registration is version-checked. The slot is shared by every copy of
+`@arcjet/guard` in the process, so a registration is only used by the exact
+build that wrote it — the stored value is a live object whose internals are
+guaranteed within one build and not across them. A copy that finds a
+registration from another version leaves it alone and fails open, exactly as if
+nothing were registered, and reports `AJ3006` on its own logger. Two versions
+in one process therefore do not share a client.
+
+Registration is also guarded. A second client does not displace the first — the
 attempt is reported as `AJ3004` on the **incumbent's** logger, so a library or a
 stray second `launchArcjet()` cannot quietly redirect an application's telemetry
 to a different key. Registering the client that is already registered is a
@@ -537,9 +547,15 @@ The client also implements `Symbol.dispose`, so `using` replaces the `finally`:
 using arcjet = registerTestClient();
 ```
 
-Note that the `using` *syntax* needs Node.js 24 or compilation through
-TypeScript; Node.js 22 defines `Symbol.dispose` but cannot parse `using`. Call
-`unregister()` directly if you are targeting Node.js 22 without a compile step.
+`unregister()` and `[Symbol.dispose]` are the same function under two names, so
+neither can drift from the other.
+
+Two caveats, both on the consuming project rather than on this package. The
+`using` *syntax* needs Node.js 24 to run natively or compilation through
+TypeScript — Node.js 22 defines `Symbol.dispose` but cannot parse `using`. And
+because `[Symbol.dispose]` appears in the published types, a project compiling
+with `skipLibCheck: false` needs `esnext.disposable` in its `lib` even if it
+never writes `using`; `unregister()` is unaffected either way.
 
 `guard()` on the test client records the call and returns a fail-open `ALLOW`,
 because no rule actually ran. It is not a mock server and does not let you stub
--- a/arcjet-guard/README.md
+++ b/arcjet-guard/README.md
@@ -427,6 +427,126 @@ cannot encode is reported locally as `AJ1017` and also travels with that event i
 `local_warnings`. A queue-full event or failed batch never reaches the server, so
 those drops can only be reported locally.
 
+## Registering a client (optional)
+
+Passing the client explicitly is the recommended path, and everything above does
+exactly that. Registration is a shortcut for the case it cannot cover: code too
+deep in an application to be handed a client, where `capture()` is often most
+useful.
+
+`launchArcjet()` never touches global state. Registering is always a separate,
+explicit call:
+
+```ts
+// instrumentation.ts, or whatever runs at startup
+import { launchArcjet, registerArcjet } from "@arcjet/guard";
+
+const arcjet = launchArcjet({ key: process.env.ARCJET_KEY! });
+
+registerArcjet(arcjet); // now, and only now, something is global
+```
+
+`guard()`, `capture()` and `flush()` are then importable on their own, and reach
+the registered client:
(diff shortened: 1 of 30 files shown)


arcjet-guard/README.md
```diff
@@ -427,6 +427,126 @@ cannot encode is reported locally as `AJ1017` and also travels with that event i
...
+test("refund captures an event", () => {
+  const arcjet = registerTestClient();
+  try {
+    refund("inv_1");
+
+    assert.equal(arcjet.captures[0]?.action, "refund.issued");
+  } finally {
+    arcjet.unregister();
+  }
+});
```
qw-in, line 556: Can we lead with `using` and then in like a 
\<details> tag show how it needs to be done before node@24?

arcjet-guard/src/registry.ts
```diff
@@ -0,0 +1,241 @@
...
+  const globalWithArcjet: GlobalWithArcjet = globalThis;
+  if (globalWithArcjet[symbolArcjetClient] !== undefined) {
+    throw new Error(
+      "An Arcjet client is already registered. Call unregisterArcjet() first — " +
+        "an earlier test probably left one behind.",
+    );
+  }
+
+  globalWithArcjet[symbolArcjetClient] = client;
+}
```
qw-in: This should be in a `/testing` directory so it doesn't accidentally get included in production bundles

arcjet-guard/README.md
```diff
@@ -427,6 +427,126 @@ cannot encode is reported locally as `AJ1017` and also travels with that event i
...
+
+`@arcjet/guard/testing` registers an in-memory client that records calls and
+talks to nothing:
+
+```ts
+import { registerTestClient } from "@arcjet/guard/testing";
+import { refund } from "./refund.ts";
+
+test("refund captures an event", () => {
+  const arcjet = registerTestClient();
```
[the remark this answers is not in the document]
qw-in: > _Claude Code, replying on @qw-in's behalf._

Correct, and fixed in 722da845.

`refund()` is `async` — the version in the `capture()` example above does `await issueRefund(id)` before it captures — so the unawaited call left the assertion racing a capture that happens a microtask later. The callback is now `async` and the call is awaited, in both this example and the `finally` variant, and in the matching `@example` blocks on `registerTestClient`.

The surrounding prose was misleading in the same direction. It said captures "are recorded synchronously, so assertions need no waiting", which reads as though no waiting were needed for the code under test either. It now says recording is synchronous *once the code under test reaches `capture()`*, and the example calls out the `await` explicitly.

arcjet-guard/README.md
qw-in: > _Claude Code, replying on @qw-in's behalf._

Done in 722da845 — `using` now leads, and the pre-Node-24 form is in a `<details>` block titled "Without `using` — Node.js 22, or no TypeScript compile step".

```ts
test("refund captures an event", async () => {
  using arcjet = registerTestClient();

  await refund("inv_1");

  assert.equal(arcjet.captures[0]?.action, "refund.issued");
});
```

The `details` block holds the `try`/`finally` form, the note that `unregister()` and `[Symbol.dispose]` are one function under two names, and the `skipLibCheck: false` / `esnext.disposable` caveat — that last one was previously in the main flow, where it made the whole feature read as encumbered rather than as a footnote for strict-lib consumers.

The `@example` blocks on `registerTestClient` are ordered to match, so the JSDoc hover leads with `using` too.

arcjet-guard/src/registry.ts
qw-in: > _Claude Code, replying on @qw-in's behalf._

Moved in 722da845.

`registerArcjetForTesting` now lives in `src/testing/register.ts`, and the test client moved from `src/testing.ts` to `src/testing/index.ts` — so the `./testing` subpath resolves to `dist/testing/index.js`, matching the `vercel-ai/v7` shape.

The reason it was in `registry.ts` at all was that it needs the slot primitives, which the public `registerArcjet` also needs. Rather than have `registry.ts` export something only tests use, those primitives moved to `src/registration-slot.ts` (`readRegistration`, `writeRegistration`, `hasRegistration`, the version check, the `isClient` guard). Both sides import from there, and neither module carries a test-only export.

Verified against the built output rather than assumed — walking the relative-import graph from each production entry:

```
node:    clean (17 modules, none under testing/)
fetch:   clean (15 modules, none under testing/)
bun:     clean (18 modules, none under testing/)
```

and `registerArcjetForTesting` no longer appears anywhere in `dist/registry.js`. It exists only in `dist/testing/`, which nothing in the production graph reaches, so it cannot be pulled into an application bundle even without tree-shaking.

The tests split along the same seams: `registration-slot.test.ts` for the version and hostile-slot cases, `testing/register.test.ts` for the throw-on-leak behaviour, `registry.test.ts` for the registry proper. The shared recorder doubles went to `test/_shared/registry-recorder.ts` — they can't live in `src/`, because the build globs `src/**/*.ts` minus `*.test.ts` and would emit a helper file into `dist/`, which is the same class of problem as this thread.

504 unit tests pass, plus the node/fetch/cloudflare runtime suites, and 504/504 on Node 22.21.0 in Docker.

arcjet-guard/src/testing/index.ts
```diff
@@ -0,0 +1,196 @@
...
+ * Captures are recorded as they happen, so assertions need no waiting. There is
+ * no transport and no queue, which is why unregistering is synchronous — there
+ * is nothing to drain.
+ *
+ * @example
+ * ```ts
+ * import { registerTestClient } from "@arcjet/guard/testing";
+ * import { refund } from "./refund.ts";
+ *
+ * test("refund captures an event", () => {
```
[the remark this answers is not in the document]
qw-in: Huh, I thought this was already resolved. Will do it in next round of changes

davidmytton (approved):

arcjet-rei (approved):
I see the pain points you're trying to address with this, and I'm willing to approve it in the interests of gathering further data about how useful or confusing this design turns out to be in practice, but I have reservations.

I think it probably introduces as many problems as it solves, because understanding the distinction between bounded / method invocation and global / function invocation for similar functionality requires developers to hold the operational model of their application and its configuration in their head, and the kind of applications where developers might need multiple client configurations are likely to be complicated enough to make holding this mental model most difficult. Also, this seems like something it would be incredibly easy for agents to mess up, again particularly in the kinds of larger or more complex applications where this functionality is likely to be most useful.

An alternative way to solve this problem (different parts of a large application need different client configuration, and threading state through to the right use site, potentially in library or module code that wants to optionally make use of functionality like Arcjet without having to pass it in or explicitly configure it) is Async Local Storage. If we leave responsibility to developers to thread state around, it's kind of a pain in the ass, but with some work we could wrap that up in helper functions that make it so that you can get something you can call `capture()` on without too much difficulty. (Check out ALS's low-level [`.bind()`](https://nodejs.org/api/async_context.html#static-method-asyncresourcebindfn-type-thisarg) and [`.runInAsyncContext()`](https://nodejs.org/api/async_context.html#asyncresourceruninasyncscopefn-thisarg-args) for the kinds of lower-level machinery we'd want to use.)

Like I said, I'm not opposed to trying this out, and I think this implementation is about as safe as it's possible for something like this to be (including the branding checks / use of Symbols, and using globalThis instead of a module's exports object. But before we commit to releasing it to users, we should investigate its ergonomics when used in the large.

PR #6208, opened 2026-08-07
feat(guard): add Eve support as @arcjet/guard/vercel-eve/v0

arcjet-rei:
## Summary

Adds support for [Eve](https://eve.dev), Vercel's agent framework, as a new subpath export: `@arcjet/guard/vercel-eve/v0`. No new package is published — the helpers ship inside the existing `@arcjet/guard` release, alongside the Vercel AI SDK helpers at `@arcjet/guard/vercel-ai/v7`.

```ts
import { launchArcjet, detectPromptInjection, tokenBucket } from "@arcjet/guard";

import {
  guardInbound,    // screen inbound text at a channel boundary
  guardApproval,   // gate a tool or connection call
  guardTool,       // wrap an authored tool's execute
  arcjetHooks,     // observe-only lifecycle capture
  eveAgentContext, // derive an Arcjet agent context from an Eve session
} from "@arcjet/guard/vercel-eve/v0";
```

`examples/eve-agent` is a runnable agent using all four helpers, and `arcjet-guard/skills/integrate-arcjet-guard-eve/` documents the integration for coding agents.

## Why this is not a port of `vercel-ai/v7`

Eve is filesystem-first. Tools are files, and there is no author-controlled call site where a request is assembled — so the places a guard can attach are different, and the helper set is different as a result.

| Helper | Eve surface | What it can do |
| --- | --- | --- |
| `guardInbound` | a channel handler | The only point that can decline a turn **before it starts**. Nothing later in the lifecycle can stop one, so inbound screening for prompt injection and PII belongs here. |
| `guardApproval` | `approval` on an authored tool, an OpenAPI connection, or an MCP connection | The primary enforcement surface, and the only one that reaches connection-derived tools — which may be *all* of an agent's tools, since an agent can have no authored tools at all. |
| `guardTool` | an authored tool's `execute` | The only surface that observes an execution outcome at the call site. |
| `arcjetHooks` | `defineHook` | Capture only. Eve's hooks return `void` and cannot reject a turn, so they are the wrong home for policy and the right home for observability. |

Correlation is read from Eve's session rather than generated, preferring the root session id so a subagent's decisions land on the conversation someone actually queries. A channel handler runs before the session exists, so `guardInbound` takes the identity the channel already has — a thread id, a delivery id — and `arcjetHooks` emits a record at session start tying that identity to the session id. The result is that one conversation reads as one sequence rather than a scatter of unrelated decisions.

Because `guardApproval` records the gate and `arcjetHooks` records the outcome, a denied call produces two events distinguishable by phase. Gates that pass without executing use a distinct `allowed` outcome, so "the tool ran" and "the tool was permitted to run" stay tellable apart.

## Why the segment is `v0`

The [subpath namespace convention](https://github.com/arcjet/arcjet/pull/8658) names paths `@arcjet/guard/<vendor-sdk>/v<major>`, and that ADR anticipated this namespace as `vercel-eve/v1`. It ships as `v0` instead: the version segment names *the SDK's* major, and Eve has not reached 1.0. Under semver each 0.x minor may break, so `v1` would name nothing. `v1` is added additively when Eve ships a stable release. This is now written into the convention itself, so the next pre-1.0 SDK does not have to re-litigate it.

As with `vercel-ai`, the version is explicit and there is no unversioned alias — `@arcjet/guard/vercel-eve` deliberately does not resolve.

## Eve is a type-only dependency

`@arcjet/guard` supports Node 22; Eve requires Node >= 24. The namespace therefore imports Eve for **types only** and never at runtime, which keeps guard installable and testable on Node 22 and means adding this namespace costs nothing to users who do not use Eve. Eve is declared as an optional peer. A test enforces the rule at the source level, and CI additionally runs the namespace's tests with Eve absent from `node_modules` so the guarantee is enforced rather than remembered.

One consequence is visible in the API: `guardTool` cannot call Eve's `defineTool`, so it copies property descriptors rather than spreading. `defineTool` stamps a non-enumerable key that a spread silently drops, which would break Eve's ability to match a tool result back to its definition.

## Notes for reviewers

Three things depart from what the [subpath namespace ADR](https://github.com/arcjet/arcjet/pull/8658) recorded, and are worth a look:

- **The `v0` segment**, above.
- **Two changes to the shared, SDK-agnostic layer** rather than none: `retryAfterSeconds` was extracted so both namespaces produce identical denial shapes, and `correlationIdProblem` is now exported from the agent context module. Both are internal; neither adds a public path.
- **Eve exposes no sandbox-command enforcement point.** An earlier, now-superseded design claimed one. `defineSandbox`'s `onSession` does receive session context and is a workable session-scoped guard site, but it runs once per session, not per command; `bootstrap` receives no session context; and `SandboxSession`'s `run`/`spawn` have no interception seam. This is recorded in the [Eve guard surfaces ADR](https://github.com/arcjet/arcjet/pull/8717) (still in draft), which also documents the Eve internals the helpers depend on.

The shared layer remains internal — `@arcjet/guard/agents` is still not in the export map. Eve is the second integration to exercise its shape without changing it, which is the evidence that promoting it to the root export was waiting on; that promotion is deliberately not in this PR.

## Out of scope

- Promoting the SDK-agnostic layer to the root `@arcjet/guard` export.
- A separate Eve extension package.
- `defineDynamic` tools. Eve's compiler hoists a dynamic tool's inline `execute` for durable replay and cannot see through a wrapper, so those are gated with `approval` instead.


arcjet-rei:
## Verification

Run from the repository root unless noted.

```bash
npm run build     --workspace @arcjet/guard
npm run typecheck --workspace @arcjet/guard   # runs both tsconfig.json and tsconfig.lint.json
npm run lint      --workspace @arcjet/guard
npm run test-unit --workspace @arcjet/guard
```

Runtime suites — `node` and `fetch` were each run under Node 22, 24 and 26 locally, matching the CI matrix:

```bash
npm run test-runtime-node       --workspace @arcjet/guard
npm run test-runtime-fetch      --workspace @arcjet/guard
npm run test-runtime-bun        --workspace @arcjet/guard
npm run test-runtime-deno       --workspace @arcjet/guard
npm run test-runtime-cloudflare --workspace @arcjet/guard
```

The example:

```bash
cd examples/eve-agent && npm ci && npm run build && npm run typecheck
```

### The type-only guarantee

The claim that the namespace never imports Eve at runtime is checked two ways, deliberately. A source scan asserts every `eve` import is written `import type` — that catches the mistake where it is made. The runtime check catches the consequence, including a build step that re-emits a type import as a value import:

```bash
mv node_modules/eve /tmp/eve-stash
node --test 'arcjet-guard/src/vercel-eve/**/*.test.ts'
mv /tmp/eve-stash node_modules/eve
```

The whole namespace suite passes with Eve absent. The same steps run in CI on every leg of the Node matrix.

Note that `npm run typecheck` **fails** with Eve absent, with unresolved `eve` module errors. That is expected and is not a defect — types are a build-time dependency, which is why Eve is a devDependency as well as an optional peer. The guarantee is about the runtime graph.

## Still to verify manually

Two things could not be checked in this environment and need someone with credentials:

1. **An end-to-end correlated sequence against a real Arcjet site.** One conversation should produce, under a single correlation id: the inbound screening decision, a session-start record carrying both that id and the session id, then one gate decision and one result event per tool call. The negative paths matter as much as the happy one — a screening denial should start **no** session at all, and a gate denial should produce two events distinguishable by phase, with the model explaining the denial rather than retrying in a loop. Needs a valid `ARCJET_KEY`.

2. **The coding-agent skill.** `arcjet-guard/skills/integrate-arcjet-guard-eve/` has not been exercised by a fresh agent session working from it alone.

## Notes

- Two devDependency bumps ride along in `arcjet-guard/package.json` and are unrelated to Eve: `ai` 7.0.36 → 7.0.38 and `@ai-sdk/provider-utils` 5.0.12 → 5.0.13. They align the installed AI SDK with the version `ai@7.0.38` expects.
- `arcjet-guard` is excluded from the root formatter (`.oxfmtrc.json`), so its `format:check` is not enforced in CI and had pre-existing failures before this branch. Files added or changed here pass it; the pre-existing ones were deliberately left alone rather than reformatted, to keep this diff reviewable.
- The design and implementation plans were removed in the final commit, following the convention the two previous namespace features used. They are still readable at `e7281d086d1ff88e94025533420c950ac7bcae49` if anyone wants the working notes.

arcjet-guard/package.json
```diff
@@ -95,24 +99,29 @@
...
     "miniflare": "4.20260708.1",
     "oxlint-tsgolint": "0.24.0",
     "tsdown": "0.22.7",
     "typescript": "7.0.2"
   },
   "peerDependencies": {
     "@ai-sdk/provider-utils": ">=5 <6",
-    "ai": ">=7 <8"
+    "ai": ">=7 <8",
+    "eve": ">=0.30 <1"
```
qw-in: What happens if I want to use @arcjet/guard (just generally - not for `ai`) and I'm outside this range? Does the install fail?

Genuine question I'm not sure what the behavior is in different package managers

arcjet-guard/package.json
```diff
@@ -95,24 +99,29 @@
     "@connectrpc/connect-web": "2.1.2"
   },
   "devDependencies": {
-    "@ai-sdk/provider-utils": "5.0.12",
+    "@ai-sdk/provider-utils": "5.0.13",
     "@types/node": "22.20.1",
-    "ai": "7.0.36",
+    "ai": "7.0.38",
+    "eve": "0.31.0",
```
qw-in, line 109: No conflicts yet but we'll need a different strategy for devDependencies soon I'll bet!

arcjet-guard/src/agents/context.ts
```diff
@@ -14,7 +14,7 @@ const CORRELATION_ID_RE: RegExp = /^[ -~]{1,256}$/;
  * The `typeof` check comes first because `RegExp.test()` coerces its argument,
  * so a number would otherwise satisfy the pattern.
  */
-function correlationIdProblem(value: unknown): string | undefined {
+export function correlationIdProblem(value: unknown): string | undefined {
```
qw-in, line 20: Should we tag `@internal` on these sorts of exports?

qw-in (approved):
Looks good to me!

arcjet-guard/package.json
arcjet-rei: If Eve's not present in `package.json`, you're fine. It's an `optionalPeerDependency`, so its absence has no effect. If you **do** have Eve installed and it's outside the range, npm will balk on the install. However, this question led me down a bit of a rabbit hole and I've realized this range may be overly conservative, so I've got Claude figuring out the minimum version of Eve our helpers are compatible with, and I'll broaden the range here to match that.

arcjet-rei: Broadened to `eve >= 0.25.1` in https://github.com/arcjet/arcjet-js/pull/6208/commits/bc02cf35e58184c581400ceda8a8e942de783ce6, because that's the version that includes some of the needed machinery. It is an ancient build of Eve that is a whole two weeks old. That commit also adds a CI test to ensure that whatever the floor is, it still works with our changes, so if we introduce a backwars-incompatible change to our Eve support, we know about it.

arcjet-guard/package.json
arcjet-rei: There _is_ a conflict, or rather a coupling, that we probably need to make more explicit, and that did have an impact on the development of this PR. Given builds of `eve` have explicit/fixed dependencies on particular builds of `ai`, and the failure mode isn't that the install fails, it's that the type checking fails because the types are coming from different instances of the `ai` package when they're not kept in lockstep. This is an aspect of TypeScript I find particularly unlovely, but what are you gonna do. In any case, we could add lint checks that would catch this problem as a follow-up.

arcjet-guard/src/agents/context.ts
arcjet-rei: Done in 5000448e (and with a followup in 3a279b01 for the other instances of "these sorts" of exports that weren't already tagged, so now `shouldWarn`, `captureEvent`, `correlationIdProblem`, `retryAfterSeconds`, `arcjetProtectedTool`, and `ulid` are also marked `@internal`, verified by actual reachability rather than just looking for re-export).

PR #6226, opened 2026-08-14
feat(guard): add Mastra support as @arcjet/guard/mastra/v1

davidmytton:
Adds a Mastra integration for `@arcjet/guard` under the versioned subpath `@arcjet/guard/mastra/v1`, alongside the existing `vercel-ai/v7` and `vercel-eve/v0` namespaces.

## What's new

- **`guardTool`** — wraps a Mastra `createTool({ execute })`. DENY returns a structured `ArcjetDenialResult` to the model; the tool never runs.
- **`guardProcessor`** — a Mastra `Processor` for `inputProcessors` / `outputProcessors`. On DENY calls `abort()` to raise a tripwire; `processInputStep` also screens later agentic steps so tool continuations cannot skip the inbound gate.
- **`guardHooks`** — `beforeToolCall` / `afterToolCall` for MCP / workspace / toolset tools that were not authored through `guardTool`.
- **`mastraAgentContext`** — reads `MASTRA_THREAD_ID_KEY`, resource, then `workflow.runId` for correlation. Never mints an id.

Mastra `requireApproval` is human HITL, so there is no `guardApproval`; channels already hit `processInput`, so there is no `guardInbound`.

## Packaging

- `@mastra/core@^1` is added as an **optional** peer (plus devDependency for tests). Nothing in the mastra namespace value-imports `@mastra/core` — enforced by a static scan and by a new CI job that runs the namespace's tests with `@mastra/core` deleted from `node_modules`.
- Export map adds `./mastra/v1` only; there is deliberately no unversioned `./mastra` alias or wildcard.
- Renovate is configured to keep `@mastra/core` on 1.x and never automerge majors.

## Docs and skill

- README gains a Mastra usage section, a split peer-install block, and a fail-closed table row.
- Adds `skills/integrate-arcjet-guard-mastra/SKILL.md` for Claude Code.
- CONTRIBUTING and a new root `AGENTS.md` direct new examples to `arcjet/examples` in https://github.com/arcjet/examples/pull/193/commits/d17cc19ba2fa405a9ef114ad92e8c80a7dc9ca9d (per #6217), not `examples/` in this repo.

## Tests

New tests cover assignability to Mastra slots, context derivation, denial shape, the shared gate, the processor (input, step, output, role-spoofing, extraction edge cases), the tool wrapper (double-wrap rejection, denial shape, fail-closed and fail-open, brand preservation), hooks, the index barrel and export map, the optional-peer configuration, and a static scan proving every `@mastra/core` import is type-only.


.github/workflows/guard.yml
```diff
@@ -108,6 +108,21 @@ jobs:
           node --test 'arcjet-guard/src/vercel-eve/**/*.test.ts'
         working-directory: ${{ github.workspace }}
 
+      # The mastra namespace imports @mastra/core for types only. A static scan
+      # proves the imports are written `import type`; this proves the
+      # consequence, catching a build step that re-emits one as a value import.
+      #
+      # Runs before the eve-absent step because that one deletes a different
+      # optional peer. Typecheck legitimately fails without @mastra/core.
+      - name: Unit tests with mastra absent
```
qw-in, line 117: Thinking we'll want a general pattern for this soon-ish. JS makes it so hard!

Makes sense for now

qw-in (approved):
Happy to get it tested!

.github/workflows/guard.yml
davidmytton: Yeah, just copied the existing pattern for this one.

PR #6248, opened 2026-08-24
feat(guard): add LangChain createAgent support as @arcjet/guard/langchain/v1

davidmytton:
Adds a LangChain JS namespace for `createAgent` + `createMiddleware({ wrapToolCall })`, exporting `guardTool`, `guardMiddleware`, and `langchainContext`, plus a skill and README section. This is not LangGraph's Graph API (`StateGraph` + `ToolNode`) — that is `@arcjet/guard/langgraph/v1` — and not `vercel-ai/v7`.

`langchain` is a new optional peer at `>=1.2.0 <2`. `@langchain/core` keeps the `>=1 <2` range `langgraph/v1` already shipped — the two floors differ deliberately, see below. Nothing in the namespace imports either peer at runtime except one deny-path dynamic load of `ToolMessage`, so the module still loads with the peers absent — enforced by a static type-only scan plus a peers-absent run of the whole `src/langchain` suite. No new `@langchain/langgraph` peer, and no unversioned `@arcjet/guard/langchain` export.

## Two denial envelopes, deliberately not collapsed

This is the one thing to review closely. The two helpers deny differently because only one of them passes through `baseHandler`.

`guardTool` returns the plain `ArcjetDenialResult`. It does not throw and does not fabricate a `ToolMessage`, because `createAgent`'s `baseHandler` already wraps a non-`ToolMessage` result in a real one with no `status` field, i.e. success (`ToolNode.js:244-248`). Same envelope as `langgraph/v1`.

```ts
import { guardTool } from "@arcjet/guard/langchain/v1";

const lookupOrder = guardTool(arcjet, lookupOrderTool, {
  action: "order.looked-up",
  rules: (input) => [limit({ key: input.orderNumber, requested: 1 })],
});
```

`guardMiddleware`'s `wrapToolCall` must return a real `ToolMessage`, carrying the payload as `content`. Its return value is **not** passed through `baseHandler`, so a bare object fails `ToolMessage.isInstance` and crashes the messages reducer. It does not set `status: "error"`, and it does not throw: `#handleError` re-raises middleware errors by default ("if error is from middleware and `handleToolErrors` is not true, bubble up", `ToolNode.js:275-282` and `#handleError:147-150`), which would drop `arcjetDenied` out of `invoke` entirely.

```ts
import { guardMiddleware } from "@arcjet/guard/langchain/v1";

const agent = createAgent({
  model,
  tools: [lookupOrder, ...mcpTools],
  middleware: [guardMiddleware(arcjet, { sessionId: conversationId })],
});
```

Policy sits on `wrapToolCall` only. `humanInTheLoopMiddleware` / `interrupt()` is human-in-the-loop, not a policy gate, so there is no `guardApproval` and nothing denies in `afterModel`. Already-branded tools are skipped when `request.tool` resolves, so `guardTool` and `guardMiddleware` do not double-call Guard on the same tool; MCP, unwrapped, and runtime-discovered tools (where `request.tool` is undefined) are still gated.

## Correlation and inbound

`langchainContext` reads `configurable.thread_id`, then caller-owned `sessionId` / `conversationId`. It never mints an id, never reads `traceId`, and never treats interrupt/resume as correlation. `wrapToolCall` only sees `runtime.configurable.thread_id` as of langchain 1.2.34, which `ToolNode` populates as `configurable: lgConfig?.configurable` (`ToolNode.js:180-183`).

There is no inbound helper, because `createAgent` has no first-class inbound channel. Screen with the core client before `agent.invoke` and act on both outcomes:

```ts
const decision = await arcjet.guard({
  label: "message.received",
  rules: [detectPromptInjection()(userText)],
  ...langchainContext({ configurable: { thread_id: conversationId } }),
});
if (decision.conclusion === "DENY" || decision.hasFailedOpen()) {
  throw new Error("message blocked");
}
```

`wrapModelCall` / `beforeModel` / `afterModel` intercept the model call, not user text, so they are not this gate.

## Why the two peer floors differ

`langchain` needs `>=1.2.0 <2` on its own terms: `wrapToolCall` only sees `runtime.configurable.thread_id` as of 1.2.34. It is a brand-new peer, so no existing consumer is affected.

`@langchain/core` deliberately stays at `>=1 <2`. It is a *shared* peer, so tightening it would have constrained `langgraph/v1` consumers who never install `langchain` — and `@langchain/langgraph@1.4.10` peer-requires only `@langchain/core: ^1.1.48`, so a consumer on core 1.1.x would have been pushed out for no reason. The floor would also have bought this namespace nothing, since `langchain@1.5.10` peer-requires `@langchain/core: ^1.2.9`, which is strictly stronger and is what actually binds for anyone using this namespace.

## Out of scope

Server-side provider tools and headless `.implement()` tools. Docs land at `/guards/langchain-js/`; `/guards/langchain/` is the live Python page and is untouched. The `langchain-agent` example is a follow-up on [`arcjet/examples`](https://github.com/arcjet/examples), not this repo.

## Verification

| check | result |
| --- | --- |
| `npm run lint`, `npm run typecheck` | pass |
| `npm run format:check` (root) | pass, 298 files |
| `npm run test-unit` (whole package) | 1336 pass, 0 fail |
| `node scripts/test-peers-absent.mjs` (all 7 namespaces) | pass |

The end-to-end tests drive a real `createAgent` with `FakeToolCallingModel`, asserting that a DENY never reaches the handler, arrives as a completed `ToolMessage` rather than an interrupt, has `status !== "error"`, and that a `guardTool`-branded tool is guarded exactly once.

Review findings are folded in as separate commits: `langchainContext` was returning the first `configurable`-shaped object it found, so an empty `configurable` alongside a real id elsewhere sent the decision out uncorrelated; `guardMiddleware`'s policy-factory and `onDeny` failure paths were untested; the README's `onGuardError` and denial-envelope tables had no LangChain row, the latter describing only LangGraph's bare-object return, which is the crash case if applied to `wrapToolCall`; a missing `tool_call_id` now warns rather than silently emitting a blank one; and the `@langchain/core` peer narrowing described above was reverted.

The one red check, `StepSecurity Harden-Runner`, is not caused by this change and is not a required check — details in a comment below.

<div><a href="https://cursor.com/agents/bc-3d83ad5e-bc33-4159-82f6-5c241ce1f9b3?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-3d83ad5e-bc33-4159-82f6-5c241ce1f9b3&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/arcjet-guard/skills/integrate-arcjet-guard-langchain/SKILL.md
+++ b/arcjet-guard/skills/integrate-arcjet-guard-langchain/SKILL.md
@@ -1,0 +1,296 @@
+---
+name: integrate-arcjet-guard-langchain
+description: Integrate Arcjet security into a LangChain JS createAgent using @arcjet/guard — wrap tool() / StructuredTool, put guardMiddleware on createAgent({ middleware }) for MCP / unwrapped tools, and read configurable.thread_id for correlation. Use when asked to add Arcjet to langchain createAgent, rate limit its tools, screen inbound messages, or block prompt injection / PII. This is LangChain JS, not the Python page.
+license: Apache-2.0
+compatibility: Requires the target app to use LangChain JS (langchain >=1.2.0 <2 and @langchain/core >=1.2.0 <2) on Node.js >= 22. This is createAgent + createMiddleware({ wrapToolCall }), not LangGraph StateGraph / ToolNode. wrapToolCall only sees runtime.configurable.thread_id as of langchain 1.2.34.
+metadata:
+  author: arcjet
+---
+
+# Integrate Arcjet Guard into a LangChain JS createAgent
+
+`@arcjet/guard`'s LangChain v1 namespace wraps the agent's existing Arcjet
+client. It never talks to the Arcjet API itself. Three surfaces, one
+decision rule:
+
+- **An authored tool** (`tool()` / `StructuredTool`) → `guardTool()`.
+  DENY returns a plain `ArcjetDenialResult`. Do not throw. Do not
+  fabricate a `ToolMessage`. `createAgent`'s `baseHandler` wraps a
+  non-ToolMessage in a success `ToolMessage`.
+- **MCP / unwrapped / runtime-discovered tools** → `guardMiddleware()`.
+  A `createAgent({ middleware })` middleware whose `wrapToolCall` is
+  the invoke()-wide gate. It denies by returning a **real**
+  `ToolMessage` (`content` = JSON of the payload, `tool_call_id` =
+  `request.toolCall.id`, `name` = `request.toolCall.name`) without
+  calling `handler`. Already-branded tools are skipped when
+  `request.tool` can be looked up. Do not set `status: "error"`. Do
+  not throw (throws bubble and drop `arcjetDenied`).
+- **Correlation** → `langchainContext()` reads
+  `configurable.thread_id` (what wrapToolCall sees on
+  `runtime.configurable` as of langchain 1.2.34), then caller-owned
+  `sessionId` / `conversationId`. It never mints a new id. It never
+  reads `traceId`. It never treats `interrupt` / resume as
+  correlation.
+
+This namespace is LangChain JS **`createAgent` + `wrapToolCall`**. Not
+LangGraph Graph API (`StateGraph` + `ToolNode`) — that is
+`@arcjet/guard/langgraph/v1`. Not `vercel-ai/v7`. Server-side provider
+tools and headless `.implement()` tools are out of scope. Do not also
+wrap the same tool with `@arcjet/guard/langgraph/v1` or
+`@arcjet/guard/vercel-ai/v7`.
+
+Docs live at
+[docs.arcjet.com/guards/langchain-js/](https://docs.arcjet.com/guards/langchain-js/).
+Do **not** use `/guards/langchain/` — that is the live Python page.
+
+## Screen inbound before `agent.invoke` — there is no inbound hook. SDK middleware that is not `wrapToolCall` is not Guard.
+
+There is no first-class inbound channel, so there is no
+`guardInbound`. Put prompt-injection (and other inbound rules) in the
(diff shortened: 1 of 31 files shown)


arcjet-guard/skills/integrate-arcjet-guard-langchain/SKILL.md
```diff
@@ -0,0 +1,296 @@
+---
+name: integrate-arcjet-guard-langchain
+description: Integrate Arcjet security into a LangChain JS createAgent using @arcjet/guard — wrap tool() / StructuredTool, put guardMiddleware on createAgent({ middleware }) for MCP / unwrapped tools, and read configurable.thread_id for correlation. Use when asked to add Arcjet to langchain createAgent, rate limit its tools, screen inbound messages, or block prompt injection / PII. This is LangChain JS, not the Python page.
```
arcjet-rei:
```suggestion
description: Integrate Arcjet security into a LangChain JS createAgent using @arcjet/guard — wrap tool() / StructuredTool, put guardMiddleware on createAgent({ middleware }) for MCP / unwrapped tools, and read configurable.thread_id for correlation. Use when asked to add Arcjet to langchain createAgent, rate limit its tools, screen inbound messages, or block prompt injection / PII. This is LangChain JS, not the Python SDK.
```

PR #6256, opened 2026-08-28
fix(transport): port HTTP/2 PING keep-alive and connection recycling (#6255)

### LINKED ISSUE #6255: Port the HTTP/2 connection recycling from arcjet-guard to @arcjet/transport (#6137 follow-up)
## Summary

`arcjet-guard` got HTTP/2 connection recycling and PING keep-alive in 742cb4b (#6137, "detect and recycle silently dropped HTTP/2 connections"). In that thread a maintainer wrote:

> I'll put up a second PR with a similar change to `@arcjet/transport`.

That port doesn't appear to have landed. `transport/src/index.ts` at 1.11.0 still builds the session with only an idle timeout:

```js
const sessionManager = new Http2SessionManager(baseUrl, { idleConnectionTimeoutMs: 340 * 1e3 }, ...);
if (!("Deno" in globalThis)) sessionManager.connect();
```

No `pingIntervalMs`, no `pingIdleConnection`. With pings defaulting to `Infinity`, `requiresVerify()` is never true, so the session is never probed before reuse — it sits fully idle for the whole 340 s window.

It also still carries the premise #6137 explicitly corrected. The commit replaced "AWS Global Accelerator doesn't support PING" with the observation that GA ignores *dataless TCP keepalives* while HTTP/2 PING frames are stream data and do reset the idle timer. `transport/` still has the original reasoning.

## Why this matters for `@arcjet/next` on serverless

We run `@arcjet/next` on Vercel (Node 22, `iad1`), one module-scope client per the best-practices docs. Vercel freezes function instances between requests, so the client-side 340 s idle timer often doesn't fire before the peer decides the connection is dead and sends `GOAWAY`. The session is then destroyed with an error that reaches us as an uncaught exception:

```
ConnectError: [canceled] received GOAWAY without any open streams
    at Http2Session.callbackTrampoline (node:internal/async_hooks:130:17)
    at Http2Session.onGoawayData (node:internal/http2/core:748:11)
    at ClientHttp2Session.emit (node:events:519:28)
    at .../@connectrpc/connect-node/dist/esm/http2-session-manager.js:469:26
```

64 occurrences over five weeks, `level: fatal`, `handled: no`, across 15 releases and still ongoing. Notably **zero user impact** — no request fails, because the error surfaces asynchronously on an instance that is idle. The cost is noise at fatal severity, which crowds out real alerts.

Two independent time samples, instance start → error: 36 min and 11 min. Both far beyond the 340 s idle window.

## Scope note

The version numbers make this easy to misread. arcjet-js versions in lockstep, so `@arcjet/next`, `@arcjet/transport` and `@arcjet/guard` are all 1.11.0, and guidance phrased as "fixed in 1.9.0, upgrade to 1.9.0 or later" reads as universal. We were already on 1.10.0 and still saw the issue, because the fix only ever touched `arcjet-guard/src/`.

## The ask

Port `arcjet-guard/src/transport-recycle.ts` and the `transport-http2.ts` PING configuration to `transport/`, and update the Global Accelerator comment to match the corrected reasoning in #6137.

## Related

There is also an unfixed uncaught-exception hazard on the connect-es side — `onGoaway` destroys the session with an error but doesn't attach the swallow listener that `abort()` already uses for the same hazard: connectrpc/connect-es#1768, with a fix in connectrpc/connect-es#1769. Either change alone would stop the crash; recycling the connection on the Arcjet side addresses the cause rather than the symptom.

Happy to test a patch against our production traffic if that's useful.

davidmytton:
## Summary

Ports the HTTP/2 connection recovery from `#6137` (`arcjet-guard`) into `@arcjet/transport`, and centralizes the factory so Guard and the main SDK share one implementation.

Fixes `#6255` (async `GOAWAY` / silently dead HTTP/2 sessions for `@arcjet/next` and other adapters that use `@arcjet/transport`).

## Changes

- Add PING keep-alive (`pingIntervalMs: 55s`, `pingTimeoutMs: 5s`, `pingIdleConnection: true`) and deadline-based connection recycling to the Node HTTP/2 path in `@arcjet/transport`.
- Correct the Global Accelerator comment (HTTP/2 PING frames *do* reset the idle timer; dataless TCP keepalives do not).
- Export `@arcjet/transport/http2` (unconditional — not subject to the package `bun`/`deno` conditions) so `@arcjet/guard` can reuse the same factory on Node and Bun.
- Point Guard’s Node/Bun direct path at that shared factory; remove the duplicated recycle module from `arcjet-guard`.

## Behavior note: HTTP/2-through-CONNECT also gets PINGs

`proxyHttpVersion: "2"` already shared the same `Http2SessionManager` factory as the direct path (and the pre-existing comment said pings/idle timeout behave identically for both). Previously that meant “no PINGs on either path.” After this PR both paths get the PING keep-alive + recycling — including periodic HTTP/2 PING frames on the tunneled session. That is intentional: a tunneled connection is end-to-end HTTP/2 with the origin (the proxy only forwards bytes), and RFC 7540 requires PING support. Default proxying remains HTTP/1.1.

## Why not fully unify `createTransport`?

- **Bun diverges on purpose:** `@arcjet/transport`’s `bun` entry uses fetch/HTTP/1.1; Guard’s Bun entry still wants Node HTTP/2 for the direct path (Bun’s `fetch` has no HTTP/2).
- **Proxy detection stays duplicated:** Guard’s `detect-proxy` copy is intentionally edge-safe with no `@arcjet/logger` / `@arcjet/env` imports (existing drift test).

## Test plan

- [x] `@arcjet/transport` unit tests — 54 pass (includes moved recycle suite)
- [x] `@arcjet/guard` unit tests — 1424 pass
- [x] `@arcjet/guard` Node runtime — 35 pass (includes production `createHttp2Transport` factory test)

<div><a href="https://cursor.com/agents/bc-01a0494c-bd53-7e6e-84b4-07d66fdc8b33?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-01a0494c-bd53-7e6e-84b4-07d66fdc8b33&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/arcjet-guard/package.json
+++ b/arcjet-guard/package.json
@@ -127,6 +127,7 @@
   "dependencies": {
     "@arcjet/analyze": "1.11.0",
     "@arcjet/logger": "1.11.0",
+    "@arcjet/transport": "1.11.0",
     "@bufbuild/protobuf": "2.14.0",
     "@connectrpc/connect": "2.1.2",
     "@connectrpc/connect-node": "2.1.2",
--- a/arcjet-guard/src/transport-http2.ts
+++ b/arcjet-guard/src/transport-http2.ts
@@ -2,83 +2,18 @@
  * Direct HTTP/2 transport factory shared by the `@arcjet/guard` Node and Bun
  * entry points.
  *
- * Both Node and Bun talk to the Arcjet API over HTTP/2 via
- * `@connectrpc/connect-node` (Bun implements `node:http2`, but its `fetch` does
- * not support HTTP/2 — {@link https://github.com/oven-sh/bun/issues/7194}). The
- * proxy strategy differs between the two runtimes, so each entry point handles
- * proxying itself and reuses this for the direct, no-proxy case.
+ * Re-exports the shared factory from `@arcjet/transport/http2` so Guard and the
+ * main SDK stay on the same PING keep-alive and deadline-based recycling
+ * behavior. The unconditional `/http2` subpath is required on Bun: the
+ * `@arcjet/transport` package `"bun"` condition resolves to fetch/HTTP/1.1,
+ * while Guard's Bun entry still wants Node HTTP/2 for the direct (no-proxy)
+ * path — Bun implements `node:http2`, but its `fetch` does not support HTTP/2
+ * ({@link https://github.com/oven-sh/bun/issues/7194}).
  *
  * @packageDocumentation
  */
 
-import type { Transport } from "@connectrpc/connect";
-import { createConnectTransport, Http2SessionManager } from "@connectrpc/connect-node";
-
-import { withConnectionRecycling } from "./transport-recycle.ts";
-
-/**
- * A direct HTTP/2 transport plus the session manager that owns its connection.
- *
- * The session manager is exposed so callers (and tests) can tear the
- * connection down deterministically.
- */
-export interface Http2TransportHandle {
-  transport: Transport;
-  sessionManager: Http2SessionManager;
-}
-
-/**
- * Create a direct HTTP/2 Connect transport, optimistically pre-connecting.
- *
- * The session is pre-connected so the first `.guard()` call doesn't pay the
- * full TCP + TLS setup cost. PING keep-alive and deadline-based connection
- * recycling detect a silently dropped connection (an intermediary expiring an
- * idle flow without notifying either end) and replace it, instead of letting a
- * dead session fail every call until the process restarts.
- *
- * @param baseUrl Base URL for the Arcjet API.
- * @returns The transport and its session manager.
- */
-export function createHttp2Transport(baseUrl: string): Http2TransportHandle {
-  const sessionManager = new Http2SessionManager(baseUrl, {
-    // Detect and survive silently dropped connections:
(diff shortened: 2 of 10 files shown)


qw-in (approved):
> Bun diverges on purpose

Bun has some upcoming [http2/3 changes that are still experimental](https://bun.com/blog/bun-v1.4#http-2-http-3-in-fetch-experimental)

---

Makes sense, its been tested in protect code!

PR #6272, opened 2026-09-07
feat(guard): surface a policy expression rule result

davidmytton:
Regenerates the Guard client from the current protos (`make -C proto build-guard-js` in the `arcjet` repo) and handles the result variant they add.

## What changed upstream

`decide.proto` gained `ResultPolicyExpression`, the result of a remote-policy rule decided by the policy's expression language. It is the primary variant a policy v2 produces — every Rego rule reports through it. The proto change is purely additive: no field was removed or renumbered, and every existing rule type keeps its number, so the wire contract with released SDKs is unchanged.

## Why this SDK needs a change

`policyResultFromProto` did not match the new variant, which left `result` unassigned and stopped the switch compiling — so this one fails loudly rather than silently.

Left unhandled it would be worse than a build failure. Enforcement follows the aggregate decision conclusion, not per-rule detail, so a denied request is still denied — but a denying rule would read as `type: "UNKNOWN", conclusion: "ALLOW"` in `policyResults`, the wrong conclusion for that rule to anything inspecting it. (The Python and Go SDKs have no compiler check here and did fail silently; those PRs are the counterparts.)

## This change

- handles `policyExpression`, taking the conclusion from the message
- adds `RuleResultPolicyExpression`, carrying the conclusion and nothing more; which conditions fired is a property of the compiled artifact named by `policyRevision`, not something the result restates

The switch keeps `case undefined` and no `default`. The exhaustiveness lint requires the former and `no-useless-switch-case` forbids pairing them, which together mean a variant added to the proto fails this build rather than silently reading as `UNKNOWN` — the behaviour that made this worth catching.

Also drops the generated `guard/policy/v1` client: that proto no longer exists upstream, so it is no longer generated, and nothing in this package imported it.

Verified end to end against a real policy: Rego compiled by the control plane's own compiler, published to a local decide-service worker, returning `POLICY_EXPRESSION` results with the correct per-rule conclusions.


davidmytton:
Addressing the review's test focus area: `dcce97b` adds a `policy expression results` suite to `src/convert.test.ts`, covering DENY and ALLOW and asserting the warnings array, as asked.

The compiler already refuses an unhandled variant here — that is why this SDK failed loudly while the Go and Python ones fell through silently — but what a caller reads is the conclusion rather than the switch, so the tests assert that instead of relying on the build. One asserts the result type as well, so the ALLOW case cannot pass by reading as `UNKNOWN`.

On the other focus areas, since they came up on all three SDK PRs:

- **`pr.result.value.conclusion` is the right and only source.** `ResultPolicyExpression` carries nothing else; which conditions fired is a property of the compiled artifact named by `policyRevision`. An older or malformed response has no `policyExpression` case set, so it takes `case undefined` and reads as `UNKNOWN` — unchanged.
- **Adding to the `RuleResult` union is the point, and yes it affects exhaustive switches.** A caller discriminating on `result.type` gains a `POLICY_EXPRESSION` arm to handle. Without it, a denying Rego rule reads as `type: "UNKNOWN", conclusion: "ALLOW"` — the wrong conclusion for that rule.
- **Nothing imported the removed `guard/policy/v1` client.** I grepped this package before deleting it. It is reachable through published package paths, though, so if v1 generated modules are considered public surface, that is a maintainer call rather than something the diff settles.
- **The generated v2 bindings are unedited `buf generate` output** from `make -C proto build-guard-js` in the `arcjet` repo, against the protos as merged in arcjet/arcjet#9172.

qw-in (approved):

PR #6277, opened 2026-09-09
feat(guard): add actor and inputs to every adapter policy

davidmytton:
Every JS vendor wrapper now accepts optional `actor` and `inputs` so a remote Guard policy that declares those names can evaluate.

Values can be static or resolved from that adapter’s native call (parsed input plus the framework’s trusted runtime or context, the same idea as Vercel AI’s `(input, ctx)`). Build each input with `policyInput`.

Applies to Vercel AI, Eve, Mastra, LangChain, LangGraph, Genkit, OpenAI Agents, Strands, TanStack AI, Google ADK, Claude Agent SDK, and Claude Managed Agents.

<div><a href="https://cursor.com/agents/bc-fab4e779-692a-4b58-a53e-957f7cd52ad2?cursor_ref=pr_footer&cursor_cta=open_in_web"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-fab4e779-692a-4b58-a53e-957f7cd52ad2&cursor_ref=pr_footer&cursor_cta=open_in_cursor"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>

### DIFF
--- a/arcjet-guard/src/genkit/v1/guard-tool.ts
+++ b/arcjet-guard/src/genkit/v1/guard-tool.ts
@@ -484,7 +484,7 @@ async function runGuardedTool<TInput>(
   let rules: RuleWithInput[] | undefined;
   let policyMetadata: ArcjetMetadata | undefined;
   let remote: Awaited<ReturnType<typeof resolveActorInputs>> = {};
-  let callOptions = options;
+  let callOptions: unknown;
   try {
     // oxlint-disable-next-line typescript/no-unsafe-type-assertion -- args are the tool's parsed input; policy factories are typed against it
     const typedArgs = args as TInput;
--- a/arcjet-guard/src/genkit/v1/guard-tool.ts
+++ b/arcjet-guard/src/genkit/v1/guard-tool.ts
@@ -66,17 +66,16 @@ export interface GuardToolPolicy<TInput> {
    */
   rules?: RuleWithInput[] | ((input: TInput) => RuleWithInput[]);
   /**
-   * Trusted actor identity, or a resolver over the tool input. Derive it from
-   * authenticated server-side context; never trust a model-produced tool input
-   * as the actor identity — a policy can be conditioned on the actor, so a
-   * model-controlled value could escape scope.
+   * Trusted actor identity, or a resolver `(input, options) => …` matching
+   * a Genkit `ToolAction` call. Derive it from `options.context`; never
+   * trust a model-produced tool input as the actor identity.
    */
-  actor?: ActorResolver<TInput>;
+  actor?: ActorResolver<[TInput, unknown?]>;
   /**
-   * Typed remote-policy inputs, or a resolver over the tool input. Build each
-   * value with {@link policyInput}.
+   * Typed remote-policy inputs, or a resolver `(input, options) => …`. Build
+   * each value with {@link policyInput}.
    */
-  inputs?: InputsResolver<TInput>;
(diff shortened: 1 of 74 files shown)


arcjet-guard/src/genkit/v1/guard-tool.ts
```diff
@@ -460,13 +481,15 @@ function runGuardedTool<TInput>(
...
   let policyMetadata: ArcjetMetadata | undefined;
+  let remote: Awaited<ReturnType<typeof resolveActorInputs>> = {};
   try {
     // oxlint-disable-next-line typescript/no-unsafe-type-assertion -- args are the tool's parsed input; policy factories are typed against it
     const typedArgs = args as TInput;
     sessionId = resolveSessionId(policy, typedArgs);
     rules = typeof policy.rules === "function" ? policy.rules(typedArgs) : policy.rules;
     policyMetadata =
       typeof policy.metadata === "function" ? policy.metadata(typedArgs) : policy.metadata;
+    remote = await resolveActorInputs(policy, typedArgs, options);
```
davidmytton: [P2] Make the active Genkit context available to the resolvers

During a real `ai.generate({ context: { userId: "authenticated-user" } })` call with the pinned Genkit 1.41.0, the `options` passed here has no `context`: Genkit carries the authenticated generation context through async-local storage. I reproduced the tool handler receiving `{ userId: "authenticated-user" }` while `actor: (_input, options) => options?.context?.userId ?? "anonymous"` sent `"anonymous"` to Guard. The same happens in `guard-middleware.ts:288`. This prevents the documented context-based resolvers from identifying the caller during normal generation, potentially missing actor-specific policy scope or blocking the call when the resolver throws. Could both paths make the active Genkit context available to the resolvers and add an `ai.generate()` regression test with authenticated context?

arcjet-guard/src/mastra/v1/guard-processor.ts
```diff
@@ -266,12 +279,19 @@ export function guardProcessor(
       "mastra.phase": phase,
       ...policyMetadata,
     };
+    let remote: Awaited<ReturnType<typeof resolveActorInputs>> = {};
+    try {
+      remote = await resolveActorInputs(policy, input, requestCtx);
+    } catch {
+      denyTurn(abort, unavailableReason());
```
davidmytton: [P2] Honor onGuardError when actor/input resolution fails

This catch always calls `denyTurn()`, including when the caller explicitly sets `onGuardError: "allow"`. I reproduced both an async `actor` resolver rejection and an async `inputs` resolver rejection aborting `processInput` with that setting; neither reaches `runGate`. A temporary failure in an identity or policy-input lookup therefore stops the turn despite the configured fail-open behavior. Could this path honor `onGuardError`, consistently with the other adapters, and add coverage for rejecting resolvers with `onGuardError: "allow"`?

davidmytton (commented):
Two issues found in the actor/inputs resolver paths, both reproduced against this head.

PR #6285, opened 2026-09-15
test: track connections so HTTP/2 tests terminate in node 24.20.0

qw-in:
Update `@connectrpc/connect` to `2.2.0` and fix a change where the http2 session would hang in the tests:

> This patch changes the test helper only. It changes no source file. Before, the
helper tracked the HTTP/2 sessions of the test server. Now it tracks the TCP
sockets that the server accepts, and close() destroys those sockets.

@arcjet-rei I'd appreciate you having a look at the http2 handling change!


arcjet-rei (approved):
I couldn't understand why a dependency bump would lead to changes in the functions used, so I had Claude take a look at the change. It spent a loooooooong time looking at everything, and the tl;dr is that this is actually a patch for an upstream regression in Node 24 (less tl;dr version appended). This fixes that regression, and the dependency bump is worth taking, but the PR title and description should be updated to reflect what's actually going on here. Everything else is fine!

> ## What this patch does
> 
> This patch changes the test helper only. It changes no source file. Before, the
> helper tracked the HTTP/2 sessions of the test server. Now it tracks the TCP
> sockets that the server accepts, and `close()` destroys those sockets.
> 
> ## The cause is Node.js, not the dependency update
> 
> I put the lockfile back to `@connectrpc/connect` 2.1.2 and kept the old test
> helper. The test run did not complete, in the same way. The new version of
> connect-es is therefore not the cause.
> 
> I then ran `transport/test/wire.test.ts` on different versions of Node.js:
> 
> - Node 22.23.2 — 21 tests pass
> - Node 24.19.0 — 21 tests pass
> - Node 24.20.0 — 21 tests pass
> - Node 24.21.0 — the test run does not complete
> - Node 26.8.2 — 21 tests pass
> 
> The release date of Node.js 24.21.0 is 2026-09-07. With this patch, all 54
> transport tests pass on Node 24.21.0.
> 
> ## What Node.js 24.21.0 changed
> 
> `lib/internal/http2/core.js` is identical in 24.20.0 and 24.21.0. The change is
> in `src/node_http2.cc`, from nodejs/node#65093, "http2: adapt receive deferral
> for Node.js 24". `Http2Session::Close()` now defers its work when JavaScript
> calls it from inside `nghttp2_session_mem_recv()`. The same commit also stops
> `SendPendingData()` while the session receives data. Node.js 26 does not have
> that second condition, and Node.js 26 passes the tests.
> 
> ## What you see in the test
> 
> On Node 24.20.0 and on Node 24.21.0, `session.destroy()` leaves the socket open.
> Its state is `{ destroyed: false, readable: true, writable: false }`. On 24.20.0,
> each socket closes 1 to 5 milliseconds later. On 24.21.0, the socket of the
> "HTTP fallback" test never closes. `server.close()` waits for that socket, and
> its callback never runs. In that test, the client resets the stream after the
> 503 response, which is the path that nodejs/node#65093 changed.
> 
> ## CI
> 
> The last test run on `main` used Node 24.20.0 from the runner tool cache. The
> tests on `main` pass for that reason. When the runner image gets 24.21.0, the
> transport tests will not complete, on all three operating systems. This patch
> prevents a failure that is not yet visible.
> 
> ## One correction to the new comment
> 
> The comment in `proxy.ts` says that a session in the `closed` state ignores
> `destroy()`. That is not correct. `destroy()` runs, and it sets `destroyed` to
> `true`. It disconnects the session from the socket, but it does not close the
> socket. `session.socket` is a JavaScript `Proxy`, and after `destroy()` every
> property of that `Proxy` throws `ERR_HTTP2_SOCKET_UNBOUND`. A reference to
> Node.js 24.21.0 and nodejs/node#65093 in that comment would help the next reader.

PR #6289, opened 2026-09-16
feat(guard): check a guard label before sending it

arcjet-rei:
A guard label is compared against a published policy by exact string match, so a label the service would reject matches nothing and the guard silently does not run. Nothing in the decision says so: it reads `conclusion: "ALLOW"` and `hasFailedOpen()` is false.

`validateGuardLabel` refuses such a label up front. Twenty-one adapter factories across twelve vendor namespaces check a literal `action` at construction, as do the two policies nested inside another — the Claude Agent SDK inbound hook and the Vercel Eve approval response, each of which carries an `action` of its own. An empty action is left alone: these adapters treat it as unset and fall back to `tool.invoked`, which is itself valid.

`AJ1023` stops being informational. It means the service replaced the label with `invalid-label`, so the call is unevaluated policy and follows `onGuardError`, failing closed by default. Six places classify a decision — `runGuarded` plus a `runGate` per vendor namespace — so that rule lives in one predicate all six call, rather than being copied six times.

Capture warns instead of failing, and still sends the action as written: a capture call has no response to carry `AJ1023`, and a capture records what the application did.

Most of the diff is one inserted line per factory, plus tests.

Fixes ENG-1241. Implements the ADRs in arcjet/arcjet#9250 and arcjet/arcjet#9251.


arcjet-guard/src/agents/label.ts
```diff
@@ -0,0 +1,145 @@
...
+ *
+ * Too loose is recoverable, because the service still reports the rejection at
+ * call time as `AJ1023`. Too strict is not.
+ *
+ * The cases both sides agree on are in `test/_shared/guard-label-cases.json`,
+ * copied from the `arcjet` monorepo. See
+ * `docs/adrs/2026-09-15-each-sdk-checks-a-guard-label-before-sending-it.md`
+ * there for why every SDK carries a copy and what the shared cases do and do
+ * not gate.
+ */
```
qw-in, line 20: The ADR doesn't exist in this repo and may not make sense to reference directly in the SDKs

arcjet-guard/src/agents/label.test.ts
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Read the vendored cases, failing loudly on a malformed file.
+ *
+ * A fixture that parsed to nothing would leave every case below passing while
+ * testing none of them.
+ */
+function readSharedCases(): LabelCase[] {
+  const raw: unknown = JSON.parse(
+    readFileSync(new URL("../../test/_shared/guard-label-cases.json", import.meta.url), "utf8"),
+  );
```
qw-in, line 25: Is this supported in our runtimes yet?

```ts
import data from './data.json' with { type: 'json' };

```

qw-in (approved):

arcjet-guard/src/agents/label.ts
arcjet-rei: Good catch! Normally those get filtered out. Will fix.

arcjet-guard/src/agents/label.test.ts
arcjet-rei: I don't know exactly when JSON important (and `with { type: }`  landed, but I personally am fine with test code just doing the direct and simple thing.

arcjet-guard/src/claude-agent-sdk/v0/gate.ts
```diff
@@ -93,6 +94,18 @@ export async function runGate<T>(
     if (decision.conclusion === "ALLOW" && decision.hasFailedOpen()) {
```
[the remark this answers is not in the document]
arcjet-rei: Correct, and it was a real gap. `runGuarded` had both branches; the five `runGate` implementations had only the fail-closed one, so under `onGuardError: "allow"` a rejected label passed through a gate with no warning at all — reading as an evaluated allow.

Fixed in 0120e3c01, which adds the matching warn-only branch to all five gates:

```ts
if (decision.conclusion === "ALLOW" && labelRejectedByService(decision)) {
  // Failing open here still means the guard did not run, so say so rather
  // than letting this read as an evaluated allow.
  warnUnavailable(action, "failed-open", false);
}
```

The gates have no `judgedFully`, since they return a verdict rather than executing and capturing an outcome, so the warning is the whole of the fix there. `runGuarded` keeps both the warning and `judgedFully = false`.

This is the second time this pair of branches has been missed one side at a time — the same asymmetry existed in `runGuarded` itself and was caught while adding capture handling.

arcjet-guard/src/agents/label.ts
arcjet-rei: Fixed in 0120e3c014ef115b3a64645a06c837d13ba03d16.

